PR #14, opened 2021-05-19
Use multi-stage Docker build. Compile contracts into image.

nguyer:
This change significantly reduces the size of the published Docker container, down to ~30MB from ~950MB. It also includes compiled solidity contracts in the image which can be installed if needed.


peterbroadhurst (approved):
🗜️ 🐳

PR #16, opened 2021-05-20
Serve frontend

nguyer:
This PR adds the ability to serve a static HTML/CSS/JS frontend, if enabled in the config. This repo does not contain a frontend, so the config file will need to point to some place on the filesystem that a compiled frontend exists.

The following entry in `firefly.core` will serve HTML files from a directory called `frontend`.
```yaml
ui:
  path: frontend
```

> **NOTE**: The frontend is **not** enabled by default. If no `ui.path` key is present in the config file, the frontend will not be served.

The Dockerfile has also been updated to build the latest commit of https://github.com/kaleido-io/firefly-ui and include that in the image.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ firefly
 coverage.txt
 **/debug.test
 .DS_Store
+frontend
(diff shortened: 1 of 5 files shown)


peterbroadhurst (approved):
I can fill in the coverage on this one as I'm rounding stuff out there for the stuff I'm doing ☂️

nguyer:
Thanks. I am doing a bit more testing on this. Having some trouble with UI paths in the Docker container.

PR #23, opened 2021-05-26
Add basic E2E test

awrichar:
This is the first Go module I've ever written, so help me out here.

Naming and placement of this file are also up for discussion.

Would also like to have at least one other person test this before merging. Note that it currently doesn't work with the `master` version of [firefly-cli](https://github.com/kaleido-io/firefly-cli) due to some bugs - awaiting fixes from @nguyer on that.

### DIFF
--- a/test/e2e/restclient.go
+++ b/test/e2e/restclient.go
@@ -1,0 +1,36 @@
+package e2e
+
+import (
+	"github.com/go-resty/resty/v2"
+	"github.com/kaleido-io/firefly/pkg/fftypes"
+)
+
+var (
+	urlGetNamespaces     = "/namespaces"
+	urlBroadcastDatatype = "/namespaces/default/broadcast/datatype"
+	urlGetData           = "/namespaces/default/data"
+)
+
+func GetNamespaces(client *resty.Client) (*resty.Response, error) {
+	return client.R().
+		SetResult(&[]fftypes.Namespace{}).
+		Get(urlGetNamespaces)
+}
+
+func BroadcastDatatype(client *resty.Client, name string) (*resty.Response, error) {
+	return client.R().
+		SetBody(fftypes.Datatype{
+			Name:    name,
+			Version: "1",
+			Value: fftypes.JSONObject{
+				"property1": "string",
+			},
+		}).
+		Post(urlBroadcastDatatype)
+}
+
+func GetData(client *resty.Client) (*resty.Response, error) {
+	return client.R().
+		SetResult(&[]fftypes.Data{}).
+		Get(urlGetData)
+}
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -x
+set -euo pipefail
+
+CWD=$(dirname "$0")
+CLI=ff
+STACK_DIR=~/.firefly/stacks
+STACK_NAME=firefly-e2e
+STACK_FILE=$STACK_DIR/$STACK_NAME/stack.json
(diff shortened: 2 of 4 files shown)


peterbroadhurst (commented):
Think it would be good to move the current stuff that's in `test` down a level (the recommendation here seems good to me: https://github.com/golang-standards/project-layout/tree/master/test)
e.g. structure it:
`test/data` - contains `key.pem` and the `config` dir
`test/e2e` - exactly per this PR

test/e2e/run.sh
```diff
@@ -0,0 +1,29 @@
...
+CLI=ff
+STACK_DIR=~/.firefly/stacks
+STACK_NAME=firefly-e2e
+STACK_FILE=$STACK_DIR/$STACK_NAME/stack.json
+DOWNLOAD_CLI=true
+CREATE_STACK=true
+
+if $DOWNLOAD_CLI
+then
+	go get github.com/kaleido-io/firefly-cli/ff
```
peterbroadhurst, line 14: How where does this binary go, if I do a `go get` like this from within a sub-directory inside of a module? How would I find it to execute it?

awrichar: I believe `go get` installs to `~/go/bin`

awrichar: So this may depend on you having that on your `PATH`...

peterbroadhurst: Yeah, that was what I was thinking. I believe the `Makefile` uses some go tools... will go take a look at how I worked it out there

peterbroadhurst:
(... leaving a formal review to @nguyer on this one)

test/e2e/restclient.go
```diff
@@ -0,0 +1,36 @@
...
+	urlGetData           = "/namespaces/default/data"
+)
+
+func GetNamespaces(client *resty.Client) (*resty.Response, error) {
+	return client.R().
+		SetResult(&[]fftypes.Namespace{}).
+		Get(urlGetNamespaces)
+}
+
+func BroadcastDatatype(client *resty.Client, name string) (*resty.Response, error) {
```
nguyer, line 20: It might make sense to pass in a `*fftypes.Datatype` here rather than just a name. That way different tests could define different data types.

test/e2e/stack.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func GetMemberPort(filename string, n int) (int, error) {
+	jsonFile, err := os.Open(filename)
+	if err != nil {
+		return 0, err
+	}
+	defer jsonFile.Close()
+
+	jsonBytes, err := ioutil.ReadAll(jsonFile)
```
nguyer, line 24: You can combine both of these operations into a single `ioutil.ReadFile()`

nguyer (approved):

nguyer:
Couple of minor suggestions, but looks like a great start!

awrichar:
@peterbroadhurst I opened #24 for moving the test data into a subfolder.

test/e2e/run.sh
awrichar: On a further look, `go get` may have the unintended consequence of adding a dependency to `go.mod`
https://stackoverflow.com/a/52764550/638122

test/e2e/restclient.go
awrichar: Yea, will probably change that in the future

test/e2e/run.sh
awrichar: Seems the dependency issue can be worked around with `GO111MODULE=off` to perform more of a "global" install

PR #24, opened 2021-05-26
Move test data to subfolder

awrichar:
The `test/config` folder is moved to `test/data/config` (and references updated). The file `test/key.pem` is deleted because it was empty and did not seem to be used (please correct me if it's needed).


peterbroadhurst (approved):
Thanks @awrichar - looks like something funny with the test dropped coverage. I can sort that out with my current branch work - will pull this branch into my branch so it merges with that.

PR #26, opened 2021-05-26
Use "go install" instead of "go get" for installing CLI

awrichar:
As of Go 1.16, "go install" is the cleanest way to install global
tools without messing up go.mod

See https://maelvls.dev/go111module-everywhere

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kaleido-io/firefly
 
-go 1.15
+go 1.16
 
 require (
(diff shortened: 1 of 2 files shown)


peterbroadhurst (approved):
Marking approval, but think we should also update this if 1.16 is minimum:
https://github.com/kaleido-io/firefly/blob/ef62985616c2dbf3e4e818dcfa414d837d397e00/go.mod#L3

PR #27, opened 2021-05-26
Add GitHub workflow for E2E tests

awrichar:
Open to thoughts on whether this should be integrated with the existing workflow (currently just named "Go") or remain a separate one.


awrichar:
Moved this to Draft status since it looks it may take some debugging to get GitHub to run the job correctly.

PR #28, opened 2021-05-27
Lower memory limit in Makefile to fix Docker Hub builds

nguyer:
I've run several successful builds locally while limiting my Docker VM's memory to 1.5 GB so I'm pretty confident this will work in Docker Hub where the limit is 2.0 GB.


peterbroadhurst (approved):
🗜️ 🐋

PR #32, opened 2021-05-28
Increase linter timeout to 5m

nguyer:
I think we're just over the default threshold of 1 minute for `golangci-lint` when we run it in Docker Hub. Likely caused by reducing its memory usage, it runs more slowly now than it used to. This should get it to build in Docker Hub again.


peterbroadhurst (approved):
💣 ⏱️ 🐎

PR #38, opened 2021-06-01
Minor Docker build optimizations

awrichar:
Move tool installs before local builds, and add Solidity build
output to .dockerignore.


peterbroadhurst (approved):
⏲️ 🐳

PR #51, opened 2021-06-09
Dynamic config

nguyer:
This PR adds a new `/config` endpoint which can be used for dynamically adding or modifying config at runtime via a REST API. Any change in configuration will result in a restart of all services and plugins (not the process itself though).

Due to how powerful (dangerous) the new configuration endpoints are, they have been moved to a separate “admin” HTTP listener which will listen on a separate port, if enabled at all. By default, it is disabled. To enable the admin endpoints, add the following lines to your firefly.core config:

```yaml
admin:
  enabled: true
  address: 0.0.0.0
  port: 5001
```

To set a configuration key make a `PUT` request to `/admin/api/v1/config/{key}` with a JSON request body.

For example, if you want to set the HTTP URL for ethconnct, you would make a `PUT` request to `/admin/api/v1/config/blockchain.ethereum.ethconnect` with a request body of:

```json
{
  "url": "http://some_url"
}
```

The payload will be merged with any existing config. This means that it is not possible to _remove_ configuration keys that are present in the config file, via the REST API.


nguyer:
One note on this PR, in the interest of time, a lot of the HTTP listener code got duplicated for the admin listener. I'm not a fan of this duplication, but I'd need some more time to think through a good way to refactor the code to support both listeners without duplication.

peterbroadhurst:
@nguyer - wonder if you wouldn't mind doing the merge with `main` before I review.

+1 to the point about the `apiserver` - don't think you started the issue there though. That package hasn't quite got the same structure as the rest of the code for pluggability, as many of those conventions appeared after it did. So I'm definitely up for helping with the refactor there, and that it can be a follow-on activity to this PR.

nguyer:
@peterbroadhurst Yeah this PR got pretty clobbered by the rewrite. I will fix that.

PR #60, opened 2021-06-10
Config API

nguyer:
This PR adds a new `/config` endpoint which can be used for dynamically adding or modifying config at runtime via a REST API. Any change in configuration will result in a restart of all services and plugins (not the process itself though).

Due to how powerful (dangerous) the new configuration endpoints are, they have been moved to a separate “admin” HTTP listener which will listen on a separate port, if enabled at all. By default, it is disabled. To enable the admin endpoints, add the following lines to your firefly.core config:

```yaml
admin:
  enabled: true
  address: 0.0.0.0
  port: 5001
```

To set a configuration key make a `PUT` request to `/admin/api/v1/config/{key}` with a JSON request body.

For example, if you want to set the HTTP URL for ethconnct, you would make a `PUT` request to `/admin/api/v1/config/blockchain.ethereum.ethconnect` with a request body of:

```json
{
  "url": "http://some_url"
}
```

The payload will be merged with any existing config. This means that it is not possible to _remove_ configuration keys that are present in the config file, via the REST API.


nguyer:
@peterbroadhurst I think I've resolved the git-ugliness with this one now, so we shouldn't have old commits mixed in here anymore. This one is ready for review now.

pkg/fftypes/config.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fftypes
+
+type ConfigRecord struct {
```
peterbroadhurst, line 19: As there is a `512` limit on the database record, it would be nice to have a `Validate` method on this that does a check length. Not a blocker to merge, as (unlike some other cases) it doesn't matter if we defer down to the DB for an error. But it would be more friendly to have an error at the API layer

internal/apiserver/route_delete_config_record.go
```diff
@@ -0,0 +1,44 @@
...
+
+var deleteConfigRecord = &oapispec.Route{
+	Name:   "deleteConfigRecord",
+	Path:   "config/{key}",
+	Method: http.MethodDelete,
+	PathParams: []*oapispec.PathParam{
+		{Name: "key", Example: "database", Description: i18n.MsgTBD},
+	},
+	QueryParams:     nil,
+	FilterFactory:   database.ConfigRecordQueryFactory,
```
peterbroadhurst, line 35: Shouldn't be a filter factory on this one I think

pkg/database/plugin.go
```diff
@@ -323,11 +323,23 @@ type PeristenceInterface interface {
...
+	GetConfigRecords(ctx context.Context, filter Filter) (offset []*fftypes.ConfigRecord, err error)
+
+	// DeleteConfigRecord - Delete config record
+	DeleteConfigRecord(ctx context.Context, key string) (err error)
 }
 
 // Callbacks are the methods for passing data from plugin to core
 //
-// If Capabilities returns ClusterEvents=true then these should be brodcast to every instance within
+// If Capabilities returns ClusterEvents=true then these should be broadcast to every instance within
```
peterbroadhurst, line 342: 👍  - thanks for the spot here

peterbroadhurst (approved):
Super thanks @nguyer 
Marked approval, with a couple of really minor things to consider.

PR #63, opened 2021-06-10
Fix E2E test

awrichar:
This gets the E2E test running again with the latest API. Can (should) be run manually before commits with `./test/e2e/run.sh`. Eventually will get it added to a GitHub action as well.

It's not as clean or as thorough as it could be, but it's a start. Feedback/future suggestions welcome.


peterbroadhurst (approved):
Awesome, thanks @awrichar 
Going to try adding in private messaging on top of this

PR #65, opened 2021-06-10
Updates to private messaging from e2e testing with CLI

peterbroadhurst:
Resolves a few nits in the end-to-end flow:
- Ensuring the group creation event is created, and included in the same namespace as the first message on that group
- Making a batch rewind wake up the event poller immediately, to assemble the message
- Generate a `group_confirmed` event only (no separate `message_confirmed`) when a group arrives


nguyer (approved):
Looks good to me

PR #82, opened 2021-06-16
Do not auto-reset, and add preinit mode requiring at least one config record

peterbroadhurst:
- Add a `POST` `/admin/config/reset` to initiate the reset separately from adding config records
- Add an `admin.preinit` option such that if there are no config records, a server will only start the admin server and not attempt to initialize. This means it will wait for the first config+reset to be `PUT` to it before attempting to start
- Added fix to https://github.com/hyperledger-labs/firefly/issues/79
  - In addressing this, I made some progress on the refactor discussed in https://github.com/hyperledger-labs/firefly/pull/51#issuecomment-857957200

Fixes #79

PR #85, opened 2021-06-17
Call Reset() before reading config on startup

nguyer:
This should fix default values not getting applied at startup.


peterbroadhurst (approved):
Thanks @nguyer - sorry I missed this in #82

PR #94, opened 2021-06-22
Fix pending unpinned sends

peterbroadhurst:
Currently if you send an unpinned message, it is correctly marked `pending: false` (with a confirmed timestamp) on the receiving node(s), but the sending node leaves it forever in a `pending` state on the sender.
This means in the UI, all the requests are shown _before_ the replies 🙃 

The problem was the unpinned send code was setting `pending` and `confirmed` _before_ sealing the message, which then reset those back to default.

image


nguyer (approved):
Brilliant! :shipit:

peterbroadhurst:
Coverage change is a timing issue - looking to address it before merge.

PR #98, opened 2021-06-24
Add publicURL config to admin and http config, and file based logging

peterbroadhurst:
- Public URL
    - Adds config option for cases where the listener address (such as `0.0.0.0`) is not the same as the URL that should be used externally for FireFly.
    - This affects the Swagger UI, which needs to perform an call from the browser to download the Swagger YAML.
- File logging
    - Uses lumberjack writer
    - Adds following config options:
    - `log.filename` - filename to log to
    - `log.maxAge` - maximum age before rotating
    - `log.maxSize` - maximum size
    - `log.maxBackups` - maximum number of files to keep
    - `log.compress` - whether to compress the backups


internal/apiserver/server.go
```diff
@@ -341,20 +341,19 @@ func (as *apiServer) swaggerUIHandler(url string) func(res http.ResponseWriter,
...
-	proto := "https"
-	if !apiConfigPrefix.GetBool(HTTPConfTLSEnabled) {
-		proto = "http"
+func (as *apiServer) getPublicURL(conf config.Prefix, pathPrefix string) string {
+	publicURL := conf.GetString(HTTPConfPublicURL)
+	if publicURL == "" {
+		proto := "https"
+		if !conf.GetBool(HTTPConfTLSEnabled) {
+			proto = "http"
+		}
```
nguyer, line 350: Should we consider the case where TLS is terminated at a load balancer or reverse proxy? In this case, I would think `HTTPConfTLSEnabled` would be `false` because FireFly isn't handling TLS itself, but the public URL would still need the `https` prefix. Maybe we just let `HTTPConfPublicURL` have the protocol built into it as well? In other words, maybe we should make `HTTPConfPublicURL` be the full URL?

nguyer (changes requested):
One question about the URL, but otherwise looks good.

I could be wrong, but I'm pretty sure if FireFly fails to read from it's config file it won't output anything to a log file, which may or may not be desirable. Wroth nothing though.

internal/apiserver/server.go
peterbroadhurst: > maybe we should make HTTPConfPublicURL be the full URL

... This was the intention of the change... let me double check (I moved things around a couple of times and maybe I lost it)

peterbroadhurst: Yeah, think that's in place. If `publicURL` is set, we require it to be a full URL

nguyer: Nevermind. I see now that this is inside an `if` which only happens when `publcURL` is not set. This looks good.

nguyer (approved):

Issue #81, opened 2021-06-16
Sync <-> Async bridge - REST Gateway for messages/event correlation


peterbroadhurst:
Checklist:

- [x] Webhooks implementation
- [x] Front-end request/reply
- [x] Docs

![Firefly _ Asset Trail - architecture whiteboard (16)](https://user-images.githubusercontent.com/6660217/122261895-cf20f700-cea2-11eb-9a20-99751152a650.jpg)

Current FireFly support using REST to submit messages to a remote party, then a pluggable interface for events to trigger in that party where the only implementation right no is WebSockets (long lived bi-directional pipe from a remote runtime).

This issue proposes adding two features, that combine to achieve the above pattern:
1. A REST interface on the front-end that is a sync<->async bridge
  - Takes the input payload and packages it into a message+data
  - Sends to a single participant, or group, as specified in the URL of the call
  - Blocks the REST call until a response with a `cid` (correlation ID) matching the request ID is received as an event
  - To begin with will be RPC style input/output `POST` rather than a full generated REST interface
2. A WebHook plugin for events
  - Takes messages that match the subscription, and extracts the data
  - Packages that data in a HTTP `POST` to a configured server
  - Waits for the response to the HTTP call
  - Packages that response with the correct `cid` into a message on the same group hash that the message arrived on

peterbroadhurst:
A more detailed view of how the end-to-end flow works:
![image](https://user-images.githubusercontent.com/6660217/123445279-be106e00-d5a5-11eb-8e39-032139a51697.png)

Source (sequencediagram.org):
```
title Sync <-> Async Example: Unpinned request, pinned response

participant "Member1 App\nRequester" as app1
database "FireFly DB" as db1
boundary "FireFly Core\n(Multiparty GW API)" as firefly1
entity "Data\nExchange 1" as dx1

entity "Blockchain" as blockchain

entity "Data\nExchange 2" as dx2
boundary "FireFly Core\n(Multiparty GW Proxy)" as firefly2
database "FireFly DB" as db2
participant "Member2 App\nResponder" as app2

app1->firefly1: POST /call/tag/my-request

activate firefly1 #20639b
activate firefly2 #20639b
firefly1-->firefly1: Sync<->Async\nsetup for reply
firefly1<->db1: Store\n"my-request"\nrequest
group Unpinned request
firefly1->dx1: Send message
dx1-->dx2: Secure private transfer
dx2->firefly2: New message
[...]

PR #100, opened 2021-06-28
Sync/async bridge front-end

peterbroadhurst:
Closes out the functional requirements in #81 

Core updates:
- New route: `POST` `/api/v1/namespaces/:ns/request/message`
   - Performs a request/reply exchange
   - Request is a `MessageInOut` in the body
     - Just like `.../send/message`
     - Data can be in-line or referenced
     - Requires `header.tag` to be set in message
   - Response is a `MessageInOut` in the body
      - Data is in-lined for you
- New event plugin called `system`
  - Simply delivers events to other components in the core
  - Cannot be disabled
  - Used with ephemeral events by the sync/async bridge
  - API prevents creation of durable subs externally on the transport
- Added `syncasync.Bridge` component
  - Included on periodic table as `[Sa]`
  - Only responsibility currently managing in-flight events
  - Could be extended to more gateway-like behavior in the future, taking arbitrary HTTP requests and encoding/propagating
- Added `filter.author` option on subscriptions
  - In pinned request scenarios, this allows the sender to exclude themselves from the list of responders
    - Without a filter on their subscription, they could attempt to respond to themselves
  - Currently only matters when the request is pinned
    - If a message is unpinned, there is no `message_confirmed` event on the sending party

Other enhancements:
- Fix to events where we could cross-emit on connections from the wrong transport
  - Resulted in `FF10190` `MsgMismatchedTransport` errors
  - Made significantly more likely by adding in the system events listener
- Adds `Request-Timeout` header to allow the client to customize the server-side timeout
  - No point the server waiting longer than the client is going to wait
  - Used header name from this old draft as basis: https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-00.html
  - This is in addition to the config-based request timeout (default=2mins) we already have
- Removes requirement to include yourself in a group
  - If you omit yourself, the identity configured on the node will be added
  - You can still add a specific identity that's a child of that in multi-tenant scenarios
- Added more config on `webhooks` subscriptions, after from thinking through use cases
  - `input.replytx` - set the reply transaction type dynamically from the request
  - `input.path` - add a path suffix to the URL (performs safe escaping)
- Removes storing `group_hash` on events
  - Was unnecessary as we already have filters on message specific items, and bloats events table
  - No need for a migration to remove it on existing envs
- Renames `MessageInput` to `MessageInOut` as it was confusing using it on output before
  - Also moved the helper to add data to a response to `fftypes` so it's in one place
- Fixed the Schema generation for the OpenAPI on create subscriptions
  - Previously it had all the dynamic options at top level, whereas they should be under `options`
- Change e2e test to teardown before building the docker image
  - Seemed to be necessary after recent CLI changes, which meant `ff remove` was deleting images
- Updated e2e test to include request/reply example
  - Replaced the old NoTx exchange test


internal/apiserver/server.go
```diff
@@ -293,7 +293,19 @@ func (as *apiServer) apiWrapper(handler func(res http.ResponseWriter, req *http.
...
+		reqTimeout := as.apiTimeout
+		reqTimeoutHeader := req.Header.Get("Request-Timeout")
+		if reqTimeoutHeader != "" {
+			customTimeout, err := fftypes.ParseDurationString(reqTimeoutHeader, time.Second /* default is seconds */)
+			if err != nil {
+				log.L(req.Context()).Warnf("Invalid Request-Timeout header '%s': %s", reqTimeoutHeader, err)
+			} else {
+				reqTimeout = time.Duration(customTimeout)
+			}
+		}
```
nguyer, line 305: Do we want to have a `maxRequestTimeout` to have an upper bound on this value? If someone sets an absurdly large `Request-Timeout` maybe we should return a `204`? If FireFly is deployed behind a load balancer or reverse proxy, admins may want to configure a `maxRequestTimeout` to not exceed the timeout of the network device in front of FireFly. Otherwise requests could get timeouts from the load balancer/proxy which would look different than timeouts from FireFly itself.

nguyer (approved):

nguyer:
@peterbroadhurst It looks like there may be a couple new config sections that need to be added to the CLI to be compatible with these changes? Do you mind posting the relevant config snippets here if there are any that are required?

internal/apiserver/server.go
peterbroadhurst: Sounds great 👍  - will add that

peterbroadhurst:
> Do you mind posting the relevant config snippets here if there are any that are required?

I do not believe there is anything new needed here.
The e2e test uses the CLI and it's all running fine with this.

I do have a TODO to sort out documentation on the config file that's in a format more consumable for non-devs, which maybe can build on what's done for the Swagger gen.

PR #102, opened 2021-06-28
Local docs build

peterbroadhurst:
- Add a build that you can run on your laptop with: `bundle exec jekyll serve`
- Add a link to the docs from the readme


nguyer (approved):

nguyer:
@peterbroadhurst Should we add a command to the `Makefile` so we can just run `make docs` to generate an updated version?

PR #107, opened 2021-07-01
Swagger generation and docs integration

nguyer:
This PR adds semi-automatic swagger generation. The reason that it is semi-automatic, and not fully automated, is that modifying the API contract should be a conscious decision, so it should not happen without the developer specifically intending to do so.

To update the `swagger.yaml` file in the repo, just run
```
make swagger
```

If source code changes which will affect the API spec in any way (including descriptions, examples, etc) are made without also intentionally updating the swagger file, a unit test will fail, asking if you forgot to run `make swagger`.

Additionally, the swagger UI has now been integrated into the docs pages hosted on GitHub pages. This page loads the auto generated `swagger.yaml` file generated as described above.


peterbroadhurst (commented):
@nguyer - is there a reason we chose not use the hosted version of the Swagger UI (like the runtime), or to dynamically pull in the latest during the docs build?
... wondering if we've thought about the implications of vendoring in the full swagger UI code directly into the hyperledger Git.

nguyer:
> @nguyer - is there a reason we chose not use the hosted version of the Swagger UI (like the runtime), or to dynamically pull in the latest during the docs build?
> ... wondering if we've thought about the implications of vendoring in the full swagger UI code directly into the hyperledger Git.

Hmm good point. Let me look at linking to the CDN hosted JS. The CSS might be a bit trickier. The only way to link to an external CSS file that I'm aware of is to actually add it to the theme/template. I'll look at reducing the complexity of what we're including the repo though.

nguyer:
@peterbroadhurst Okay I got it using the CDN like the built in spec page in the Go code. Tested locally on Chrome, FireFox, and Safari and it appears to render correctly on all three.

peterbroadhurst (approved):
This is great. Should be obvious to contributors, and keep us all in sync.

nguyer:
This resolves https://github.com/hyperledger-labs/firefly/issues/99

PR #114, opened 2021-07-02
Helm Chart for FireFly OS

hfuss:
Provides a simple Helm chart for deploying FireFly Core and DX. Features simple options for templating the `firefly.core` config as a `Secret`, as well as more open-ended overrides for future plugins, more complicated scenarios, etc.

DX can have mTLS certificate autogenerated and managed via [cert-manager](https://github.com/jetstack/cert-manager) for simplicity.

Includes a GHA workflow that uses [chart-testing](https://github.com/helm/chart-testing) and [KinD](https://kind.sigs.k8s.io/) for chart linting and "integration" testing i.e. do pods get installed and come up.

# TODOs
- [x] Get integration test passing
- [x] Liveness / readiness probes
- [x] Log level options
- Docs
  - [x] README.md
  - [x] templates/NOTES.txt
  - [x] Comments in values.yaml


deploy/charts/firefly/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Introduction
+
+
+
+## Prerequisites
+
+* Kubernetes 1.14+
+* Helm 3.6.0
+* PV provisioner support in the underlying infrastructure
```
peterbroadhurst, line 28: Worth describing the dependency for Cert Manager when using DX a bit more?

I could see from the chart there's a self-signed cert generated with [genselfsignedcert](https://helm.sh/docs/chart_template_guide/function_list/#genselfsignedcert), but it seemed like there was some orchestration via `cert-manager` required for it to function. I wasn't 100% sure why.

deploy/charts/firefly/templates/_helpers.tpl
```diff
@@ -0,0 +1,153 @@
...
+
+{{- define "firefly.core.config" -}}
+{{- if .Values.config.debugEnabled }}
+log:
+  level: debug
+debug:
+  port: 6060
+{{- end }}
+http:
+  port: 5000
```
peterbroadhurst, line 80: Should this reference the `service.httpPort`?

deploy/charts/firefly/templates/_helpers.tpl
```diff
@@ -0,0 +1,153 @@
...
+  address: 0.0.0.0
+admin:
+  port: 5001
```
peterbroadhurst, line 83: Should this reference the `service.adminPort`?

peterbroadhurst (commented):
Looks great @hfuss 👍 
Couple of minor questions on there

hfuss:
Passing IT build on the Kaleido fork: https://github.com/kaleido-io/firefly/runs/3025249695?check_suite_focus=true

deploy/charts/firefly/templates/_helpers.tpl
hfuss: A lot of time in charts the they fix the ports of the containers, and then only make the service ports configurable but I never really liked that convention. I went w/ your suggestions so that the service ports match what the container listens on, and the values are reused to template the service, pod, and application configs. Thanks!

deploy/charts/firefly/README.md
hfuss: TL;DR I removed the `genSelfSignedCert` `Secret` as that would have only created confusion for when the user tried to autocreate a certificate via `certificate.enabled` but didn't have cert-manager installed. I updated the README to then say cert-manager is recommended but not required.

Long and short of it:

I was trying to use `not (.Capabilities.APIVersions.Has "cert-manager.io/v1")` to detect if a user had opted-in to using cert-manager but didn't actually have it installed and then I'd use `fail` to have the chart throw an error message. This worked great.

However, when linting the chart `not (.Capabilities.APIVersions.Has "cert-manager.io/v1")` did not work as expected due to https://github.com/helm/helm/issues/3377#issuecomment-704574852. It also required a K8s cluster be present to lint a chart which felt wrong. So linting with my `ci/it-values.yaml` was failing as a result.

To get around the linting failuring, originally I had the mess you saw with `genSelfSignedCert` as a fallback for when they set `certificate.enabled` but didn't have cert-manager installed. But I decided thats sort of a silent failure as that type of cert would not work (DX requires a `subject` in the cert which you can't provide via that method) so instead to make the linter happy w/o confusing the user, I have the `not (.Capabilities.APIVersions.Has "cert-manager.io/v1")` check adding a YAML comment to the manifest saying this will fail and I'm letting Helm/K8s itself handle the failure if the API doesn't exist i.e. cert-manager isn't installed. 🙃

hfuss:
@drewmarshburn I can't request your review but love your input.

@peterbroadhurst this is ready to :shipit: I believe, but you (or a hyperledger-labs admin ?) will need to enable the new Helm workflow so this chart gets tested on future PRs that affect it. It will only run if `deploy/charts` is updated.

deploy/charts/firefly/templates/core/statefulset.yaml
```diff
@@ -0,0 +1,92 @@
...
+              protocol: TCP
+            {{- end }}
+            {{- if .Values.config.adminEnabled }}
+            - name: admin
+              containerPort: {{ .Values.core.service.adminPort }}
+              protocol: TCP
+            {{- end }}
+          livenessProbe:
+            tcpSocket:
+              port: {{ if and .Values.config.adminEnabled .Values.config.preInit }}admin{{ else }}http{{ end }}
```
drewmarshburn, line 61: Why is the admin port sometimes used for the liveness probe instead of always HTTP?

deploy/charts/firefly/values.yaml
```diff
@@ -0,0 +1,237 @@
...
+  postgresUrl: ""
+
+  # Whether or not to apply schema migrations automatically on startup, not recommended for production
+  postgresAutomigrate: false
+
+  # The URL of the HTTPS DataExchange for the node to use for the dataexchange plugin i.e. private messaging, only needed if `dataexchange.enabled` is set to false
+  dataexchangeUrl: ""
+
+  # The API key used to authenticate to HTTPS DataExchange, only needed if `dataexchange.enabled` is set to false
+  dataexchangeAPIKey: ""
```
drewmarshburn, line 32: Is this accounted for under `dataexchange.apiKey`?

deploy/charts/firefly/values.yaml
```diff
@@ -0,0 +1,237 @@
...
+  fireflyContractAddress: "/instances/contractAddress"
+
+  # The Postgres URL (including credentials and DB) for FireFly to for the database plugin
+  postgresUrl: ""
+
+  # Whether or not to apply schema migrations automatically on startup, not recommended for production
+  postgresAutomigrate: false
+
+  # The URL of the HTTPS DataExchange for the node to use for the dataexchange plugin i.e. private messaging, only needed if `dataexchange.enabled` is set to false
+  dataexchangeUrl: ""
```
drewmarshburn, line 29: I think I understand how this is technically a property of the core config (since this gets exposed from the node), but I wonder if this would made more sense as `dataexchange.url`?

deploy/charts/firefly/templates/dataexchange/secret.yaml
```diff
@@ -0,0 +1,25 @@
...
+      },
+      "p2p": {
+        "hostname": "0.0.0.0",
+        "port": {{ .Values.dataexchange.service.p2pPort }},
+        "endpoint": "https://{{ include "firefly.dataexchangeP2PHost" . }}"
+      },
+      {{- if .Values.dataexchange.apiKey }}
+      "apiKey": {{ .Values.dataexchange.apiKey | quote }},
+      {{- end }}
+      "peers": []
```
hfuss, line 23: TODO rm

deploy/charts/firefly/templates/dataexchange/certificate.yaml
```diff
@@ -0,0 +1,34 @@
...
+#          provide your own TLS certificate via tlsSecret.name or install cert-manager to resolve this.
+{{- end }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: "{{ include "firefly.fullname" . }}-dx"
+  labels:
+    {{- include "firefly.dataexchangeLabels" . | nindent 4 }}
+spec:
+  # NOTE: issuer should always be an internal / self-signed CA so that the subject is included
```
hfuss, line 13: TODO ref this issue in the comment https://github.com/jetstack/cert-manager/issues/3651

deploy/charts/firefly/README.md
peterbroadhurst: Thanks for the update, and the detailed background - very helpful (including as reference for others).

deploy/charts/firefly/templates/core/statefulset.yaml
hfuss: If FF is in a `preInit` state the HTTP port is not accessible.

Although folks likely will never want to deploy it in a `preInit` state in a cloud-native setting, we gave them the option and so had to make the probes align with that.

deploy/charts/firefly/values.yaml
hfuss: It depends. The chart's approach to FF config templating has three modes:
1. you override all of it w/ your own templating via `config.dataexchangeOverride`
2. you are deploying a DX alongside your FF instance, and so `dataexchange.enabled` is set. Therefore, if the DX was configured to use an API key via `dataexchange.apiKey` the config will be templated to use that.
3. you are not deploying a DX alongside FF via this chart, and have an external DX instead. You can just provide its config via `config.dataexchangeAPIKey` and `config.dataexchangeURL` at that point.

Relevant code: https://github.com/hyperledger/firefly/pull/114/files#diff-d5f35d3bee4948ee3394457bc8206e8aeb5bf8dffd5cffcd06915d0634f43c95R161-R180

Its a bit convoluted, but this was the only way I could think to provide autowiring of the DX config if using the chart to deploy the DX, but still providing simple and advanced methods for configuring it yourself.

deploy/charts/firefly/values.yaml
hfuss: See https://github.com/hyperledger/firefly/pull/114/files#discussion_r718603904https://github.com/hyperledger/firefly/pull/114/files#discussion_r718603904

drewmarshburn (approved):

peterbroadhurst (approved):
🎈 
🐳

PR #124, opened 2021-07-09
Add support for tokens

awrichar:


awrichar:
Added a possible token contract based on ERC-1155. See also #129 for background on contract exploration.

solidity_firefly/contracts/ERC1155MixedFungible.sol
```diff
@@ -0,0 +1,92 @@
+pragma solidity ^0.6.0;
```
peterbroadhurst, line 1: Could you pop a license header in this one?

peterbroadhurst (commented):
Great to see the progress here - with the asset manager periodic table element coming to life.

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+CREATE TABLE tokenpool (
+  seq            SERIAL          PRIMARY KEY,
+  pool_id        VARCHAR(80)     NOT NULL,
```
awrichar, line 4: pool_id should be a FireFly-generated UUID
on-chain ID should be another field (protocol_id?) - see transactions table for one example

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -7,6 +7,13 @@ CREATE TABLE tokenpool (
 );
 
 CREATE UNIQUE INDEX tokenpool_id ON tokenpool(pool_id);
-CREATE UNIQUE INDEX tokenpool_uri ON tokenpool(base_uri);
+
+CREATE TABLE tokenaccount (
+  seq            SERIAL,
+  member         VARCHAR(1024)   NOT NULL,
+  pool_id        VARCHAR(80)     NOT NULL,
```
awrichar, line 14: Use the FireFly UUID for pool

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -7,6 +7,13 @@ CREATE TABLE tokenpool (
 );
 
 CREATE UNIQUE INDEX tokenpool_id ON tokenpool(pool_id);
-CREATE UNIQUE INDEX tokenpool_uri ON tokenpool(base_uri);
+
+CREATE TABLE tokenaccount (
+  seq            SERIAL,
+  member         VARCHAR(1024)   NOT NULL,
```
awrichar, line 13: Should be the on-chain address (check naming)

awrichar: `onchain_id`

db/migrations/postgres/000024_create_token_tables.up.sql
awrichar: Also need another field for the token_id (relevant for non-fungible)

awrichar:
After discussing with some stakeholders, we've decided to split some of this functionality out as a separate micro-service here: https://github.com/hyperledger-labs/firefly-tokens-erc1155
Everything specific to ERC-1155 and the formulation of token IDs will be placed in that repository for better modularity, with a more generic REST API between it and FireFly for creating Token Pools and Accounts.

awrichar:
The initial PR for the tokens shim service is posted: https://github.com/hyperledger-labs/firefly-tokens-erc1155/pull/2

Closing this PR and will open a new one to plug into the micro-service once it has shaped up further.

awrichar:
This is now replaced by #154

Issue #109, opened 2021-07-01
Database change events - exposed over webhooks, for all resources


peterbroadhurst:
Currently we use in-memory database change events (connected via Go channels in-process from the SQLCommon DB code, through the Database Interface callbacks, to the various components like the batch assembly, event dispatch etc.).

The following problems exists with the current code:
- They are only implemented for a couple of key data objects
- There's no way for the UI to subscribe externally to these local Database events to live update
- As we evolve to active/active FireFly nodes, there's no way for events to propagate to other servers in the cluster

There are two problems that such change events need to solve:
1. How the UI detects changes that might cause views to be refreshed (this issue)
2. How an active/active FireFly environment performs low-latency propagation of "new message"/"new pin"/"new event" information to other FireFly servers in the cluster (future issues)

Having investigated change events in different database technologies, it seems like they are cumbersome and complex to implement.
- Most only support the native primary key of a database
- They require database specific features to be used
- They can cause overhead on the database

So actually both (1) and (2) above can be supported by some slightly smaller extensions to the approach of in-memory go channel events that come from the FireFly node that makes a change. It can simply propagate those to the UI, or other servers over a WebSocket as follows:

- Add ephemeral WebSocket (only) events for the UI to listen externally to these database triggers
  - Emitted for any local database change that might be significant, with as much information as we have
    - The information might be very minimal in some cases (like something updated on a tablee)
  - Lightweight events that can only be subscribed to by ephemeral websocket
    - Do not go into the `events` table - which itself is a database update, that would cause such an event

Checklist:
- [x] Design (in Git issue)
- [x] Implement
- [ ] Docs

peterbroadhurst:
... did start an exploratory branch on this, but it needs some thinking.

Particularly about how much to continue down the boilerplate approach, vs. moving the common part up a level (for SQL)

All tables that could have change events, have a Sequence based ID on the table for natural ordering
- Could be better codified in the plugin interface for adding non-SQL DBs later

We then have a split between tables that:
- Have a UUID as the API reference key
  - Logically could be the same as the sequence key, in a DB implementation that uses appropriate IDs natively
  - However, the datatypes assume an `int64` for the sequence key today
- Have a Hash as the API reference key
  - Groups - for detail reasons on how they are used

peterbroadhurst:
Updated scope after much further investigation, which included raising #119

Abandoned an approach of extending the in-memory events to all tables.

PR #125, opened 2021-07-09
Remove QL pure Go database, in favour of SQLite - pure Go and CGO

peterbroadhurst:
## Goodbye QL 😢 , Hello SQLite 🚀 

When we originally built FireFly's SQL implementation, the pure Go SQLIte implementation was not supported by the `goland-migrate` package.

As we wanted unit tests to be able to fully verify the SQL database layer with a real database, we introduce [QL](https://pkg.go.dev/modernc.org/ql) as a database that was supported by the migration tool.

Along the journey, we also added a super trivial "fake" blockchain interface for some early experimentation that worked on top of this.

Since then we've learned:
- QL is missing large pieces of functionality that we had to work around
  - A full types system
  - Individual column sorting
  - A "normal" column that can be auto-incremented as an ID (worked around with an `id(tableName)` function
  - Database triggers (see #109)
- The QL transaction model is very limited
  - Seems to only support one active transaction
  - When not used fully in line with the model, it results in deadlocks (rather than errors)
- QL 

This lead to code complexity to work around the limitations in unit tests, and also some costly investigations when attempts were made to use QL outside of the unit tests in a full FireFly runtime environment.

Now [golang-migrate](https://github.com/golang-migrate/migrate/tree/master/database/sqlite) has support of the pure Go implementation, I propose this PR to *completely remove all traces of QL* from the codebase.

We think it would be the wrong thing for anyone to pick up QL with FireFly, compared to the much more mature SQLite technology - which now has all the same non-functional characteristics (in process execution, option of a pure Go implementation as required)


peterbroadhurst:
@nguyer has updated the CLI to support SQLite in https://github.com/hyperledger-labs/firefly-cli/pull/60

Going to pop this back into draft, while I do some e2e testing with the full CGO version of SQLite

peterbroadhurst:
I'm currently investigating an issue where the standard assumptions we make on how database transactions function, might not apply with the pure Go implementation of SQLite (not seeing the same concerns with the more well adopted CGO version).

In https://github.com/hyperledger-labs/firefly/pull/125/commits/85d38703898f538387f201a98e3de32bc7fe11ff I added a test case that simply runs 5 go routines, doing 100 transactions with our standard semantics.
With more than 2 routines, we see:
```
time="2021-07-10T13:18:45-04:00" level=error msg="SQL commit failed: database is locked (5)"
```

Also observed when `:memory:` was used (rather than a real file on disk) the behavior was that each connection seemed to get its _own_ unique database - as the same test reported: `SQL insert failed: SQL logic error: no such table: testconc` with just 2 go routines.

Finally, running end to end tests with `sqlitego` was reporting:

```
[2021-07-09T20:06:07.051Z] ERROR update offset attempt 10: FF10114: Database begin transaction failed: SQL logic error: cannot start a transaction within a transaction (1) pid=1 role=batchmgr
```

So a bit more work to do to get to the bottom of it, but it might be we constrain use of the pure Go implementation of SQLite for unit tests only, where there's no concurrency, until we get to the bottom of it.

peterbroadhurst:
So I did some testing with well adopted CGO `sqlite3` package, and found:
- The behavior with `:memory:` where each connection seems to get its own database, was consistent
- The test ran reliably, without the `database is locked` error

So I'm going to remove the pure Go implementation of SQLite for now.
I will leave the ability to compile a pure Go FireFly, but for now that will be without an in-process database option.

peterbroadhurst:
@nguyer - thanks for all the work on the CLI.
I wonder if this should be tested with your fixes for `persistBatch` (`bp.ctx` -> `ctx`) and `beginOrUseTx` (logging).
Maybe I can review/merge your PR on that first, then I'll retest this with the e2e before we merge

peterbroadhurst:
Note per https://github.com/hyperledger-labs/firefly-cli/pull/60#issuecomment-877828743 we've got to merge the CLI changes first before the e2e test can go green on this one.

nguyer:
@peterbroadhurst Given what we now know about the Go implementation of SQLite, do you think it's appropriate to remove that as an option from the CLI?

peterbroadhurst:
Yeah, you’ll actually find that on the CLI PR comments 👍

nguyer (approved):
These changes look good to me. Thanks!

PR #126, opened 2021-07-12
Database events across all objects

### LINKED ISSUE #109: Database change events - exposed over webhooks, for all resources

peterbroadhurst:
Fixes #109 

- Doesn't include docs - need to do reference docs for WebSockets to cover the new `&changeevents=.*` option
- Updates e2e test with change events - turned on for all tests, so they're chatty, but hopefully that makes it easier to debug if they fail too


peterbroadhurst:
https://github.com/hyperledger-labs/firefly/pull/126/commits/e748d8f2766704b40fdf1b9e5da0c1c0a32920e5
adds a fix for this issue, that is a dangling resultset, specific to a timing condition in node registration.

I went through and checked this is the only place we've got the missing close.

Error example:
```
[2021-07-15T15:00:54.902Z]  INFO Confirming system broadcast 'ff_define_node' [469aa02c-68b6-48bb-af57-20605b0db261] dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.902Z] DEBUG SQL-> query: SELECT id, message_id, name, parent, identity, description, profile, created FROM orgs WHERE identity = $1 dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.903Z] DEBUG SQL<- query dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.903Z] DEBUG SQL-> query: SELECT id, message_id, owner, name, description, dx_peer, dx_endpoint, created FROM nodes WHERE name = $1 AND owner = $2 dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.905Z] DEBUG SQL<- query dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.905Z] DEBUG SQL-> query: SELECT id FROM nodes WHERE name = $1 AND owner = $2 dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.907Z] DEBUG SQL<- query dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.907Z] DEBUG SQL-> update: UPDATE nodes SET message_id = $1, owner = $2, name = $3, description = $4, dx_peer = $5, dx_endpoint = $6, created = $7 WHERE id = $8 dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.907Z] ERROR SQL update failed: pq: unexpected Parse response 'C' sql=[ UPDATE nodes SET message_id = $1, owner = $2, name = $3, description = $4, dx_peer = $5, dx_endpoint = $6, created = $7 WHERE id = $8 ] dbtx=JnMd6Dqc pid=1 role=aggregator
[2021-07-15T15:00:54.907Z] ERROR SQL rollback failed: driver: bad connection dbtx=JnMd6Dqc pid=1 role=aggregator
```

nguyer (approved):
Looks good!

PR #130, opened 2021-07-14
Rename topic field to topics

nguyer:
In a message header, previously there was a field named `topic` but it was of type `[]string`. It makes more sense to call this field `topics` because it is a list of several topic names. This PR changes this, but it is important to note that this is an API breaking change (though small in size).

Example message:

```json
{
  "header": {
    "tag": "new_widget_created",
    "topics": ["widget_id_12345"]
  },
  "data": [
    {
      "value": {
        "id": "widget_id_12345",
        "name": "superwidget"
      }
    }
  ]
}
```

### DIFF
--- a/docs/gettingstarted/broadcast_data.md
+++ b/docs/gettingstarted/broadcast_data.md
@@ -67,7 +67,7 @@ Status: `202 Accepted` - the message is on it's way, but has not yet been confir
     "author": "0x0a65365587a65ce44938eab5a765fe8bc6532bdf", // set automatically in this example to the node org
     "created": "2021-07-01T18:06:24.5817016Z", // set automatically
     "namespace": "default", // the 'default' namespace was set in the URL
-    "topic": [
+    "topics": [
       "default" // the default topic that the message is published on, if no topic is set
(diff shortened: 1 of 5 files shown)


JP1721114498 (approved):

docs/gettingstarted/broadcast_data.md
```diff
@@ -67,7 +67,7 @@ Status: `202 Accepted` - the message is on it's way, but has not yet been confir
     "author": "0x0a65365587a65ce44938eab5a765fe8bc6532bdf", // set automatically in this example to the node org
```
peterbroadhurst, line 67: @nguyer - have we moved this guidance documentation on the example result to somewhere else?

peterbroadhurst: ok - understand now, this isn't Git saying there's a diff, it's just pointing out that we've got non-JSON in a JSON "```json" block (which we did intentionally as it highlights better in our docs theme than "```javascript`

peterbroadhurst (approved):

PR #137, opened 2021-07-18
Add PUT for subscriptions and fix delete to clean up offset

peterbroadhurst:
Working on an app, I found a few things that are annoying when setting up subscriptions addressed in this PR:
- Deleting and recreating a subscription does not give you the events again
  - Because we weren't deleting the offset associated with the subscription, so when you recreated it you got the same offset
  - This is just a bug
- It's hard to create logic in your app to create/update subs
  - You have to query, then compare, then delete and recreate
  - Made harder when the above bug is fixed, as you'd have to work out the offset to recreate at

So this PR:
1. Fixes the bug - deleting a subscription, cleans up the offset
2. Introduces a `PUT` for `/api/v1/namespaces/:ns/subscriptions` where the `namespace` + `name` determine uniqueness
    - Adds an `updated` field on subscriptions for when they are changed
    - Adds handling to ignore no-op updates
    - Updates all connected apps, but does *not* reset the subscription offset
3. Changes `offset` to be referenced by the `RowID` and to be deleted by subscription ID (no more `ID` or `Namespace`)
   - This was necessary as otherwise I couldn't delete the offset
   - Also means there's no chance of an old subscription re-using an offset

In a PR chain with #136


internal/events/event_manager.go
```diff
@@ -168,20 +175,34 @@ func (em *eventManager) CreateDurableSubscription(ctx context.Context, subDef *f
...
+		subDef.ID = existing.ID
+		subDef.Updated = fftypes.Now()
+		subDef.Options.FirstEvent = existing.Options.FirstEvent // we do not reset the sub position
+		existing.Updated = subDef.Updated
+		def1, _ := json.Marshal(existing)
+		def2, _ := json.Marshal(subDef)
+		if bytes.Equal(def1, def2) {
+			log.L(ctx).Infof("Subscription already exists, and is identical")
+			return nil
+		}
```
nguyer, line 192: Just out of curiosity wondering if this is the best way to compare the subscriptions, by serializing them to JSON and comparing the bytes?

internal/orchestrator/orchestrator.go
```diff
@@ -71,6 +71,7 @@ type Orchestrator interface {
 	GetSubscriptions(ctx context.Context, ns string, filter database.AndFilter) ([]*fftypes.Subscription, error)
 	GetSubscriptionByID(ctx context.Context, ns, id string) (*fftypes.Subscription, error)
 	CreateSubscription(ctx context.Context, ns string, subDef *fftypes.Subscription) (*fftypes.Subscription, error)
```
nguyer, line 73: Do we still need this function or should `CreateUpdateSubscription` be the only thing that gets used now?

nguyer (approved):
Just a couple of minor questions, but looks good to me!

internal/events/event_manager.go
peterbroadhurst: Alternatives I thought through, all seemed to have the characteristic that if the structure changed, then this code would need to be updated to know about new fields.
As subscription options are pluggable by implementations, that's a lot of scope for the code to stop doing its job.

Any ideas for another way of comparing these nested structures (noting multiple pointers in them)?

internal/orchestrator/orchestrator.go
peterbroadhurst: `CreateSubscription` is used on `POST` and `CreateUpdateSubscription` is used on `PUT`.
Within orchestrator and events, they share a common implementation function `createUpdateSubscription` that takes a boolean to distinguish, to avoid code duplication.

internal/events/event_manager.go
nguyer: Those are good points. Hand't considered the plugin factor. My first thought would be to just have a `compare` method on the struct itself. But you are correct that if the structure changes then the comparison function would also have to change.

PR #139, opened 2021-07-19
Fix sequencing of submgr/webhooks and fix loop detection

### LINKED ISSUE #115: Webhook subscriptions do not work after restart
**Expected behavior:**
Webhook subscriptions work the same before/after a restart.

**Observed behavior:**
Webhook subscriptions appear not to send HTTP requests after restarting FireFly core

**Steps to reproduce:**
- Start a new stack
- Create a webhook subscription
- Send a message to test that a webhook is successfully received from the other node
- Stop/start the stack again
- Send another message that should be sent to the webhook - it appears to never arrive

Deleting and re-creating the webhook subscription again appears to work

peterbroadhurst:
- Fixes #115 - webhook subscriptions not working after restart
  - The problem here was that the submanager was initializing after Webhooks had registered the global connection, and the code was assuming no connections existed at the time submanager was starting. So it wasn't starting the dispatchers.
- Fixes an error where webhook message delivery stop functioning after the following error is logged:
  ```
  DEBUG Webhook subscription with reply enabled called with reply event '1cb8bc49-40e6-4496-bb96-620527d5eca9' pid=1
  ```
  - Here the problem was the loop detection code missing an ack on the event, so no further events were being delivered.

### DIFF
--- a/internal/events/subscription_manager.go
+++ b/internal/events/subscription_manager.go
@@ -148,7 +148,11 @@ func (sm *subscriptionManager) start() error {
 			continue
 		}
 		sm.durableSubs[*subDef.ID] = newSub
+		for _, conn := range sm.connections {
+			sm.matchSubToConnLocked(conn, newSub)
+		}
 	}
+	log.L(sm.ctx).Infof("Subscription manager started - loaded %d durable subscriptions", len(sm.durableSubs))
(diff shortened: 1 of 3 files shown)


nguyer (approved):
Looks good!

PR #143, opened 2021-07-23
Repeated groupinit messages: Make sure we sort the group members, before existence check

peterbroadhurst:
This fixes a case where we saw repeated `groupinit` messages being sent every time an unpinned message exchange was being performed in a request/reply scenario between two parties in the network.

The problem was the "_does the group already exist?_" check was being done using DB query with a hash calculated on an _unsorted_ list of members. Whereas the creation was correctly using `group.Seal()` which sorts the list of members before hashing. So we would repeat the creation over and over.


nguyer (approved):
Looks good to me

PR #145, opened 2021-07-27
Fix E2E tests with more than 3 members

nguyer:
The E2E tests expect members to be listed in the same order as they appear in the `stack.json` file. This change queries the API, asking it to sort the members in the expected order.


test/e2e/restclient.go
```diff
@@ -110,6 +110,7 @@ func GetBlob(t *testing.T, client *resty.Client, data *fftypes.Data, expectedSta
 func GetOrgs(t *testing.T, client *resty.Client, expectedStatus int) (orgs []*fftypes.Organization) {
 	path := urlGetOrganizations
 	resp, err := client.R().
+		SetQueryParam("sort", "created").
```
awrichar, line 113: 🤯

awrichar (approved):

PR #146, opened 2021-07-28
Misc cleanup surrounding Solidity contracts

awrichar:
Couple of standalone cleanup items identified while working #124. Might as well put them in while the larger development effort is still ongoing.

Depends on https://github.com/hyperledger-labs/firefly-cli/pull/66

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,10 @@ RUN make build
 
 FROM node:14-alpine3.11 AS solidity-builder
 WORKDIR /firefly/solidity_firefly
-ADD solidity_firefly .
+ADD solidity_firefly/package*.json .
 RUN npm install
 RUN npm config set user 0
+ADD solidity_firefly .
 RUN npx truffle compile
 
 FROM alpine:latest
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -11,9 +11,10 @@ RUN make build
 
 FROM node:14-alpine3.11 AS solidity-builder
 WORKDIR /firefly/solidity_firefly
-ADD solidity_firefly .
+ADD solidity_firefly/package*.json .
 RUN npm install
 RUN npm config set user 0
+ADD solidity_firefly .
```
nguyer, line 17: I'm sure there's a good reason for what you did, but why do you only bring the `package.json` in first, and then everything else after that?

nguyer (approved):
One quick question. Otherwise looks good.

Dockerfile
awrichar: It allows Docker to cache the results of `npm install` (which only depends on package.json). That way if you make a small tweak to a Solidity file, you don't have to wait for Docker to re-download all the npm packages before compiling it.

Hint: I was making a lot of small tweaks to Solidity files (:

nguyer: Ah yep. That totally makes sense. Good idea.

Issue #112, opened 2021-07-01
Provide an option on broadcast/send to wait for a message to be confirmed


peterbroadhurst:
A convenience method that waits for the `message_confirmed` (or and error/timeout) before it returns from the synchronous `POST` call.

- [x] Add a `?confirm` query parameter to the following endpoints, which are all currently `202 Accepted` only style
  - [x] `POST` `/namespaces`
  - [x] `POST` `/namespaces/{ns}/broadcast/datatype`
  - [x] `POST` `/namespaces/{ns}/broadcast/datatype`
  - [x] `POST` `/namespaces/{ns}/send/datatype`
  - [x] `POST` `/network/register/node`
  - [x] `POST` `/network/register/node/organization`
  - [x] `POST` `/network/register/organization`
- [x] Enhance the `syncasync.Bridge` implementation to support the following, or simlar:
  - ~A `PrepareSend` operation, that allocates an `inflight` ID~
  - ~A `ConfirmSend` operation, that waits for a `message_confirmed` (or `message_rejected` or timeout)~
  - [x] A `SendConfirm` operation, that sends a prepared message and blocks for the response

peterbroadhurst:
So as I do this, I believe we should make a pretty significant breaking change to our API.

- Return the actual object that you are creating in response to the `POST` not the message
- ... which makes possible: Change the routes to be more intuitive:
  - `POST` `/namespaces`
  - `POST` `/namespaces/{ns}/datatypes`
  - `POST` `/namespaces/{ns}/messages/broadcast`
  - `POST` `/namespaces/{ns}/messages/private`
  - `POST` `/network/nodes` (with a query parameter to choose to _also_ register the node)
  - `POST` `/network/organizations`

Interested in @awrichar thoughts on this (fyi @nguyer).

It's going to have a significant impact as a breaking change to the API, so I think I have to do the extra code work to retain the old routes.

PR #152, opened 2021-08-09
Provide a blocking mode for async msgs, and add more intuitive POST routes returning types not messages

peterbroadhurst:
Implements #112 

- Does _lots_ of dependency work to allow the sync-async bridge to be callable from broadcast/privatemessaging/networkmap
  - Adds a `syshandlers` package for the system event handlers that we previously bundled into `events`
  - Adds a `sysmessaging` that abstracts broadcast/private sends, and reply exchanges
  - Updates the orchestrator initialixation order to manage passing things between the internal components
- Deprecates all the old routes
- Adds new routes that:
  - Are all grouped on the path with the resources they create
  - Consistently return the object type that is the input
  - Have a `?confirm` option to block until the confirmation event arrives (or a 500 on the case it's rejected)
- Leaves the old routes there functioning exactly as today
  - Gives time for the CLI setup, and other consumers to move to the new routes


peterbroadhurst:
Investigating e2e test failure

internal/oapispec/routes.go
```diff
@@ -53,12 +53,14 @@ type Route struct {
 	JSONInputSchema func(ctx context.Context) string
 	// JSONOutputValue is a function that returns a pointer to a structure to take JSON output
 	JSONOutputValue func() interface{}
-	// JSONOutputCode is the success response code
-	JSONOutputCode int
+	// JSONOutputCodes is the success response code
+	JSONOutputCodes []int
```
awrichar, line 57: Should this simply be `OutputCodes` (there's nothing really JSON about it). And does it only list success codes, or does it support errors too?

Side note: seems like this openapi route stuff could almost be a separate package altogether.

internal/apiserver/route_post_new_message_broadcast.go
```diff
@@ -0,0 +1,104 @@
...
+		{Name: "confirm", Description: i18n.MsgConfirmQueryParam, IsBool: true},
+	},
+	FilterFactory:   nil,
+	Description:     i18n.MsgTBD,
+	JSONInputValue:  func() interface{} { return &fftypes.MessageInOut{} },
+	JSONInputSchema: func(ctx context.Context) string { return broadcastSchema },
+	JSONOutputValue: func() interface{} { return &fftypes.Message{} },
+	JSONOutputCodes: []int{http.StatusAccepted, http.StatusOK},
+	JSONHandler: func(r *oapispec.APIRequest) (output interface{}, err error) {
+		waitConfirm := strings.EqualFold(r.QP["confirm"], "true")
```
awrichar, line 99: Why "confirm" instead of "sync"?

internal/syncasync/sync_async_bridge.go
```diff
@@ -22,57 +22,67 @@ import (
...
+	"github.com/hyperledger-labs/firefly/internal/sysmessaging"
 	"github.com/hyperledger-labs/firefly/pkg/database"
 	"github.com/hyperledger-labs/firefly/pkg/fftypes"
 )
 
 // Bridge translates synchronous (HTTP API) calls, into asynchronously sending a
 // message and blocking until a correlating response is received, or we hit a timeout.
 type Bridge interface {
+	// Init is required as there's a bi-directional relationship between private messaging and syncasync bridge
+	Init(sysevents sysmessaging.SystemEvents, sender sysmessaging.MessageSender)
```
awrichar, line 36: The circular dependency is maybe a point of concern... but I don't have an alternate suggestion at the moment, so probably just something to consider down the line if it can be improved.

awrichar (approved):
I think this is overall a major step forward. The previous routes always confused me a bit, and the new grouping seems better.

I wonder why the query param is "confirm" instead of "sync" (especially since most of the underlying code/comments use sync/async terminology) - but maybe there's a point to be made about specifically waiting for _confirmation_ from the chain. I'm not terribly opinionated either way on the name, but wanted to at least raise it.

peterbroadhurst:
I think the challenge with `sync` in particular, is that in the case of `message/broadcast` and `message/private` it is an asynchronous operation you're performing. The term is so overloaded, I think it's confusing.

However, I'm definitely up for a word better than `confirm` if you can think of one?
I chose this, because what it's technically doing is waiting for the message to enter the `confirmed` state (where it is no longer pending, and `rejected` is `false`).

peterbroadhurst:
... the other word I had considered was `block` - I don't know what your gut says on whether that's better?

awrichar:
👍 Makes sense, and I can't think of anything better (enough) to warrant changing it all. I think "confirm" is better than "block" (block is overloaded too, especially in the blockchain world 😄)

git-zventus:
Could you please provide some example on how to use this new confirm query parameter? I tried this
```
curl --location --request POST 'http://127.0.0.1:5001/api/v1/namespaces/default/send/message?confirm=true' \
--header 'Content-Type: application/json' \
--data-raw '{
  "data": [
    {
      "value": "Private message 1-4 #3"
    }
  ],
  "group": {
    "members": [{
      "identity": "org_4"
    }]
  }
}'
```

awrichar:
@git-zventus the `/send` endpoint (which was async only) was deprecated as part of these changes. The new messaging endpoints (which support `confirm`) are all grouped under `/message`, and the one you're looking for is `/namespaces/default/messages/private`.

Swagger_UI

git-zventus:
Thank you so much. I just tried sending a private message from 0 to 4 and it stays pending with confirm=True

image

How do I confirm the message from node 4?

nguyer:
@git-zventus That doesn't seem right. Since this PR is already merged, do you mind opening a new issue and adding the API request that you made that resulted in this behavior? I will investigate and try to reproduce it. Thanks!

git-zventus:
I actually just updated to the latest version and it works as expected. Thanks a lot!

PR #154, opened 2021-08-11
Add support for creating token pools

awrichar:
Initial support for tokens by plugging in an instance of https://github.com/hyperledger-labs/firefly-tokens-erc1155.


internal/config/config.go
```diff
@@ -97,10 +97,12 @@ var (
 	CorsEnabled = rootKey("cors.enabled")
 	// CorsMaxAge is the maximum age a browser should rely on CORS checks
 	CorsMaxAge = rootKey("cors.maxAge")
-	// DataexchangeType is the name of the data exchange plugin being used by this firefly name
+	// DataexchangeType is the name of the data exchange plugin being used by this firefly node
 	DataexchangeType = rootKey("dataexchange.type")
 	// DatabaseType the type of the database interface plugin to use
 	DatabaseType = rootKey("database.type")
+	// TokensType is the name of the tokens plugin being used by this firefly node
+	TokensType = rootKey("tokens.type")
```
awrichar, line 105: This implies a configuration of the form:
```
tokens:
  type: https
  https:
    url: http://localhost:5108
```

I don't think "type" is actually the right word here, and I believe we want to support multiple types of tokens in a single config. I'm working follow-up changes in this area, possibly toward something like this (but open to ideas):
```
tokens:
  - connector: https
    name: erc1155
    url: http://localhost:5108
  - connector: https
    name: erc20
    url: http://localhost:5109
```

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+CREATE TABLE tokenpool (
+  seq            SERIAL          PRIMARY KEY,
+  id             UUID            NOT NULL,
+  namespace      VARCHAR(64)     NOT NULL,
+  name           VARCHAR(64)     NOT NULL,
+  pool_id        VARCHAR(80)     NOT NULL,
+  is_fungible    SMALLINT        DEFAULT 1
```
awrichar, line 8: Couldn't decide whether to store this as a boolean or a string enum. We deal with strings in the API (and theoretically could support more types than just "fungible" and "nonfungible"), but this seemed simpler? Not sure if we have a precedent.

internal/assets/manager.go
```diff
@@ -0,0 +1,103 @@
...
+func (am *assetManager) getNodeSigningIdentity(ctx context.Context) (*fftypes.Identity, error) {
+	orgIdentity := config.GetString(config.OrgIdentity)
+	id, err := am.identity.Resolve(ctx, orgIdentity)
+	if err != nil {
+		return nil, err
+	}
+	return id, nil
+}
+
+func (am *assetManager) CreateTokenPool(ctx context.Context, ns string, pool *fftypes.TokenPool, waitConfirm bool) (*fftypes.TokenPool, error) {
```
awrichar, line 68: Need a follow-up to properly use `waitConfirm`.

internal/tokens/https/https.go
```diff
@@ -0,0 +1,223 @@
...
+			var msg wsEvent
+			err := json.Unmarshal(msgBytes, &msg)
+			if err != nil {
+				l.Errorf("Message cannot be parsed as JSON: %s\n%s", err, string(msgBytes))
+				continue // Swallow this and move on
+			}
+			l.Debugf("Received %s event %s", msg.Event, msg.ID)
+			switch msg.Event {
+			case messageReceipt:
+				err = h.handleReceipt(ctx, msg.Data)
```
awrichar, line 182: Note: we get a receipt here from the Tokens microservice, which is actually wrapped from the receipt it received from ethconnect.

Because we're also directly subscribed to ethconnect, the blockchain plugin _also_ gets the ethconnect receipt, dumps a bunch of logs about how it's unrecognized, and then moves on. It all works alright, but need a follow-up to clean up that logging behavior (without breaking ethconnect receipts that we do care about).

awrichar:
E2E tests will fail until 1) tokens Docker is published (see https://github.com/hyperledger-labs/firefly-tokens-erc1155/pull/3) and 2) CLI support for tokens is added.

awrichar:
I'm also positive that codecov bot is coming for me, because I didn't add new tests. Just wanted to vet that I'm on the right track before I add them 😇

internal/config/config.go
awrichar: Added a commit to make this change.

internal/tokens/tkfactory/factory.go
```diff
@@ -0,0 +1,60 @@
...
+
+func init() {
+	for _, p := range plugins {
+		pluginsByName[p.ConnectorName()] = p
+	}
+}
+
+func InitPrefix(prefix config.Prefix) {
+	// TODO: add better config support for arrays
+	for i := 0; i < 5; i++ {
```
awrichar, line 43: Our config service currently requires hard-coding all acceptable values ahead of time. Was not sure the best way to allow an array of indeterminate length. Currently I'll allow up to 5 elements, but it seems hacky.

go.mod
```diff
@@ -44,7 +43,7 @@ require (
 	github.com/spf13/cast v1.3.1 // indirect
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
-	github.com/spf13/viper v1.7.1
+	github.com/spf13/viper v1.8.1
```
awrichar, line 46: Had to upgrade Viper to support YAML arrays.

internal/apiserver/route_post_tokens.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		"name": {
+			"type": "string"
+		}
+	}
+}`
+
+var postTokens = &oapispec.Route{
+	Name:   "postTokens",
+	Path:   "namespaces/{ns}/tokens/{connector}",
```
awrichar, line 45: This should probably end in `/pools`

internal/apiserver/route_post_tokens.go
```diff
@@ -0,0 +1,63 @@
...
+var createPoolSchema = `{
+	"type": "object",
+	"properties": {
+		"type": {
+			"type": "string",
+			"enum": ["fungible", "nonfungible"]
+		},
+		"name": {
+			"type": "string"
+		}
```
awrichar, line 39: Should also allow passing an identity as `author`

internal/apiserver/route_post_tokens.go
```diff
@@ -0,0 +1,63 @@
...
+	Method: http.MethodPost,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+	},
+	QueryParams:     nil,
+	FilterFactory:   nil,
+	Description:     i18n.MsgTBD,
+	JSONInputValue:  func() interface{} { return &fftypes.TokenPool{} },
+	JSONInputMask:   []string{"ID", "Namespace", "PoolID"},
+	JSONInputSchema: func(ctx context.Context) string { return createPoolSchema },
```
awrichar, line 56: Could omit this, although the enum will not be generated.

db/migrations/postgres/000024_create_token_tables.up.sql
awrichar: @peterbroadhurst says it should be a 64 char string

awrichar:
Had to rebase after merging #156.

internal/tokens/https/https.go
awrichar: Fixed in #156

internal/assets/manager.go
```diff
@@ -0,0 +1,103 @@
...
+	pool.ID = fftypes.NewUUID()
+	pool.Namespace = ns
+
+	if err := am.data.VerifyNamespaceExists(ctx, ns); err != nil {
+		return nil, err
+	}
+
+	id, err := am.getNodeSigningIdentity(ctx)
```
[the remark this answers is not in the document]
peterbroadhurst: That then cascades down to here in the batch broadcast case, where we resolve it to a blockchain identity:
https://github.com/hyperledger-labs/firefly/blob/4e2ea9e4f757caeafef4ac6c995c3fc5983d0446/internal/broadcast/manager.go#L99-L106

There's no further verification of that identity before the transaction is _sent_.
For broadcast messages, there is a rejection if we can't resolve the identity in the networkmap on confirmation of that message - but I don't think that's needed for tokens (and needs to evolve for messaging, as we will need to use HD wallets and other anonymized identities). So any identity can be used, as long as it can be resolved.

internal/database/sqlcommon/tokenpool_sql.go
```diff
@@ -0,0 +1,92 @@
...
+			return err
+		}
+		existing = transactionRows.Next()
+		transactionRows.Close()
+	}
+
+	if existing {
+		if err = s.updateTx(ctx, tx,
+			sq.Update("tokenpool").
+				Set("id", pool.ID).
```
peterbroadhurst, line 61: This should be the `Where` (rather than being in the `Set`)?
Or do we have a new pattern being established in this collection?

internal/database/sqlcommon/tokenpool_sql_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/hyperledger-labs/firefly/mocks/databasemocks"
+	"github.com/hyperledger-labs/firefly/pkg/database"
+	"github.com/hyperledger-labs/firefly/pkg/fftypes"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestUpsertTokenPoolFailBegin(t *testing.T) {
```
peterbroadhurst, line 32: We seem to be missing the e2e test here - that's the most valuable one by far (the UTs just fill in the remaining coverage).

internal/events/token_pool_created.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/hyperledger-labs/firefly/internal/log"
+	"github.com/hyperledger-labs/firefly/pkg/fftypes"
+	"github.com/hyperledger-labs/firefly/pkg/tokens"
+)
+
+func (em *eventManager) TokenPoolCreated(tk tokens.Plugin, pool *fftypes.TokenPool, signingIdentity string, additionalInfo fftypes.JSONObject) error {
+
+	log.L(em.ctx).Infof("-> TokenPoolCreated id=%s author=%s", pool.ID, signingIdentity)
```
peterbroadhurst, line 27: Cool to see this pattern for a method entry/exit trace going in.
However, I do think it should probably be `Debugf` rather than `Infof` - for info level, we should just have the one notification when the actual successful thing has happened

internal/i18n/en_translations.go
```diff
@@ -189,4 +189,7 @@ var (
 	MsgRejected                    = ffm("FF10269", "Message with ID '%s' was rejected. Please check the FireFly logs for more information")
 	MsgConfirmQueryParam           = ffm("FF10270", "When true the HTTP request blocks until the message is confirmed")
 	MsgRequestMustBePrivate        = ffm("FF10271", "For request messages you must specify a group of private recipients", 400)
+	MsgUnknownTokensPlugin         = ffm("FF10272", "Unknown tokens plugin '%s'")
+	MsgMissingTokensPluginConfig   = ffm("FF10273", "Invalid tokens configuration - name and connector are required")
```
peterbroadhurst, line 193: Wonder if this is a little inconsistency in the use of the word `connector`.
Here we're using it to infer the transport type to talk between FireFly core, and the Tokens connector.
Probably `type`, or `transport` might be better 🤷  ... know there's on a limited number of names we can use for similar things here.

internal/orchestrator/orchestrator.go
```diff
@@ -295,7 +314,38 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
...
+	if err = or.dataexchange.Init(ctx, dataexchangeConfig.SubPrefix(or.dataexchange.Name()), &or.bc); err != nil {
+		return err
+	}
+
+	if or.tokens == nil {
+		or.tokens = make(map[string]tokens.Plugin)
+		for i := 0; i < tokensConfig.ArraySize(); i++ {
+			prefix := tokensConfig.ArrayEntry(i)
+			name := prefix.GetString("name")
+			connector := prefix.GetString("connector")
```
peterbroadhurst, line 326: See note above about naming here.

internal/orchestrator/orchestrator.go
```diff
@@ -295,7 +314,38 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
...
+			log.L(ctx).Infof("Loading tokens plugin name=%s connector=%s", name, connector)
+			plugin, err := tkfactory.GetPlugin(ctx, connector)
+			if err != nil {
+				return err
+			}
+			or.tokens[name] = plugin
+		}
+	}
+	i := 0
+	for _, plugin := range or.tokens {
```
peterbroadhurst, line 340: Any reason not to just init these in the loop above?

internal/orchestrator/orchestrator_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"testing"
 
 	"github.com/hyperledger-labs/firefly/internal/config"
+	"github.com/hyperledger-labs/firefly/internal/tokens/tkfactory"
```
peterbroadhurst, line 25: Should this be `tifactory` for consistency - e.g. it's the `[Ti] Tokens Interface` plug point.

internal/tokens/https/https.go
```diff
@@ -0,0 +1,223 @@
...
+	return h.callbacks.TokensTxUpdate(h, requestID, updateType, message, data)
+}
+
+func (h *HTTPS) handleTokenPoolCreate(ctx context.Context, data fftypes.JSONObject) (err error) {
+	ns := data.GetString("namespace")
+	name := data.GetString("name")
+	clientID := data.GetString("client_id")
+	tokenType := data.GetString("type")
+	poolID := data.GetString("pool_id")
+	authorAddress := data.GetString("author")
```
peterbroadhurst, line 128: As much as a note to self of a hard problem to think about as we work through this, but I'm wondering how the difference between blockchain Identity and FireFly Identity (think a DID here once we've got a useful Identity plugin built) is managed.

e.g. if someone else in the network creates a token pool, what identities are stored where, and are we consistent in how we name those identities on the API properties.

Feel it's super easy for us to miss this, in the case where there's no translation (so it happens to be the case today that the blockchain identity and the FireFly identity are the same string).

pkg/database/plugin.go
```diff
@@ -350,6 +350,11 @@ type iConfigRecordCollection interface {
 	DeleteConfigRecord(ctx context.Context, key string) (err error)
 }
 
+type iTokenPoolCollection interface {
+	// UpsertTokenPool - Upsert a token pool
+	UpsertTokenPool(ctx context.Context, data *fftypes.TokenPool, allowExisting bool) (err error)
```
peterbroadhurst, line 355: We will need the getters too. You might find it easiest to just copy/paste those all over from another collection and pop them in at the start (that's what we've done for adding other collections).

peterbroadhurst (changes requested):
This looks great @awrichar 
A few minor suggestions that it would be good to get in before the first merge, then a couple of more rambling larger thoughts for you to ponder and think about whether there's more to do in this PR, or they're feedback to consider in future PRs.

internal/assets/manager.go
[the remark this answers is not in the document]
awrichar: Added a new commit along these lines - maybe you can double-check?

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+CREATE TABLE tokenpool (
+  seq            SERIAL          PRIMARY KEY,
+  id             UUID            NOT NULL,
+  namespace      VARCHAR(64)     NOT NULL,
+  name           VARCHAR(64)     NOT NULL,
+  pool_id        VARCHAR(80)     NOT NULL,
```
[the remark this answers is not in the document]
awrichar: Works for me - the current plugin happens to use a uint256, which I probably calculated to have a max length of 80 digits at some point.

db/migrations/postgres/000024_create_token_tables.up.sql
```diff
@@ -0,0 +1,14 @@
...
+  type           VARCHAR(64)     NOT NULL
+);
+
+CREATE UNIQUE INDEX tokenpool_id ON tokenpool(id);
```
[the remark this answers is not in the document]
awrichar: Currently `seq` is the primary key, not `id` (may have copied this from another table)

docs/swagger/swagger.yaml
```diff
@@ -4119,6 +4119,79 @@ paths:
           description: Success
         default:
           description: ""
+  /namespaces/{ns}/tokens/{connector}/pools:
```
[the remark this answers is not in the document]
awrichar: We had a conversation one time that ended with "don't use 'token type' anywhere because it may be a loaded term". I agree `connector` is probably not correct here though.

internal/apiserver/route_post_token_pool.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/hyperledger-labs/firefly/internal/config"
+	"github.com/hyperledger-labs/firefly/internal/i18n"
+	"github.com/hyperledger-labs/firefly/internal/oapispec"
+	"github.com/hyperledger-labs/firefly/pkg/fftypes"
+)
+
+var createPoolSchema = `{
```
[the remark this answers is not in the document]
awrichar: Right, but I do want Swagger UI to enumerate the types (fungible/nonfungible), else it's very difficult to guess what is valid to put there...

internal/database/sqlcommon/tokenpool_sql.go
awrichar: Yea this looks totally wrong. I'll check.

internal/database/sqlcommon/tokenpool_sql_test.go
awrichar: Yea, I was planning to add when I add getters.

internal/events/token_pool_created.go
awrichar: This was definitely copied from somewhere; I'll revisit.

internal/i18n/en_translations.go
awrichar: We're really selecting which type of plugin to use (of which the only choice is currently "https", and which implies not only the transport but the structure of all the data going in and out of that API).

internal/orchestrator/orchestrator.go
awrichar: Seemed like all the previous logic was designed so that this method could be called with any plugin already set to non-nil, and it would just invoke Init without doing the discovery of plugins. At least that's how the unit tests are structured - so I added separate tests for discovery and for init of the plugin.

internal/orchestrator/orchestrator_test.go
awrichar: Fine by me as long as [Ti](http://www.ti.com) doesn't come for us 😄

internal/tokens/https/https.go
awrichar: Agreed; it's hard to mentally think about the separation when they're currently equivalent.

pkg/database/plugin.go
awrichar: Yea, was just picking a single thin thread since it already ended up as 1500+ LoC. But I can look at adding the getters if you think they should go in with the first push.

internal/apiserver/route_post_token_pool.go
```diff
@@ -0,0 +1,64 @@
...
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+		{Name: "connector", Description: i18n.MsgTBD},
+	},
+	QueryParams:     nil,
+	FilterFactory:   nil,
+	Description:     i18n.MsgTBD,
+	JSONInputValue:  func() interface{} { return &fftypes.TokenPool{} },
+	JSONInputMask:   []string{"ID", "Namespace", "PoolID"},
+	JSONInputSchema: func(ctx context.Context) string { return createPoolSchema },
```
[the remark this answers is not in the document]
awrichar: OK, works for me

internal/events/token_pool_created.go
awrichar: Looks like I copied from `BatchPinComplete`

db/migrations/postgres/000024_create_token_tables.up.sql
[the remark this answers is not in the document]
awrichar: It's also becoming quite confusing to call this `tokenpool.pool_id` when we also have `tokenpool.id`. Suggestions on naming this one (which is specifically the ID assigned by the connector backend)?

pkg/database/plugin.go
awrichar: I went ahead and pushed up some of the basic work for database getters, along with a very basic E2E database test.

db/migrations/postgres/000024_create_token_tables.up.sql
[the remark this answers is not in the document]
awrichar: Suggest `protocol_id`

db/migrations/sqlite/000024_create_token_tables.up.sql
```diff
@@ -0,0 +1,12 @@
...
+  seq            INTEGER         PRIMARY KEY AUTOINCREMENT,
+  id             UUID            NOT NULL,
+  namespace      VARCHAR(64)     NOT NULL,
+  name           VARCHAR(64)     NOT NULL,
+  pool_id        VARCHAR(80)     NOT NULL,
+  type           VARCHAR(64)     NOT NULL
+);
+
+
+CREATE UNIQUE INDEX tokenpool_id ON tokenpool(id);
```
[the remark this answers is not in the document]
peterbroadhurst: Ignore this - my mistake.

internal/apiserver/route_post_token_pool.go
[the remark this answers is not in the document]
awrichar: Went ahead and removed this in the latest

docs/swagger/swagger.yaml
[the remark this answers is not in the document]
peterbroadhurst: Yeah, think in this particular case we want the human doing the config, to set this (via the connector `name` in the config) to be an actual type... so just here (not in the DB or config) feels like `type` is good.

pkg/database/plugin.go
```diff
@@ -706,3 +719,12 @@ var BlobQueryFactory = &queryFields{
...
 	"created":    &TimeField{},
 }
+
+// TokenPoolQueryFactory filter fields for token pools
+var TokenPoolQueryFactory = &queryFields{
+	"id":          &UUIDField{},
+	"type":        &StringField{},
+	"namespace":   &StringField{},
+	"name":        &StringField{},
+	"protocol_id": &StringField{},
```
peterbroadhurst, line 729: We don't use underscores in our api field names (we use camel case - and here is case insensitive) - only the database schemas use underscores.
You should have a mapping in the SQL library to map the name
```suggestion
	"protocolid": &StringField{},
```

pkg/fftypes/tokenpool.go
```diff
@@ -0,0 +1,33 @@
...
+	TokenTypeFungible    TokenType = "fungible"
+	TokenTypeNonFungible TokenType = "nonfungible"
+)
+
+type TokenPool struct {
+	ID         *UUID     `json:"id,omitempty"`
+	Type       TokenType `json:"type"`
+	Namespace  string    `json:"namespace,omitempty"`
+	Name       string    `json:"name,omitempty"`
+	ProtocolID string    `json:"protocol_id,omitempty"`
```
peterbroadhurst, line 31: Important that we're consistent on our API
```suggestion
	ProtocolID string    `json:"protocold,omitempty"`
```

Note in `transactions.go` I see `protocolID` with a capital `D` 🤔  ... we need to sort out whether we're an `somethingID` or `somethingId` in this PR, and stick to it from now on.

peterbroadhurst (changes requested):
One last consistency change before we merge, sorry @awrichar

pkg/fftypes/tokenpool.go
peterbroadhurst: Have confirmed `streamID`, `backendID`, `protocolID`, `requestID` all are consistent... so sounds like we have a pattern. So it's now or never to change it to `Id` - otherwise we have our horse.

awrichar: I think we have some sort of Go style plugin that actually enforces `ID` over `Id`

pkg/database/plugin.go
awrichar: Alright, so the convention is:
- ProtocolID in Go structs
- protocolid in JSON bodies
- protocol_id in database columns

pkg/fftypes/tokenpool.go
```diff
@@ -28,6 +28,6 @@ type TokenPool struct {
 	Type       TokenType `json:"type"`
 	Namespace  string    `json:"namespace,omitempty"`
 	Name       string    `json:"name,omitempty"`
-	ProtocolID string    `json:"protocol_id,omitempty"`
+	ProtocolID string    `json:"protocolid,omitempty"`
```
peterbroadhurst, line 31:
```suggestion
	ProtocolID string    `json:"protocolID,omitempty"`
```

Per discussion, the pattern in place is `ID` in capitals.

peterbroadhurst (approved):
🪙  🦄  🚀

PR #156, opened 2021-08-14
Remove retry for correlating transactions

awrichar:
Counterpart to https://github.com/hyperledger-labs/firefly-ethconnect/pull/139.

This will reduce the noisy logs when using tokens, because tokens generates ethconnect events that are "unexpected" from the perspective of the blockchain plugin. There will still be a single warning printed for each token transaction, but that's...a little better?

### DIFF
--- a/internal/events/transaction_update.go
+++ b/internal/events/transaction_update.go
@@ -26,17 +26,17 @@ import (
 	"github.com/hyperledger-labs/firefly/pkg/fftypes"
 )
 
-func (em *eventManager) TxSubmissionUpdate(bi blockchain.Plugin, txTrackingID string, txState fftypes.OpStatus, protocolTxID, errorMessage string, additionalInfo fftypes.JSONObject) error {
+func (em *eventManager) TxSubmissionUpdate(bi blockchain.Plugin, tx string, txState fftypes.OpStatus, protocolTxID, errorMessage string, additionalInfo fftypes.JSONObject) error {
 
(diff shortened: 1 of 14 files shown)


internal/events/transaction_update.go
```diff
@@ -17,34 +17,27 @@
...
-		}
-		return (err != nil && attempt <= em.opCorrelationRetries), err
-	})
+	operations, _, err := em.database.GetOperations(em.ctx, filter)
+	if err == nil && len(operations) == 0 {
+		err = i18n.NewError(em.ctx, i18n.Msg404NotFound)
+	}
 	if err != nil {
-		log.L(em.ctx).Warnf("Failed to correlate tracking ID '%s' with a submitted operation", txTrackingID)
+		log.L(em.ctx).Warnf("Failed to correlate tracking ID '%s' with a submitted operation", tx)
```
peterbroadhurst, line 40:
```suggestion
		log.L(em.ctx).Debugf("TX '%s' submission update ignored, as it was not submitted by this node", tx)
```

peterbroadhurst (approved):
Looks good to me - with the one small suggestion.

For reference for others reading the PR, the ID in the blockchain plugin (ethconnect/fabconnect etc.) will now match the ID of the `transaction` object in FireFly.

FYI @panghalamit @jimthematrix - support for API-supplied tracking IDs will be needed in FabConnect as well.

PR #158, opened 2021-08-17
Add config infrastructure for arrays of config

peterbroadhurst:
Hopefully the test makes it clear how this works.

Had to jump through a couple of hoops in the `viper` API to allow defaults to work correctly, but per the test it should be all good now.


go.mod
```diff
@@ -44,7 +43,7 @@ require (
 	github.com/spf13/cast v1.3.1 // indirect
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
-	github.com/spf13/viper v1.7.1
+	github.com/spf13/viper v1.8.1 // indirect
```
awrichar, line 46: why indirect?

peterbroadhurst: 🤷  - no idea... I'll remove that

internal/config/config_test.go
```diff
@@ -115,6 +116,36 @@ func TestPluginConfigArrayInit(t *testing.T) {
...
+- name: sally
+  key1: explicit value
+  key2:
+  - arr1
+  - arr2
+`))
+	assert.Equal(t, 2, tokPlugins.ArraySize())
+	assert.Equal(t, 0, NewPluginConfig("nonexistent").Array().ArraySize())
+	bob := tokPlugins.ArrayEntry(0)
+	assert.NoError(t, err)
```
awrichar, line 139: Feels like this assert should be a few lines higher

internal/config/config.go
```diff
@@ -376,8 +389,15 @@ func GetKnownKeys() []string {
 
 // configPrefix is the main config structure passed to plugins, and used for root to wrap viper
 type configPrefix struct {
-	prefix string
-	keys   map[string]bool
+	prefix   string
+	keys     map[string]bool
+	defaults map[string][]interface{}
```
awrichar, line 394: Is this ever used? Seems like `defaults` are only used on `configPrefixArray` types and not `configPrefix` types.

awrichar:
One further problem with using this in practice - some useful config groups are pre-defined in utilities like `wsclient.InitPrefix()` and `restclient.InitPrefix()`. Unfortunately, these take in a parameter of type `config.Prefix` and not `config.PrefixArray`. Is there a good way to support (for instance) an array of REST servers and leverage these existing definitions?

peterbroadhurst:
Additions look great to me @awrichar 👍 
Feel free to review+merge when you're ready

awrichar (approved):

PR #160, opened 2021-08-20
Make sync-async helpers less specific to messaging

awrichar:
This changeset simplifies the `inflightRequest` interface and adds 3 new generic constructs:
* `inflightResponse`
* `requestType`
* `requestSender`

The core helpers of sync-async (`sendAndWait` and `eventCallback`) now deal with these generic types and perform much less messaging-specific inspection of the requests and responses. There are clearer extension points for new request-response flows that may or may not relate to messaging at all.

The end goal is that this will make it easier to wrap tokens (and other future operations) in sync-async handlers.


peterbroadhurst (approved):
Looks great 👍

PR #161, opened 2021-08-20
Add an "id" parameter to SendMessageWithID

awrichar:
Given the method naming, it feels more intuitive to pass this in (rather
than requiring the sender to set it on a nested field before invoking).

Signed-off-by: Andrew Richardson <andrew.richardson@kaleido.io>


peterbroadhurst (approved):
👍  🆔   📥

PR #162, opened 2021-08-23
Add "confirm" parameter when creating token pools

awrichar:
This PR primarily adds the "confirm" parameter to token pool creation, which routes a request from Asset Manager to the Sync Async Bridge.

I also realized that by using the approach here (passing in a `RequestSender` callback), the circular dependency between `syncasync` and `syshandler` could be broken. I adjusted the other 3 existing sync-async operations (confirmed broadcast, confirmed private, request reply) to follow the same pattern. This means that `syncasync` now contains no logic at all for _sending_, only logic for tracking inflight IDs and unpacking system events to match them up.

Open to feedback on whether this callback pattern is the optimal design, but I do think removal of the circular calls makes the code easier to trace.


internal/apiserver/route_post_request_message.go
```diff
@@ -41,7 +41,11 @@ var postRequestMessage = &oapispec.Route{
 	JSONOutputValue: func() interface{} { return &fftypes.MessageInOut{} },
 	JSONOutputCodes: []int{http.StatusOK}, // Sync operation
 	JSONHandler: func(r *oapispec.APIRequest) (output interface{}, err error) {
-		output, err = r.Or.SyncAsyncBridge().RequestReply(r.Ctx, r.PP["ns"], r.Input.(*fftypes.MessageInOut))
+		msg := r.Input.(*fftypes.MessageInOut)
+		if msg.Header.Group == nil && (msg.Group == nil || len(msg.Group.Members) == 0) {
```
peterbroadhurst, line 45: We've tried to avoid putting this type of functional decision logic into the router, and instead have it in the layers below. However, I understand there's no layer currently that implements this above the sync-async bridge.

My suggestion is to:
- Move these couple of lines of logic down into a common method on `orchestrator`
- Stop exposing the `SyncAsyncBridge()` function out of orchestrator, and instead make it an internal call

awrichar: Done

peterbroadhurst (approved):
Thanks @awrichar 

➡️ 🕐 ↩️

PR #163, opened 2021-08-23
Reject token pool events with duplicate info

awrichar:


peterbroadhurst (changes requested):
There's a pattern in the code for existing cases currently, where if you do an "upsert" and the object already exists, you get a signal error for `IDMismatch`:
https://github.com/hyperledger-labs/firefly/blob/3dedf94c1e47a5487e6535075793ba3df07197bc/internal/database/sqlcommon/namespace_sql.go#L53-L77

Feels like we should be consistent.

awrichar:
@peterbroadhurst Ah you're right, and I actually copied that logic already (so the duplicate ID check here is redundant). I think I still need to check for a duplicate name, but probably need to roll that down into `UpsertTokenPool` as well instead of having it in the event handler.

awrichar:
Reworked to do the uniqueness check in the database layer.

peterbroadhurst (approved):
Thanks @awrichar

PR #166, opened 2021-08-24
Add GET routes for token pools

awrichar:


internal/assets/manager.go
```diff
@@ -117,6 +119,44 @@ func (am *assetManager) CreateTokenPoolWithID(ctx context.Context, ns string, id
 	return pool, am.database.UpsertOperation(ctx, op, false)
 }
 
+func (am *assetManager) scopeNS(ns string, filter database.AndFilter) database.AndFilter {
```
awrichar, line 122: A few of these private helper methods were copy-pasted from `orchestrator`. Wonder if anything should be centralized/reused?

peterbroadhurst: I don't think any existing package is at the right scope level to be used for this, but certainly a candidate to introduce one if you'd like to here. I remember thinking the same the last time I duplicated these... they're pretty small, but it goes against the grain to copy/paste them.

peterbroadhurst (approved):
Marking approval, but it needs a `make swagger` before we merge.

internal/assets/manager.go
awrichar: OK, I'll leave them duplicated for now but will be on the lookout for a way to consolidate.

awrichar:
Updated the Swagger, but looks like it now wants @nickgaski to approve.

PR #174, opened 2021-08-30
Support multi-threaded "make mocks"

awrichar:
Generating mocks can be sped up significantly by splitting each file out
as a separate target and using e.g. "make mocks -j8".


peterbroadhurst (approved):
🐚 
🧨

PR #175, opened 2021-08-30
Remove unneeded "allowExisting" param from some db methods

awrichar:
If the param is always true, no need to have a param.

Note that token pools were sometimes erroneously passing false, but
should have been passing true (passing false would result in infinite
retries for a name conflict).


peterbroadhurst (approved):
🔗 
⏬ 
⛓️

PR #178, opened 2021-08-30
Add skeleton functionality for token accounts

awrichar:


db/migrations/postgres/000026_create_tokenaccount_table.up.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+CREATE TABLE tokenaccount (
+  seq            SERIAL          PRIMARY KEY,
+  namespace      VARCHAR(64)     NOT NULL,
+  pool_id        UUID            NOT NULL,
```
awrichar, line 5: I'm storing a `namespace` and a `pool_id` here. That's slightly redundant (since pool IDs will be globally unique). However, it seems that almost every items stores its namespace, so I followed that convention.

There's also an alternate possibility of storing `namespace` and `name`, which are also guaranteed to be unique. Then the `pool_id` would be literally useless and could likely be dropped.

pkg/database/plugin.go
```diff
@@ -459,7 +471,8 @@ const (
 type HashCollectionNS CollectionName
 
 const (
-	CollectionGroups HashCollectionNS = "groups"
+	CollectionGroups        HashCollectionNS = "groups"
+	CollectionTokenAccounts HashCollectionNS = "tokenaccounts"
```
awrichar, line 475: I chose `HashCollectionNS` (as the token account should be identifiable by the combination of pool+tokenIndex+identity). Want to be sure that's correct though, since we don't have a lot of things that fall into this bucket.

pkg/fftypes/tokenaccount.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (t *TokenAccountIdentifier) Hash() *Bytes32 {
+	b, _ := json.Marshal(&t)
+	var b32 Bytes32 = sha256.Sum256(b)
+	return &b32
+}
+
+type TokenAccount struct {
+	Identifier TokenAccountIdentifier `json:"identifier"`
```
awrichar, line 38: I don't love the generic name "Identifier" here, but couldn't come up with a better one. As with other types, I needed a grouping for the fields that are hashed.

pkg/fftypes/tokenaccount.go
```diff
@@ -0,0 +1,41 @@
...
+package fftypes
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+)
+
+type TokenAccountIdentifier struct {
+	Namespace  string `json:"namespace"`
+	PoolID     *UUID  `json:"poolId,omitempty"`
```
awrichar, line 26: Once again, the combo of Namespace+PoolID feels a bit odd... can't decide if this is the best combo of fields to return or not.

nguyer (approved):

db/migrations/postgres/000026_create_tokenaccount_table.up.sql
awrichar: The pool ID is generated by FireFly when the pool is created. My point was that I'm not being consistent about how to identify a pool. You can uniquely identify a pool by pool ID, or by namespace + name - so it basically has two primary keys.

The API routes for pools are specified in terms of the namespace + name. In the database here, I'm relating a token account to a given pool using namespace + pool ID. Just felt a little inconsistent once I finished coding it up...

peterbroadhurst: Most objects _intentionally_ have these two keys. The distinction (that's lived throughout the KAT days) is:
- The UUID is generated by the submitter of the object into the network. It lets you distinguish between two objects that were independently submitted, and to see if you "won the race" to establish the object.
- The name+namespace combination are the unique lookup key that an application can use in any environment (dev/test/preprod etc.) to refer to the same object in the network.

peterbroadhurst: e.g. the UUID is about giving the chain of custody linkage (including contributing to hash uniquness in various places as part of that). But you're right to an application it's kind of redundant at the point of use, and I think that's ok.

db/migrations/postgres/000026_create_tokenaccount_table.up.sql
```diff
@@ -0,0 +1,15 @@
...
+  token_index    VARCHAR(1024)   NOT NULL,
+  identity       VARCHAR(1024)   NOT NULL,
```
peterbroadhurst, line 7: Just a note that we need to reconcile this identity with the output of https://github.com/hyperledger-labs/firefly/issues/187 before the Tokens work can be considered MVP ready.

internal/apiserver/route_get_token_accounts.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/hyperledger-labs/firefly/internal/config"
+	"github.com/hyperledger-labs/firefly/internal/i18n"
+	"github.com/hyperledger-labs/firefly/internal/oapispec"
+	"github.com/hyperledger-labs/firefly/pkg/database"
+	"github.com/hyperledger-labs/firefly/pkg/fftypes"
+)
+
+var getTokenAccounts = &oapispec.Route{
+	Name:   "getTokenAccounts",
+	Path:   "namespaces/{ns}/tokens/{type}/pools/{name}/accounts",
```
peterbroadhurst, line 31: Am I understanding the uniqueness model correctly, that I would expect to see the following route appear in a future PR?
`GET` `namespaces/{ns}/tokens/{type}/pools/{name}/accounts/{index}`

internal/database/sqlcommon/tokenaccount_sql.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if existing {
+		if err = s.updateTx(ctx, tx,
+			sq.Update("tokenaccount").
+				Set("pool_id", account.Identifier.PoolID).
+				Set("namespace", account.Identifier.Namespace).
+				Set("token_index", account.Identifier.TokenIndex).
+				Set("identity", account.Identifier.Identity).
+				Set("balance", account.Balance).
+				Where(sq.Eq{"hash": account.Hash}),
```
peterbroadhurst, line 71: ... note to self, that I need to understand the significance of the hash (and why it's the key here) - expecting to get there through the rest of the review

internal/apiserver/route_get_token_accounts.go
peterbroadhurst: (more on this below)

pkg/database/plugin.go
peterbroadhurst: aha, this did indeed answer my earlier questions.

Yes, I think this is consistent with what we do for groups. It's slightly different, as for Groups we actually need to encode a variable size set of data together for uniqueness (a list of members), where as here it's always three strings. So I think we're saying the choice on the REST API is between:

`GET` `namespaces/{ns}/tokens/{type}/pools/{name}/accounts/{hash}`

and

`GET``namespaces/{ns}/tokens/{type}/pools/{name}/accounts/{index}/{identity}`

I guess my gut says the latter would be more obvious to people, but on the other hand `identity` in this context moving to instead being a `signing key` (because here we don't actually know the identity) expressed as a URI (per https://github.com/hyperledger-labs/firefly/issues/187). And having a URI within a pay in a URI would be wonky.

So the conclusion to use a `hash` sounds fine - if you want to query by identity/index you would need to use a custom filter query. Which we support (and would probably need indexes for in an upcoming PR)

internal/database/sqlcommon/tokenaccount_sql.go
awrichar: Mainly because I felt "accounts" were most analogous to "groups", and I followed that as a template. That is, accounts don't have any specific point of creation, they just "exist" for a particular combo of token pool + identity. They're assumed to have zero balance until a transfer/mint causes them to go positive, but since there is no actual creation event, it felt wrong to assign them a UUID - hence the use of a hash instead.

Certainly open to other routes if that doesn't seem right, but that summarizes my line of thought.

internal/apiserver/route_get_token_accounts.go
awrichar: Not entirely sure. Currently within a pool, if it is fungible, there is only one token_index (0). If it is non-fungible, there are token_index values starting at 1 and incrementing for each unique token.

A token account is a product of pool + token_index + identity. I'm not sure of the best way to craft a URL to query the balance of an individual account base on those 3 params.

db/migrations/postgres/000026_create_tokenaccount_table.up.sql
awrichar: Yes, the author/identity stuff already merged for token pools is also woefully incomplete, and it'll all need an overhaul when identities are settled in general.

internal/apiserver/route_get_token_accounts.go
awrichar: I see you already addressed this [elsewhere](https://github.com/hyperledger-labs/firefly/pull/178#discussion_r704675047)

pkg/database/plugin.go
awrichar: Yea, I was leaning more toward the latter. I couldn't decide if `{index}` actually belongs before or after `accounts` in the URI, and wasn't sure how `{identity}` should be represented in the URI (and if it should even be in the path at all, or should be a query param).

pkg/fftypes/tokenaccount.go
awrichar: After talking with others, it does seem we could drop namespace here. Also might want to use the on-chain protocolId instead of the FireFly poolId.

db/migrations/postgres/000026_create_tokenaccount_table.up.sql
awrichar: Discussed with @peterbroadhurst offline, and it seems that namespace and hash are not needed. Probably simplest to start by tracking a compound key of pool ID (or protocol ID?), token index, and identity.

pkg/database/plugin.go
```diff
@@ -352,16 +352,27 @@ type iConfigRecordCollection interface {
...
 	// GetTokenPools - Get token pools
-	GetTokenPools(ctx context.Context, filter Filter) (message []*fftypes.TokenPool, fr *FilterResult, err error)
+	GetTokenPools(ctx context.Context, filter Filter) ([]*fftypes.TokenPool, *FilterResult, error)
+}
+
+type iTokenAccountCollection interface {
+	// UpsertTokenAccount - Upsert a token account
+	UpsertTokenAccount(ctx context.Context, account *fftypes.TokenAccount) error
+
+	// GetTokenAccount - Get a token account by hash
```
peterbroadhurst, line 371:
```suggestion
	// GetTokenAccount - Get a token account by token ID and account identity
````

peterbroadhurst (approved):
One minor comment, on a comment, about a hash
With approval so you can merge after updating.

Issue #111, opened 2021-07-01
Field-level descriptions on Swagger / OpenAPI


peterbroadhurst:
- [ ] Support field-level i18n description keys on structures
  - Needs FireFly extensions to the Struct->JSON Schema parser
  - Could be contributed back `openapi3gen` in that community: https://github.com/getkin/kin-openapi
- [ ] Enhance enums (types etc.) so they show the possible values
   - [ ] Translation keys for each value
- [ ] Go through all the `fftypes` structures and add i18n keys
   > - This could be a separate work item to the others, done by a different person, as it needs field level knowledge

@gabriel-indik was pointing out the importance of this one in a conversation, as he was exploring the Swagger

peterbroadhurst:
After a bit of investigation, it seemed like we would struggle to continue to use the struct->schema generator in `kin-openapi` without an extension, so I've made a proposal to that project to get the ball rolling.
We will see what the feedback there is on the approach, before continuing. So I'm going to take ownership of this issue for now.
https://github.com/getkin/kin-openapi/pull/411

shorsher:
duplicate of https://github.com/hyperledger/firefly/issues/402 and addressed in v1.

PR #183, opened 2021-09-01
Add self-describing enum support to OpenAPI definition

peterbroadhurst:
One step on the journey to #111, allowing us to list the possible values for an `FFEnum` type (renaming `LowerCasedType`).

This builds on a feature added to the OpenAPI schema generator in https://github.com/getkin/kin-openapi/pull/411

Using the same approach in follow-on PRs, we can:
- Add in field-level descriptions with translation keys
- Remove the need for hard-coded schemas for complex objects like `Message`


peterbroadhurst:
DCU stuff ate my code... will try and re-instate and re-open this for review 😢

peterbroadhurst:
ok - think I got my code back, and made DCU happy with my comments.

nguyer (approved):
Looks good to me!

Issue #182, opened 2021-08-31
Identity plugin for multiple blockchain protocols


jimthematrix:
## Requirement
The identity plugin needs to be enhanced to handle requirements from different blockchain protocols. While the Ethereum identities are global, a single address string representing the unique account, Fabric identities are scoped to organizations.

- example Ethereum identity: `0x170dcf0187fb2f8884b63cfa2b1573c37f32eeb8`
- example Fabric identity: `Org1MSP::x509::CN=user001,OU=client::CN=fabric_ca.org1.example.com,OU=Hyperledger Fabric,O=org1.example.com,L=San Francisco,ST=California,C=US`, which contain these segments:
  - MSP ID
  - certificate type
  - user id
  - CA ID

While both types of identities can be represented in a string format, it'd be very cumbersome to require developers to input the full Fabric string as the `signer` property in the API payload.

Practically, a Fabric client would be tied to a single Fabric CA with a particular org MSP, so the API payload only needs the user ID portion to be specified, `user001` in the above example. But as soon as the identity gets into the FireFly core, it should be resolved to the fully qualified string before it gets replicated to other FireFly nodes.

In addition, the Fabric chaincode spits out the signer identity in the chaincode event in the fully qualified format, which means the matching logic in the aggregator won't work unless FireFly core resolves the supplied signer identity to use the same format.

## Proposed Solution 1
Provide separate identity plugins for each protocol. this means a new protocol support requires at least 2 plugins, the blockchain plugin and the identity plugin. there will be some overlap between the two because the blockchain plugin is also responsible for validating the identity strings

## Proposed Solution 2
Have the `OnChain` identity plugin consult the blockchain plugin for resolving the identity. this requires inter-plugin dependencies as the blockchain plugin must be initialized ahead of the identity plugin

shorsher:
Please see https://github.com/hyperledger/firefly-fir/pull/9

PR #184, opened 2021-09-01
Fabric plugin

jimthematrix:
Dependent on https://github.com/hyperledger-labs/firefly-fabconnect/pull/41

The plugin provides:
- `BatchPin` transactions and events support
- Identity resolution (note that pending enhancement for https://github.com/hyperledger-labs/firefly/issues/182 is left for a future PR given its significant implications to the identity plugin). For now these workaround are used when comparing the persisted tx signer with the event signer, which will be re-implemented as part of the solution for #182:
  - `internal/events/batch_pin_complete.go`: `subjectMatch()` 
  - `internal/events/persist_batch.go`: `persistBatchFromBroadcast()`


nguyer:
Why are there solidity files in this PR? Was that from files moving around, and maybe things got duplicated?

PR #185, opened 2021-09-02
Pack all token pool info into "data" field

awrichar:
The separate fields for the /pool API are going away. Any arbitrary
data should now be packed (as a JSON string) into the "data" field.

Depends on https://github.com/hyperledger-labs/firefly-tokens-erc1155/pull/19


internal/tokens/https/https.go
```diff
@@ -136,24 +136,28 @@ func (h *HTTPS) handleTokenPoolCreate(ctx context.Context, data fftypes.JSONObje
 		return nil // move on
 	}
 
-	hexUUIDs, err := hex.DecodeString(clientID)
-	if err != nil || len(hexUUIDs) != 32 {
-		log.L(ctx).Errorf("TokenPool event is not valid - bad uuids (%s): %+v", err, data)
+	unpackedData := createPoolData{}
+	err = json.Unmarshal([]byte(packedData), &unpackedData)
```
nguyer, line 140: Because the data field is arbitrary, is there any way someone could craft something that would be unpacked in an unexpected way here? For example, could someone override ID, TransactionID, etc?

nguyer (approved):

internal/tokens/https/https.go
awrichar: Short answer: no, I don't think it's an issue.

Longer answer: a lot of it has to do with the network architecture and which channels are considered private/immune to man-in-the-middle attacks. I'd assume we prevent direct access to the token connector, meaning the only way to "craft" data would be to actually write it on chain. In that case, if it matches the expected schema, it is a valid transaction for all intents and purposes, and all FireFly nodes listening to the stream would pick up and process it.

We do most of our validation at the db plugin layer just before upsert, and we do protect there against invalid things like changing immutable fields (namespace/name) once they've been recorded once.

PR #191, opened 2021-09-13
Gracefully shutdown http server when resetting config

nguyer:
Hopefully this should fix https://github.com/hyperledger-labs/firefly-cli/issues/80

The behavior in that issue has been very difficult to reproduce, but this code should help. Even if it doesn't fix that issue, it will not hurt anything else, and it helps prevent other http requests from possibly getting interrupted.


internal/apiserver/http_server.go
```diff
@@ -165,7 +165,11 @@ func (hs *httpServer) serveHTTP(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			log.L(ctx).Infof("API server context cancelled - shutting down")
-			hs.s.Close()
+			shutdownContext, cancel := context.WithTimeout(ctx, config.GetDuration(config.APIShutdownTimeout))
+			defer cancel()
+			if err := hs.s.Shutdown(shutdownContext); err != nil {
+				log.L(ctx).Error(err)
```
peterbroadhurst, line 171: Is there a test for this path?

nguyer: Good call out. I will add one.

peterbroadhurst (approved):

PR #192, opened 2021-09-15
Add identity manager and API input matrix logic

peterbroadhurst:
Work in progress on https://github.com/hyperledger-labs/firefly/issues/187

Key proposals in this code so far (which doesn't yet build)
- New `Identity Manager [Im]` component
   - This is _not_ a plugin, instead a consistent component to do all identity mapping
   - This would be expected to be where plugins would attach in the future
   - Implementation includes testing of the various matrix options
   - Implementation includes wiring this into `Orchestrator`
   - Implementation includes changing imports on calling packages
   - TODO (in this PR): Fix all the code that that currently calls `Resolve` on the now defunct Identity Interface
- New `fftypes.Identity` sub-object
   - Intended to replace single `author` field with `author`+`key` fields
   - Pass pointer to this part of the struct to the new `im.ResolveInputIdentity` function
   - TODO (in this PR): Embed this into all objects that can contain identity
- Stubbing out the `Identity Interface [Ii]` plugin for now
  - Still retained `onchain` key for config compatibility
  - Removed all methods from interface
  - Changed implementation to `TBD` for now in the code


internal/syshandlers/syshandler_network_org.go
```diff
@@ -37,30 +37,6 @@ func (sh *systemHandlers) handleOrganizationBroadcast(ctx context.Context, msg *
...
-	if err != nil {
-		l.Warnf("Unable to process organization broadcast %s - resolve identity failed: %s", msg.Header.ID, err)
-		return false, nil
-	}
-
-	if msg.Header.Author != id.OnChain {
-		l.Warnf("Unable to process organization broadcast %s - incorrect signature. Expected=%s Received=%s", msg.Header.ID, id.OnChain, msg.Header.Author)
-		return false, nil
-	}
-
```
nguyer, line 63: @peterbroadhurst I'm not sure if this is right. This block of code was giving me trouble so I had it commented out for most of time while I was working on this PR. My gut tells me it shouldn't be removed though.

README.md
```diff
@@ -186,6 +186,12 @@ It depends on the following Kaleido services:
   │       │   │               │    * Hashes data, and maintains mapping to payload references in blob storage
   │       │   └───────────────┘    * Integrates with blockchain interface [Bi] plugin
   │       │
+  │       │   ┌───────────────┐
+  │       ├───┤ identity [Im] │  - Central identity management service across components
+  │       │   │ manager       │    * Resolves API input identity + key combos (short names, formatting etc.)
+  │       │   │               │    * Resolves registered on-chain signing keys back to identities
+  │       │   └───────────────┘    * Integrates with Blockchain Interface and plugable Identity Interface (TBD)
```
awrichar, line 193: 🤩

db/migrations/sqlite/000027_create_key_fields.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE batches ADD COLUMN "key" VARCHAR(1024);
+UPDATE batches SET "key" = "";
```
awrichar, line 2: There's no way to make this `NOT NULL` in sqlite due to the migration path, is there? 😞

test/e2e/e2e_test.go
```diff
@@ -354,31 +354,31 @@ func TestE2EPrivateBlob(t *testing.T) {
...
+// 	// TODO: validate created pool
 
-	<-received2
-	<-changes2 // also expect database change events
-	// TODO: validate created pool
-}
+// 	<-received2
+// 	<-changes2 // also expect database change events
+// 	// TODO: validate created pool
+// }
```
awrichar, line 381: 😢

internal/identity/identitymanager.go
```diff
@@ -0,0 +1,242 @@
...
+		return "", i18n.NewError(ctx, i18n.MsgLocalOrgLookupFailed, orgKey)
+	}
+	return im.localOrgDID, err
+}
+
+func (im *identityManager) ResolveSigningKey(ctx context.Context, inputKey string) (outputKey string, err error) {
+	// Resolve the signing key
+	if inputKey != "" {
+		if cached := im.signingKeyCache.Get(inputKey); cached != nil {
+			cached.Extend(im.identityCacheTTL)
```
awrichar, line 144: Is it necessary to manually `Extend` each time we fetch? Shouldn't this be handled transparently by the LRU cache?

internal/identity/identitymanager.go
```diff
@@ -0,0 +1,242 @@
...
+			outputKey = cached.Value().(string)
+		} else {
+			outputKey, err = im.blockchain.ResolveSigningKey(ctx, inputKey)
```
awrichar, line 147: I see that this resolution always goes to the blockchain plugin. In the context of tokens, I'm wondering how identities need to be expressed. For instance, as of #211, pool creation includes an operation to the token connector and an operation to the blockchain connector. It's not guaranteed that these connectors are talking to the same blockchain, so is it possible we'd need to resolve two keys via the two plugins? Maybe a question to ponder with @peterbroadhurst...

internal/assets/manager.go
```diff
@@ -81,10 +80,7 @@ func (am *assetManager) CreateTokenPoolWithID(ctx context.Context, ns string, id
 		return nil, err
 	}
 
-	if pool.Author == "" {
-		pool.Author = config.GetString(config.OrgIdentity)
-	}
-	author, err := am.identity.Resolve(ctx, pool.Author)
+	err := am.identity.ResolveInputIdentity(ctx, &pool.Identity)
```
awrichar, line 83: I noted this in the identity code below too, but feels like we're ultimately going to need to support 2 "author" identities here (1 for the token pool connector and 1 for the blockchain). They're likely to be the same in simple cases (ie Ethereum+ERC1155), but we'll need to allow them to be specified differently in the input payload if needed.

internal/tokens/https/https.go
```diff
@@ -216,7 +216,7 @@ func (h *HTTPS) eventLoop() {
 	}
 }
 
-func (h *HTTPS) CreateTokenPool(ctx context.Context, identity *fftypes.Identity, pool *fftypes.TokenPool) error {
+func (h *HTTPS) CreateTokenPool(ctx context.Context, pool *fftypes.TokenPool) error {
 	data := createPoolData{
```
awrichar, line 220: This is mostly a note to myself, but this is where the resolved key for the token connector needs to propagate into the REST call. Opened https://github.com/hyperledger/firefly-tokens-erc1155/pull/30 to hopefully enable that to happen.

internal/batchpin/batchpin.go
```diff
@@ -32,42 +31,33 @@ type Submitter interface {
...
 
-	signingIdentity, err := bp.identity.Resolve(ctx, batch.Author)
-	if err == nil {
-		err = bp.blockchain.VerifyIdentitySyntax(ctx, signingIdentity)
-	}
-	if err != nil {
-		log.L(ctx).Errorf("Invalid signing identity '%s': %s", batch.Author, err)
-		return err
-	}
-
```
awrichar, line 57: I guess it's now assumed that the identity has been pre-verified before invoking this helper?

internal/syscore/syscore.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+
+	"github.com/hyperledger-labs/firefly/internal/events/system"
+	"github.com/hyperledger-labs/firefly/pkg/fftypes"
+)
+
+// SystemCore specifies an interface for global utility functions, without creating a cycle between components
+type SystemCore interface {
```
awrichar, line 27: I can't see where this is used...?

internal/networkmap/register_org.go
```diff
@@ -40,11 +40,32 @@ func (nm *networkMap) findOrgsToRoot(ctx context.Context, idType, identity, pare
 	return err
 }
 
+func (nm *networkMap) getLocalOrgSigningKey(ctx context.Context) (localOrgSigningKey string, err error) {
+	localOrgSigningKey = config.GetString(config.OrgKey)
+	if localOrgSigningKey == "" {
+		localOrgSigningKey = config.GetString(config.OrgIdentityDeprecated)
```
awrichar, line 46: Should we log a deprecation warning or something?

internal/broadcast/message.go
```diff
@@ -19,31 +19,40 @@ package broadcast
...
 	"github.com/hyperledger-labs/firefly/pkg/fftypes"
 )
 
 func (bm *broadcastManager) BroadcastMessage(ctx context.Context, ns string, in *fftypes.MessageInOut, waitConfirm bool) (out *fftypes.Message, err error) {
-	return bm.BroadcastMessageWithID(ctx, ns, nil, in, nil, waitConfirm)
+	return bm.BroadcastMessageWithID(ctx, ns, nil, in, nil, waitConfirm, false)
 }
 
-func (bm *broadcastManager) BroadcastMessageWithID(ctx context.Context, ns string, id *fftypes.UUID, unresolved *fftypes.MessageInOut, resolved *fftypes.Message, waitConfirm bool) (out *fftypes.Message, err error) {
+func (bm *broadcastManager) BroadcastMessageWithResolvedID(ctx context.Context, ns string, in *fftypes.MessageInOut, waitConfirm bool) (out *fftypes.Message, err error) {
```
awrichar, line 32: Can't find where this method is used...

internal/broadcast/message.go
```diff
@@ -19,31 +19,40 @@ package broadcast
...
 	resolved.Header.Namespace = ns
 	resolved.Header.Type = fftypes.MessageTypeBroadcast
-	if resolved.Header.Author == "" {
-		resolved.Header.Author = config.GetString(config.OrgIdentity)
-	}
 	if resolved.Header.TxType == "" {
 		resolved.Header.TxType = fftypes.TransactionTypeBatchPin
 	}
 
+	if !identityResolved {
```
awrichar, line 49: Personally I don't like when bits of a method are enabled/disabled with a boolean flag - feels like it should just be a separate method that you invoke beforehand if you need it.

Side note, after this block executes, the identity _is_ resolved, so it's odd that the boolean is propagated on line 75 rather than being flipped to `true`.

internal/broadcast/message.go
```diff
@@ -19,31 +19,40 @@ package broadcast
...
+	return bm.BroadcastMessageWithID(ctx, ns, nil, in, nil, waitConfirm, false)
 }
 
-func (bm *broadcastManager) BroadcastMessageWithID(ctx context.Context, ns string, id *fftypes.UUID, unresolved *fftypes.MessageInOut, resolved *fftypes.Message, waitConfirm bool) (out *fftypes.Message, err error) {
+func (bm *broadcastManager) BroadcastMessageWithResolvedID(ctx context.Context, ns string, in *fftypes.MessageInOut, waitConfirm bool) (out *fftypes.Message, err error) {
+	return bm.BroadcastMessageWithID(ctx, ns, nil, in, nil, waitConfirm, true)
+}
+
+// THIS IS WHERE THE NEW CODE FIRST CALLS ResolveInputIdentity
+// IT already has an author and key here
```
awrichar, line 37: This comment feels like a note to the developer and maybe not something that should be committed... or at least it should be cleaned up before being committed.

internal/events/aggregator.go
```diff
@@ -312,11 +312,11 @@ func (ag *aggregator) checkMaskedContextReady(ctx context.Context, msg *fftypes.
 	if err != nil {
 		return nil, err
 	}
-	l.Debugf("Group=%s Topic='%s' NextPins=%v Sequence=%d Pin=%s NextPins=%v", msg.Header.Group, topic, nextPins, pinnedSequence, pin, nextPins)
+	l.Debugf("Group=%s Topic='%s' Sequence=%d Pin=%s NextPins=%v", msg.Header.Group, topic, pinnedSequence, pin, nextPins)
 
 	if len(nextPins) == 0 {
 		// If this is the first time we've seen the context, then this message is read as long as it is
-		// the first (nonce=0) message on the context, for one of the members, and there aren't any earlier
+		// the first (nonce=0) message onw the context, for one of the members, and there aren't any earlier
```
awrichar, line 319: typo

internal/events/batch_pin_complete.go
```diff
@@ -33,28 +33,28 @@ import (
 //
 // We must block here long enough to get the payload from the publicstorage, persist the messages in the correct
 // sequence, and also persist all the data.
-func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockchain.BatchPin, signingIdentity string, protocolTxID string, additionalInfo fftypes.JSONObject) error {
+func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockchain.BatchPin, author string, protocolTxID string, additionalInfo fftypes.JSONObject) error {
```
awrichar, line 36: For consistency in terminology, isn't this what we're calling a "key" (ie an Ethereum address)? Seems like the word "author" has evolved to mean a higher level FireFly/DID string.

internal/broadcast/definition.go
```diff
@@ -24,21 +24,65 @@ import (
...
 )
 
-func (bm *broadcastManager) broadcastDefinitionAsNode(ctx context.Context, def fftypes.Definition, tag fftypes.SystemTag, waitConfirm bool) (msg *fftypes.Message, err error) {
-	signingIdentity, err := bm.GetNodeSigningIdentity(ctx)
+func (bm *broadcastManager) BroadcastDefinitionAsNode(ctx context.Context, def fftypes.Definition, tag fftypes.SystemTag, waitConfirm bool) (msg *fftypes.Message, err error) {
+	return bm.BroadcastDefinition(ctx, def, &fftypes.Identity{ /* resolve to node default */ }, tag, waitConfirm)
+}
+
+// THIS IS WHERE THE OLD CODE FIRST CALLS ResolveInputIdentity
+// It does NOT have an author and key the first time
```
awrichar, line 32: This comment feels like it can be removed. Don't think there's any need to reference OLD CODE vs NEW CODE, but we should note if there's a specific contract coming into a method (ie the identity must/must not be resolved when calling this).

internal/orchestrator/status.go
```diff
@@ -25,13 +25,17 @@ import (
 
 func (or *orchestrator) GetStatus(ctx context.Context) (status *fftypes.NodeStatus, err error) {
 
+	orgKey := config.GetString(config.OrgKey)
+	if orgKey == "" {
+		orgKey = config.GetString(config.OrgIdentityDeprecated)
```
awrichar, line 30: Should we have a helper for getting the org key, which checks both the new and deprecated key? Seems like this pattern is repeated in a few places.

internal/privatemessaging/privatemessaging_test.go
```diff
@@ -100,6 +96,12 @@ func TestDispatchBatchWithBlobs(t *testing.T) {
 		}
 	}
 
+	mim.On("ResolveInputIdentity", pm.ctx, mock.MatchedBy(func(identity *fftypes.Identity) bool {
+		assert.Equal(t, "org1", identity.Author)
+		identity.Key = "0x12345"
```
awrichar, line 101: This fields weird to be altering an arg within a `MatchedBy` method. Should this be a `.Run()` side-effect instead?

internal/privatemessaging/recipients.go
```diff
@@ -105,6 +106,17 @@ func (pm *privateMessaging) resolveNode(ctx context.Context, org *fftypes.Organi
 }
 
 func (pm *privateMessaging) getReceipients(ctx context.Context, in *fftypes.MessageInOut) (gi *fftypes.GroupIdentity, err error) {
```
awrichar, line 108: Not your change, but there's a typo in this method name.

internal/events/persist_batch.go
```diff
@@ -18,23 +18,31 @@ package events
...
 
-	// Verify the author matches
-	id, err := em.identity.Resolve(ctx, batch.Author)
+	// Verify that we can resolve the signing key back to this identity.
+	// This is a specific rule for broadcasts, so we know the authenticity of the data.
+	author, err := em.identity.ResolveSigningKeyIdentity(ctx, signingKey)
 	if err != nil {
-		l.Errorf("Invalid batch '%s'. Author '%s' cound not be resolved: %s", batch.ID, batch.Author, err)
-		return false, nil // This is not retryable. skip this batch
+		return false, err
```
awrichar, line 35: Did this become retryable?

internal/events/persist_batch.go
```diff
@@ -47,6 +55,29 @@ func (em *eventManager) persistBatchFromBroadcast(ctx context.Context /* db TX c
 	return valid, err
 }
 
+func (em *eventManager) isRootOrgBroadcast(batch *fftypes.Batch) (bool, error) {
```
awrichar, line 58: This feels...weird. Should we be tagging these messages in some special way? Unpacking all the data of every message in every batch to look for this special situation just feels very brute-force.

internal/broadcast/definition.go
```diff
@@ -24,21 +24,65 @@ import (
...
+		Data: fftypes.DataRefs{
+			{ID: data.ID, Hash: data.Hash},
+		},
+	}
+
+	// Broadcast the message
+	return bm.broadcastMessageCommon(ctx, msg, waitConfirm, false)
+}
+
+func (bm *broadcastManager) BroadcastRootOrgDefinition(ctx context.Context, def *fftypes.Organization, signingIdentity *fftypes.Identity, tag fftypes.SystemTag, waitConfirm bool) (msg *fftypes.Message, err error) {
```
awrichar, line 82: This method is nearly 100% identical to the method above - can we find a way to commonize that and make it clearer which pieces are special?

internal/networkmap/register_node.go
```diff
@@ -26,15 +27,23 @@ import (
...
-		Owner:       config.GetString(config.OrgIdentity),
+		Owner:       localOrgSigningKey, // TODO: Switch hierarchy to DID based, not signing key. Introducing an intermediate identity object
 		Name:        config.GetString(config.NodeName),
 		Description: config.GetString(config.NodeDescription),
 	}
 	if node.Name == "" {
-		node.Name = config.GetString(config.OrgIdentity)
+		orgName := config.GetString(config.OrgName)
+		if orgName != "" {
+			node.Name = fmt.Sprintf("%s.node", config.GetString(config.OrgName))
```
awrichar, line 45: Can reuse `orgName` here instead of fetching again from config.

awrichar (commented):
I left a lot of comments, but they're mostly in the category of cleanup rather than functional changes.

I think the functionality looks good, although I'll confess it's difficult to follow exactly what the new rules are for exactly when identities are resolved in each possible flow (some documentation to lay those down would be helpful). I'm relying largely on the E2E tests to assert that no major flows have been broken here =)

db/migrations/sqlite/000027_create_key_fields.up.sql
peterbroadhurst: Yeah, that's I conclusion I came to.

test/e2e/e2e_test.go
peterbroadhurst: k, seems this needs to be solved before merge

internal/identity/identitymanager.go
peterbroadhurst: I didn't find clear docs on this (the code you describe is elsewhere in FireFly).

Digging into the code here...
https://github.com/karlseguin/ccache/blob/ef4bd54683f611ad3df9ba84689234c78243c464/cache.go#L101-L117

Then here...
https://github.com/karlseguin/ccache/blob/ef4bd54683f611ad3df9ba84689234c78243c464/cache.go#L394-L410

It seems the behavior if you _don't_ extend, is that it will end up in the purge list, for LRU deletion.
While `Get` alone will make it less likely to be the thing that's deleted, that's not the same as ensuring it's retained for a period after being explicitly accessed.

I guess my vote is to leave the pattern as it is for the moment in this PR, then consider a wrapper around the caching framework in a separate PR that handles optimization of the logic.

internal/identity/identitymanager.go
peterbroadhurst: I do think this is incompletely thought through for Tokens in this PR.
My suggestion is to address the token specifics in #215

internal/assets/manager.go
peterbroadhurst: Agreed. Let's accept that this PR leaves things in a "still broken" state before+after for tokens, and address in #215 after merge.

internal/batchpin/batchpin.go
peterbroadhurst: Yes. It comes from the messages in the batch which must each have been resolved, and will be verified by all members processing the batch, so doing an additional pre-submission check here does not seem efficient.

internal/broadcast/message.go
peterbroadhurst: Agreed, pulling it out. Think this became unnecessary in some of the updates @nguyer made

internal/broadcast/message.go
peterbroadhurst: Think this is because the `BroadcastRootOrgDefinition` method was separated out, which is missing coverage.
Addressing this.

peterbroadhurst (commented):
Submitted PR for coverage and code duplication, continuing review.

internal/orchestrator/status.go
peterbroadhurst: helper, mock, and test updates added

internal/privatemessaging/privatemessaging_test.go
peterbroadhurst: changed

internal/privatemessaging/recipients.go
peterbroadhurst: fixed

internal/syscore/syscore.go
peterbroadhurst: removed 👍

test/e2e/e2e_test.go
awrichar: If it makes sense for me to be the one that does this, I'm happy to stick a commit on the end after other elements are stabilized.

internal/identity/identitymanager.go
awrichar: Yes, fine by me

peterbroadhurst (commented):
@awrichar  - believe I've addressed all the comments now.

db/migrations/postgres/000030_create_key_fields.up.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+
+ALTER TABLE batches ADD "key" VARCHAR(1024);
+UPDATE batches SET "key" = "";
```
jimthematrix, line 4: the empty string should be delimited by single quotes instead of double quotes, otherwise you'd get the following error during startup:

```
FF10163: Database migration failed: FF10163: Database migration failed: migration failed: zero-length delimited identifier at or near """" (column 28) in line 4: BEGIN;

ALTER TABLE batches ADD "key" VARCHAR(1024);
UPDATE batches SET "key" = "";
ALTER TABLE batches ALTER COLUMN "key" SET NOT NULL;

ALTER TABLE messages ADD "key" VARCHAR(1024);
UPDATE messages SET "key" = "";
ALTER TABLE messages ALTER COLUMN "key" SET NOT NULL;

ALTER TABLE tokenpool ADD "key" VARCHAR(1024);
UPDATE tokenpool SET "key" = "";
ALTER TABLE tokenpool ALTER COLUMN "key" SET NOT NULL;
```

db/migrations/sqlite/000030_create_key_fields.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE batches ADD COLUMN "key" VARCHAR(1024);
+UPDATE batches SET "key" = "";
```
jimthematrix, line 2: same as above, single quotes instead of double quotes

jimthematrix (changes requested):

pkg/fftypes/identity.go
```diff
@@ -16,10 +16,9 @@
...
 
-// Identity is a structure used to keep track of and map identity in the system.
-//
-// TODO: Mapping of more sophisticate identities (DIDs etc.) via plugins, and richer interface
+// Identity is the nested structure representing an identity, that might comprise a resolvable
+// by FireFly identity DID, a blockchain signing key, or both.
 type Identity struct {
-	Identifier string `json:"identifier,omitempty"`
-	OnChain    string `json:"onchain,omitempty"`
+	Author string `json:"author,omitempty"`
```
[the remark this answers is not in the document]
jimthematrix: never mind this one, was from an earlier thought (days ago) that has since been discarded, but neglected to delete the comment

PR #197, opened 2021-09-21
Replace operation "info" with separate "input" and "output" fields

awrichar:
Existing field has evolved to always contain plugin-specific data generated upon
completion of the operation. There is also a (separate) use case for storing
input data upon creating the operation, to support e.g. retryability.


peterbroadhurst (approved):
📥 ⚙️ 📤

PR #201, opened 2021-09-22
Add additional fields to TokenPool

awrichar:
All token pools should record the connector that created them, may optionally
include a token symbol, and may (in the near future) be linked to a message.


peterbroadhurst (approved):
🔣  🔌   ✉️

PR #202, opened 2021-09-22
Rename tokens "https" plugin to "fftokens"

awrichar:
The name "https" is too generic - the plugin represents a way to map token operations
like pool, mint, and transfer into a specific HTTP+websocket protocol. Therefore a
specific, unique name is appropriate.

Also changed the config key from "connector" to "plugin" (since its purpose is to
affect which plugin FireFly loads from the map).

Added migration paths for both changes so that the CLI and any existing stacks can
have time to be updated.


peterbroadhurst (approved):
🪙 `FFT`

PR #203, opened 2021-09-23
Use Operation ID for tracking ethconnect requests

awrichar:
The Transaction ID is longer-lived, while Operation ID is more appropriate for
a single request (since a transaction may contain multiple operations).

Signed-off-by: Andrew Richardson <andrew.richardson@kaleido.io>


awrichar:
E2E test failure is fixed by https://github.com/hyperledger/firefly/pull/204 (will update after that is merged)

awrichar:
Updated flow of the operations and transactions involved in pool creation.
PXL_20210924_205830070

internal/blockchain/ethereum/ethereum_test.go
```diff
@@ -47,6 +47,10 @@ func resetConf() {
 	e.InitPrefix(utConfPrefix)
 }
 
+func uuidMatches(id1 *fftypes.UUID) interface{} {
```
peterbroadhurst, line 50: we've got a safe equals built into UUID you can use (by having a pointer receiver, it works with `nil` pointers on either end - a nice feature of Go)
https://github.com/hyperledger/firefly/blob/25c1039c46820888c2f43eb0bce07feaa4c27fea/pkg/fftypes/uuid.go#L83-L92

peterbroadhurst (approved):
👍  - thanks for all the re-drawings of the above diagram needed to bring us some order here. 
Think this work and thinking will make a big difference as we go forwards.

Approving with one note for consideration, and also the fact you'll need a merge.

internal/blockchain/ethereum/ethereum_test.go
awrichar: I copied this helper from some other tests - seems like it can be removed everywhere.

Issue #205, opened 2021-09-23
Data endpoint requires autometa=true


nguyer:
Right now, if someone attempts to upload a file to the `/namespaces/{namespaceId}/data` endpoint, and only sets the `file` field, the request will fail with the following response:

```
HTTP 500 Internal Server Error
```
```json
{
    "error": "FF10116: Database insert failed: NOT NULL constraint failed: data.value"
}
```

As a workaround it is possible to get this to work by adding another field to the form called `autometa` and setting its value to be `true`. This upload should work without needing to set this field though.

peterbroadhurst:
You can also set `metadata` to be any valid JSON

peterbroadhurst:
Fixed in #207 to allow a `null` value

Issue #198, opened 2021-09-21
Add tags to /data endpoint


lil131:
It would be great if `tag` could be added to `/data` endpoint, like the way it works on `/message`. It will be easier to categorize and retrieve data.

panghalamit:
hd

peterbroadhurst:
Hey @lil131,
The intended design is the `datatype.name` and `datatype.version` fields can be used for exactly this purpose.

If you specify `validator` as `json` then this `datatype` must have been broadcast to the network, with a JSON Schema that can be used to verify the data.

However, if you leave `validator` as blank, the data should not be verified at all - and you can simply use the `datatype` for classification of your data (without FireFly enforcing structure on it).

If you find a difference in behaviour from this, please say - as I know most of the testing / UTs so far have focussed on:
- JSON without a datatype set at all
- JSON with a verified datatype
- Binary blobs without a datatype set at all

... and I think you're looking for:
- Binary blobs with a datatype set

... or mabye:
- JSON with a datatype set, but without requiring FireFly to verify it

lil131:
Thanks @peterbroadhurst! I'll try those on

nguyer:
I tested this and it does not work currently

peterbroadhurst:
Ok - looked into this some more.

Every `data` resource can have:
- An optional JSON `value`
- An optional `blob` reference
- _Or both_

https://github.com/hyperledger/firefly/blob/dcff6111530b683f7794bbf5e8ba0cc4d5652607/pkg/fftypes/data.go#L45-L46

The problem here is that the `verifier` defaults to `json`, and there is no way to specify `none`.
So if you specify a `datatype`, you are instructing the code to validate the `value` part of the data as JSON according to the schema, so the `datatype` must exist.

This prevents you from using the `datatype.name`/`datatype.value` fields for indexed categorization of the data, _without_ performing JSON schema verification.

Per this issue, that's a very valid requirement - so I plan to add `none` validator type, and also the tests for this scenario.

PR #207, opened 2021-09-25
Allow null value with blob data, and allow "none" validator to tag datatype on unstructured data

peterbroadhurst:
Fix for #198 

- Fixes handling of `null` on `data.value` - including as part of a FORM post upload
  - Stored as the bytes `null` in the database (as today)
  - Does not contribute to the hash, when there is a blob
  - Returned as `null` in the JSON payload on `GET`
- Adds a new `validator` of `none`
  - Must be set explicitly - `json` remains the default
  - Disabled any checking of the `datatype.name`/`datatype.value` when set
  - Allows you to just use the `datatype` for tagging purposes, without validating the `value` JSON payload itself

### DIFF
--- a/pkg/fftypes/datatype.go
+++ b/pkg/fftypes/datatype.go
@@ -27,6 +27,8 @@ type ValidatorType = FFEnum
 var (
 	// ValidatorTypeJSON is the validator type for JSON Schema validation
 	ValidatorTypeJSON ValidatorType = ffEnum("validatortype", "json")
+	// ValidatorTypeNone explicitly disables validation, even when a datatype is set. Allowing cateogration of datatype without validation.
+	ValidatorTypeNone ValidatorType = ffEnum("validatortype", "none")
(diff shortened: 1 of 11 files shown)


peterbroadhurst:
Fixes #205

pkg/fftypes/datatype.go
```diff
@@ -27,6 +27,8 @@ type ValidatorType = FFEnum
 var (
 	// ValidatorTypeJSON is the validator type for JSON Schema validation
 	ValidatorTypeJSON ValidatorType = ffEnum("validatortype", "json")
+	// ValidatorTypeNone explicitly disables validation, even when a datatype is set. Allowing cateogration of datatype without validation.
```
nguyer, line 30:
```suggestion
	// ValidatorTypeNone explicitly disables validation, even when a datatype is set. Allowing categorization of datatype without validation.
```

PR #210, opened 2021-09-27
Move TokenPoolCreated handler into AssetManager

awrichar:


internal/assets/manager.go
```diff
@@ -59,6 +67,12 @@ func NewAssetManager(ctx context.Context, di database.Plugin, ii identity.Plugin
 		data:      dm,
 		syncasync: sa,
 		tokens:    ti,
+		retry: retry.Retry{
+			InitialDelay: config.GetDuration(config.EventAggregatorRetryInitDelay),
+			MaximumDelay: config.GetDuration(config.EventAggregatorRetryMaxDelay),
```
peterbroadhurst, line 72: Think this should be a dedicated retry config under `assets.manager`

peterbroadhurst (approved):
One comment to consider before merging @awrichar

internal/assets/manager.go
awrichar: Yea, kinda thought that as I was copy-pasting it 😇 I'll add a new config section

PR #212, opened 2021-09-27
Add "created" field to token pools, add some extra E2E checks

awrichar:


internal/database/sqlcommon/tokenpool_sql.go
```diff
@@ -96,6 +97,7 @@ func (s *SQLCommon) UpsertTokenPool(ctx context.Context, pool *fftypes.TokenPool
 			return err
 		}
 	} else {
+		pool.Created = fftypes.Now()
```
awrichar, line 100: Where is the appropriate layer to set `created` timestamps? It seems to vary across different objects, but here in the Upsert method felt most logical/efficient to me.

docs/swagger/swagger.yaml
```diff
@@ -4223,6 +4223,11 @@ paths:
         schema:
           default: 120s
           type: string
+      - description: 'Data filter field. Prefixes supported: > >= < <= @ ^ ! !@ !^'
```
nguyer, line 4226: Is there somewhere we can indicate in the docs/swagger that this is a date/time? I had to read further into the source code to see where it gets converted to tell what type it actually was.

nguyer (approved):
One minor suggestion, otherwise looks good

internal/database/sqlcommon/tokenpool_sql.go
peterbroadhurst: Looks appropriate to me 👍

docs/swagger/swagger.yaml
awrichar: Probably not. The Swagger spec still needs a lot of fleshing out in general... I've noted before that the Swagger generator could almost be a separate project in itself. But I filed https://github.com/hyperledger/firefly/issues/214 for this.

PR #215, opened 2021-09-29
Add database types and route for listing token transfers

awrichar:


awrichar:
E2E tests currently failing due to dependency on https://github.com/hyperledger/firefly-tokens-erc1155/pull/24

internal/assets/tokens_transferred.go
```diff
@@ -0,0 +1,66 @@
...
+				return err
+			}
+			if pool == nil {
+				log.L(ctx).Warnf("Token transfer received for unknown pool '%s' - ignoring: %s", transfer.PoolProtocolID, protocolTxID)
+				return nil
+			}
+			if err := am.database.UpsertTokenTransfer(ctx, transfer); err != nil {
+				log.L(ctx).Errorf("Failed to record token transfer '%s': %s", transfer.ProtocolID, err)
+				return err
+			}
```
awrichar, line 41: The logic below is relying strongly on the transactional nature of the database in order to end up with the correct account balances. Just want to confirm that's OK.

The ERC1155 plugin actually allows balance querying, so we could include the ending balance with each transfer event... but that seems more complex on the plugin side and probably won't be possible with other plugins, e.g. a UTXO model.

db/migrations/postgres/000030_create_tokentransfer_table.up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+CREATE TABLE tokentransfer (
+  seq              SERIAL          PRIMARY KEY,
+  local_id         UUID            NOT NULL,
+  type             VARCHAR(64)     NOT NULL,
+  pool_protocol_id VARCHAR(1024)   NOT NULL,
+  token_index      VARCHAR(1024)   NOT NULL,
+  from_identity    VARCHAR(1024)   NULL,
```
peterbroadhurst, line 8: Are these really `identity`, or are they actually `key`?

... I know we're broken in `organization` at the moment, where the field is `identity`, but the contents are a `key` - and that needs to be fixed. I'm just concerned about propagating that problem any further.

internal/assets/manager.go
```diff
@@ -168,25 +173,147 @@ func (am *assetManager) GetTokenPools(ctx context.Context, ns string, typeName s
...
+	return am.transferTokensWithID(ctx, fftypes.NewUUID(), ns, typeName, poolName, transfer, waitConfirm)
+}
+
+func (am *assetManager) BurnTokens(ctx context.Context, ns, typeName, poolName string, transfer *fftypes.TokenTransfer, waitConfirm bool) (*fftypes.TokenTransfer, error) {
+	transfer.Type = fftypes.TokenTransferTypeBurn
+	if transfer.Author == "" {
+		transfer.Author = config.GetString(config.OrgIdentity)
+	}
+	if transfer.From == "" {
+		transfer.From = transfer.Author
```
peterbroadhurst, line 230: Again want to be sure this is `Author` here, rather than `Key`.
My understanding was the tokens features weren't going to _require_ any kind of registration of identity, so would support raw keys only. The `Author` field would be used consistently where if it's resolvable to a registered key, that would work... but we'd also support indexing transactions where we don't have an author.

Holding on the rest of the review until you've had a chance to think about this feedback, as I think it might be significant.

peterbroadhurst (commented):
... see notes about identity mapping. Want to make sure we've got our ducks in a row on that item, before going any further with the review.

awrichar:
@peterbroadhurst thanks for the initial look. I was already certain that my handling of identities was wrong, so the comments are helpful to steer my thinking a bit. I'll come up to speed on where #192 is at - agree we probably want to get that shaped up for merge before this propagates the existing problems further.

peterbroadhurst:
Thanks @awrichar I'll add a commit as such to this branch, and ask @jimthematrix for a review before merging.

peterbroadhurst:
Latest error from tokens connector after https://github.com/hyperledger/firefly-tokens-erc1155/pull/32:
```
[34;1mtokens_0_1        |[0m [32m[Nest] 17  - [39m10/11/2021, 3:52:08 PM [32m    LOG[39m [38;5;3m[RequestLogging] [39m[32mPOST /api/v1/init[39m
[34;1mtokens_0_1        |[0m [31m[Nest] 17  - [39m10/11/2021, 3:52:08 PM [31m  ERROR[39m [38;5;3m[ExceptionsHandler] [39m[31mrxjs_1.lastValueFrom is not a function[39m
[34;1mtokens_0_1        |[0m TypeError: rxjs_1.lastValueFrom is not a function
[34;1mtokens_0_1        |[0m     at RouterResponseController.transformToResult (/root/node_modules/@nestjs/core/router/router-response-controller.js:32:27)
[34;1mtokens_0_1        |[0m     at /root/node_modules/@nestjs/core/router/router-execution-context.js:173:52
[34;1mtokens_0_1        |[0m     at /root/node_modules/@nestjs/core/router/router-execution-context.js:47:19
[34;1mtokens_0_1        |[0m     at processTicksAndRejections (internal/process/task_queues.js:95:5)
[34;1mtokens_0_1        |[0m     at async /root/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

awrichar:
BigInt changes look great @peterbroadhurst; thanks for handling this and the piece in the tokens connector.

PR #216, opened 2021-09-29
Don't export internal "...WithID" methods

awrichar:


nguyer (approved):
Less is more. So they say.

PR #223, opened 2021-10-01
Fixed endpoint typo to send blob

scaraclette:
Signed-off-by: scaraclette <gustiscarlett@gmail.com>


nguyer (approved):
Thank you! :clap: I saw this the other day, and just hadn't had a chance to fix it yet. Really appreciate it!

PR #227, opened 2021-10-04
Add wsclient to public API

shorsher:
Moving `wsclient` to FireFly core's public API, to be potentially be leveraged by other plugins, sdk's, etc.

- Moved `wsclient` under `pkg` instead of `internal`
- Created a public `WSConfig` struct which defines websocket configuration

Created a new internal `wsconfig` package, which contains FF core specific config options and maps `config.Prefix` to `WSConfig`


internal/blockchain/ethereum/ethereum.go
```diff
@@ -131,10 +132,13 @@ func (e *Ethereum) Init(ctx context.Context, prefix config.Prefix, callbacks blo
 		GlobalSequencer: true,
 	}
 
-	if ethconnectConf.GetString(wsclient.WSConfigKeyPath) == "" {
-		ethconnectConf.Set(wsclient.WSConfigKeyPath, "/ws")
+	if ethconnectConf.GetString(wsconfig.WSConfigKeyPath) == "" {
+		ethconnectConf.Set(wsconfig.WSConfigKeyPath, "/ws")
```
awrichar, line 136: Should we do this here, or on the generated `wsConfig` object..? probably doesn't matter either way, just pointing out that we _could_ move it down.

awrichar (approved):

PR #231, opened 2021-10-07
Capture logs as artifacts even if e2e fails (simulated failure in this commit)

peterbroadhurst:


peterbroadhurst:
ok - test run went great.
Showed failure of the tests, and still collected logs: https://github.com/hyperledger/firefly/runs/3828183867

So I've removed the intentionally failing test, and this is ready for review.

PR #233, opened 2021-10-07
fix wsclient mocks

shorsher:
- artifact left over from wsclient migration to `pkg`


peterbroadhurst:
Looks like e2e test failures could be reproduced without this change https://github.com/hyperledger/firefly/pull/234/checks?check_run_id=3838443338
Investigating.

PR #238, opened 2021-10-08
Added a temporary copy of solidity_firefly due to external dependencies

jimthematrix:
Signed-off-by: Jim Zhang <jim.zhang@kaleido.io>


peterbroadhurst:
I believe the external project that is relying on this, is going to update (rather than requiring an update to the OSS).

Issue #218, opened 2021-09-29
Token support


awrichar:
FireFly should support the creation and tracking of both fungible and non-fungible tokens. Some of the key concepts that must be supported:
* Creating multiple types of tokens
* Minting, transferring, and burning tokens of a given type
* Tracking transaction history and current account balances (must include operations initiated by FireFly and also those initiated on-chain with no trigger from FireFly)
* Attaching data to a token transfer, for correlating off-chain items with the transfer event on-chain

The first connector will be focused on Ethereum tokens support via ERC1155. Source code for the connector is here: https://github.com/hyperledger/firefly-tokens-erc1155

This functionality is part of milestone #117.

Cross-chain functionality, including atomic swaps and the like, will not be part of this design effort (but will be part of a future item).

awrichar:
This tracking issue is a bit late to the game, but tokens have been discussed in detail on these two community calls:
https://wiki.hyperledger.org/display/FIR/2021-06-30+FireFly+Community+Call
https://wiki.hyperledger.org/display/FIR/2021-09-15+FireFly+Community+Call

Below are a few slides from those discussions that informed the initial design:
![FireFly Architecture Episode 2 pptx](https://user-images.githubusercontent.com/1993829/135339774-5a821db0-0a29-4461-924e-5cd4a20fc2da.png)
![FireFly Architecture Episode 2 pptx (1)](https://user-images.githubusercontent.com/1993829/135339840-51a25d52-2ba2-41c1-8469-c9b8b4b0cf81.png)
![FireFly - Tokens Update](https://user-images.githubusercontent.com/1993829/135339875-d90ecaf4-fd58-4b17-88f7-fddde607df84.png)

awrichar:
PRs merged to date:
#154 Add support for creating token pools
#162 Add "confirm" parameter when creating token pools
#163 Reject token pool events with duplicate info
#165 Rename token fields to match backend changes
#166 Add GET routes for token pools
#167 Handle token_pool_rejected, add new transaction type for token pools
#173 Add unit tests for tokens HTTPS plugin
#176 Align the "additionalInfo" for blockchain and token transactions
#177 Read "operator" field instead of "author" for token-pool event
#178 Add skeleton functionality for token accounts
#179 Supply requestId when creating token pools
#185 Pack all token pool info into "data" field
#193 Small fixes to token accounts
#201 Add additional fields to TokenPool
#202 Rename tokens "https" plugin to "fftokens"
[...]

PR #297, opened 2021-10-26
Only send transfer message if token transfer is successful

awrichar:
This is a rework of #279 - please see the original notes on that PR, which still apply here.

Falls under architecture item #218.

~~It contains the commits from #283 and #296 as pre-requisites, so those should be reviewed first. Only the last 2 commits are net new code.~~


peterbroadhurst (approved):
👍  marking approval here to allow merging when the spelling change and. merge conflicts are done.

PR #306, opened 2021-10-28
Add "Getting Started" documentation for tokens

awrichar:
Initial docs for #218.

~~Some of the routes described here are still being updated by @eberger727 in #307, so should wait for that PR before merging this.~~

I've published the docs in this PR to my own FireFly fork here, if you like reading the pretty pages instead of the Markdown: https://awrichar.github.io/firefly/gettingstarted/mint_tokens.html

### DIFF
--- a/docs/gettingstarted/mint_tokens.md
+++ b/docs/gettingstarted/mint_tokens.md
@@ -109,13 +109,14 @@ Other parameters:
 All transfers (as well as mint/burn operations) support an optional `message` parameter that contains a broadcast or private
 message to be sent along with the transfer. This message follows the same convention as other FireFly messages, and may be comprised
 of text or blob data, and can provide context, metadata, or other supporting information about the transfer. The message will be
-batched, hashed, and pinned to the primary blockchain as described in [key concepts](/keyconcepts/broadcast.html).
+batched, hashed, and pinned to the primary blockchain as described in [key concepts](../keyconcepts/broadcast.html).
 
-The message hash will also be sent to the token connector as part of the transfer operation, to be written to the token blockchain
+The message ID and hash will also be sent to the token connector as part of the transfer operation, to be written to the token blockchain
 when the transaction is submitted. All recipients of the message will then be able to correlate the message with the token transfer.
 
 `POST` `/api/v1/namespaces/default/tokens/transfers`
 
+Broadcast message:
 ```json
 {
   "amount": 1,
@@ -128,9 +129,29 @@ when the transaction is submitted. All recipients of the message will then be ab
 }
 ```
 
-By default, a broadcast message is used. In order to send a private message, specify `"type": "transfer_private"` in the message header,
-and include a `"group"` entry specifying the recipients. All parties in the network will be able to see the transfer (including the
-message hash), but only the recipients of the message will be able to view the actual message data.
+Private message:
+```json
+{
+  "amount": 1,
+  "to": "0x07eab7731db665caf02bc92c286f51dea81f923f",
+  "message": {
+    "header": {
+      "type": "transfer_private",
+    },
+    "group": {
+      "members": [{
+          "identity": "org_1"
+      }]
+    },
+    "data": [{
+      "value": "payment for goods"
+    }]
+  }
+}
+```
+
+Note that all parties in the network will be able to see the transfer (including the message ID and hash), but only
+the recipients of the message will be able to view the actual message data.
 
 ## Burn tokens
 
--- a/docs/gettingstarted/mint_tokens.md
+++ b/docs/gettingstarted/mint_tokens.md
@@ -16,6 +16,118 @@ nav_order: 9
 
 ---
 
-## Work in progress
+## Quick reference
 
-... the tokens APIs are under development
+- FireFly provides an abstraction layer for multiple types of tokens
+- Tokens are grouped into _pools_, which each represent a particular type or class of token
+- Each pool is classified as _fungible_ or _non-fungible_
+- In the case of _non-fungible_ tokens, the pool is subdivided into individual tokens with a unique _token index_
+- Within a pool, you may _mint_, _transfer_, and _burn_ tokens
+- Each operation can be optionally accompanied by a broadcast or private message, which will be recorded alongside the transfer on-chain
+- FireFly tracks a history of all token operations along with all current token balances
+- The blockchain backing each token connector may be the same _or_ different from the one backing FireFly message pinning
+
+## Create a pool
+
+Every application will need to create at least one token pool. At a minimum, you must always
+specify a `name` and `type` (fungible or nonfungible) for the pool.
+
+`POST` `/api/v1/namespaces/default/tokens/pools`
+
+```json
+{
+  "name": "testpool",
+  "type": "fungible"
+}
+```
+
+Other parameters:
+- You must specify a `connector` if you have configured multiple token connectors
+- You may pass through a `config` object of additional parameters, if supported by your token connector
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+
+## Mint tokens
+
+Once you have a token pool, you can mint tokens within it. With the default `firefly-tokens-erc1155` connector,
+only the creator of a pool is allowed to mint - but each connector may define its own permission model.
+
+`POST` `/api/v1/namespaces/default/tokens/mint`
+
+```json
+{
+  "amount": 10
+}
+```
+
+Other parameters:
+- You must specify a `pool` name if you've created more than one pool
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+- You may specify `to` if you'd like to send the minted tokens to a specific identity (default is the same as `key`)
+
+## Transfer tokens
+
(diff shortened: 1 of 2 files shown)


docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
 
 ---
 
-## Work in progress
+## Quick reference
 
-... the tokens APIs are under development
\ No newline at end of file
+- FireFly provides an abstraction layer for multiple types of tokens
+- Tokens are grouped into _pools_, which each represent a particular type or class of token
```
peterbroadhurst, line 22: Wonder if an example worthwhile here, such as?

Examples of how a pool might be used by a given token connector implementation:
    - ERC-721: A contract instance, ready for issuance of NFTs within unique identities in the pool
    - ERC-20 / ERC-777: A token contract instance, with a single fungible pool of value - "a coin"
    - ERC-1155: An efficient to allocate isolated pool of fungible or non-fungible tokens (see [ref implementation](https://github.com/hyperledger/firefly-tokens-erc1155))
    - ERC-1400 (and related security token standards): An partition within an overall fungible pool, representing part of the value of an overall non-fungible asset

Could do with @jimthematrix checking on the terminology above

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
+  "name": "testpool",
+  "type": "fungible"
+}
+```
+
+Other parameters:
+- You must specify a `connector` if you have configured multiple token connectors
+- You may pass through a `config` object of additional parameters, if supported by your token connector
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+
```
peterbroadhurst, line 48: Worth noting that individual connector implementations an extend this without any updates to FireFly core, by using an opaque configuration setting. For example a token connector could allow an initial supply to be specified, or configuration options specifying transfer restriction criteria.

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
+  "amount": 1,
+  "to": "0x07eab7731db665caf02bc92c286f51dea81f923f"
+}
+```
+
+Other parameters:
+- You must specify a `pool` name if you've created more than one pool
+- You must specify a `tokenIndex` for non-fungible pools (and the `amount` should be 1)
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+- You may specify `from` if you'd like to send tokens from a specific identity (default is the same as `key`)
```
peterbroadhurst, line 86: @jimthematrix - note that `from` here is  the source account from which the funds should originate.
Whereas `key` is the signing key with which the transaction will be authorized by the on-chain contract.

I believe this is consistent throughout FireFly, but I wanted to call this out as the concrete is setting here, to check you agree this is acceptable and consistent.

peterbroadhurst (commented):
Really clear and straight forwards thanks @awrichar 

Couple of comments, and deferral to @jimthematrix for a final review on this one before it goes in.

awrichar:
Should I add a section on the FireFly configuration file? ie show the sample tokens config that is generated by the CLI?

docs/gettingstarted/mint_tokens.md
awrichar: Could be, and maybe not limited to pools... Maybe we need an extra section of "Example uses" to lay out a few examples of how you could map common token standards/use cases onto FireFly.

docs/gettingstarted/mint_tokens.md
awrichar: Yea, that is the `config` parameter mentioned above, but maybe we should give examples of that type of thing?

peterbroadhurst: Sorry I missed that, looks good for now to me

docs/gettingstarted/mint_tokens.md
jimthematrix: if we agree with the above, we should also add that the current API endpoints support the ERC20, ERC777, ERC1155 and ERC721 contracts, while future work is needed to support ERC1400, in particular to support partitions (metadata to fungible tokens) within pools.

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
-## Work in progress
+## Quick reference
 
-... the tokens APIs are under development
\ No newline at end of file
+- FireFly provides an abstraction layer for multiple types of tokens
+- Tokens are grouped into _pools_, which each represent a particular type or class of token
+- Each pool is classified as _fungible_ or _non-fungible_
+- In the case of _non-fungible_ tokens, the pool is subdivided into individual tokens with a unique _token index_
+- Within a pool, you may _mint_, _transfer_, and _burn_ tokens
```
jimthematrix, line 25: suggest adding operational aliases (issue vs. mint, redeem vs. burn) used in different token specs:

Within a pool, you may _mint_ **(_issue_)**, _transfer_, and _burn_ **(_redeem_)** tokens

docs/gettingstarted/mint_tokens.md
jimthematrix: thanks, I think it's consistent with the lingo used by the various token specs (my earlier concern was that it's unfortunate that the transfer source account is designated as `from` which is also the parameter name of web3 interface's tx signing account, same for `to`. but I think this usage is part of the widely adopted convention now)

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
+  "to": "0x07eab7731db665caf02bc92c286f51dea81f923f",
+  "message": {
+    "data": [{
+      "value": "payment for goods"
+    }]
+  }
+}
+```
+
+By default, a broadcast message is used. In order to send a private message, specify `"type": "transfer_private"` in the message header,
```
jimthematrix, line 112: a sample payload for a private message would be useful here

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
+
+Other parameters:
+- You must specify a `pool` name if you've created more than one pool
+- You must specify a `tokenIndex` for non-fungible pools (and the `amount` should be 1)
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+- You may specify `from` if you'd like to send tokens from a specific identity (default is the same as `key`)
+
+## Sending data with a transfer
+
+All transfers (as well as mint/burn operations) support an optional `message` parameter that contains a broadcast or private
```
jimthematrix, line 90: I think we effectively hijacked the `_data` field of the various operations that may have been needed for further on-chain authorizations (e.g signatures by parties other than the sender to indicate approval). as a mental note we should consider supporting an additional payload parameter to include into the `_data` field as-is

jimthematrix (changes requested):
Nice writeup. Added my opinion on support for ERC-1400, and other relatively minor comments. Happy to discuss more if there are different thoughts.

docs/gettingstarted/mint_tokens.md
```diff
@@ -16,6 +16,118 @@ nav_order: 9
...
+- You must specify a `connector` if you have configured multiple token connectors
+- You may pass through a `config` object of additional parameters, if supported by your token connector
+- You may specify a `key` understood by the connector (i.e. an Ethereum address) if you'd like to use a non-default signing identity
+
+## Mint tokens
+
+Once you have a token pool, you can mint tokens within it. With the default `firefly-tokens-erc1155` connector,
+only the creator of a pool is allowed to mint - but each connector may define its own permission model.
+
+`POST` `/api/v1/namespaces/default/tokens/mint`
```
nguyer, line 54: I think the token pool name part is missing from this path

nguyer: Okay actually I see now that it just works if there's only one pool defined. Does this URL break if there is more than one pool?

nguyer (approved):
Looks great. I ran through the instructions in Postman and it worked for me too. Jim has requested changes, but if he dismisses or approves after updates, I'm fine with this being merged. Thanks!

docs/gettingstarted/mint_tokens.md
awrichar: The pool name is not in the new paths (if there is only one pool the name is optional; if there is more than one pool, you pass it in the body). The old path with the pool name is deprecated.

docs/gettingstarted/mint_tokens.md
awrichar: @jimthematrix yes, you're correct. The token connector exposes a `data` field, and FireFly packs JSON into this field to identify messages and other metadata. But FireFly itself does not (currently) expose any way for the user to control or add to this field from the level of the FireFly APIs.

docs/gettingstarted/mint_tokens.md
awrichar: Added a new section with some of these examples.

docs/gettingstarted/mint_tokens.md
awrichar: added

jimthematrix (approved):

PR #309, opened 2021-10-29
Consume websocket events in "strong datatype" tests

### LINKED ISSUE #308: Tokens E2E test is confused by messaging events from previous tests
As highlighted in https://github.com/hyperledger/firefly/actions/runs/1397046718, the tokens E2E test can be confused by late event delivery from previous tests. Technically any test could experience this issue, depending on ordering - so each test needs to ensure its events are consumed before moving on.

awrichar:
Even though these tests use confirm=true, we must create a websocket
and wait on events to propagate (otherwise the events may propagate
later and confuse the next test).

Fixes #308


peterbroadhurst (approved):
👍  - mental note here, that maybe a standard setup/teardown might help in a future change

Issue #265, opened 2021-10-19
FireFly crashed after attempting to send on closed websocket channel


awrichar:
While running E2E tests, I observed that one of the FireFly nodes suddenly went down and did not come back up. Logs (below) reveal that FireFly attempted to invoke `wc.send` from `dispatchChangeEvent` after the websocket was already closed, which triggered a fatal panic.

```
[2021-10-19T17:34:31.493Z] ERROR Read failed: websocket: close 1006 (abnormal closure): unexpected EOF pid=1 websocket=b7cd0598-f799-4a6b-a96c-c72b578665d0
[2021-10-19T17:34:31.493Z] DEBUG Sender closing pid=1 websocket=b7cd0598-f799-4a6b-a96c-c72b578665d0
[2021-10-19T17:34:31.493Z] DEBUG SQL<- commit dbtx=NHUhdPFo pid=1 role=aggregator
[2021-10-19T17:34:31.494Z] DEBUG Closing 1 dispatcher(s) for connection 'b7cd0598-f799-4a6b-a96c-c72b578665d0' pid=1
panic: send on closed channel

goroutine 834 [running]:
github.com/hyperledger/firefly/internal/events/websockets.(*websocketConnection).send(0xc0004be5a0, 0xc89840, 0xc0000e9410, 0x0, 0x0)
	/firefly/internal/events/websockets/websocket_connection.go:226 +0xce
github.com/hyperledger/firefly/internal/events/websockets.(*websocketConnection).dispatchChangeEvent(0xc0004be5a0, 0xc000101a40, 0xc000362780, 0x24)
	/firefly/internal/events/websockets/websocket_connection.go:180 +0x116
github.com/hyperledger/firefly/internal/events/websockets.(*WebSockets).ChangeEvent(0x1331000, 0xc000362780, 0x24, 0xc000101a40)
	/firefly/internal/events/websockets/websockets.go:96 +0xcc
github.com/hyperledger/firefly/internal/events.(*eventDispatcher).deliverEvents(0xc0002acc30)
	/firefly/internal/events/event_dispatcher.go:399 +0x617
created by github.com/hyperledger/firefly/internal/events.(*eventDispatcher).electAndStart
	/firefly/internal/events/event_dispatcher.go:147 +0x276
```

peterbroadhurst:
https://github.com/hyperledger/firefly/actions/runs/1720901977 hit this in https://github.com/hyperledger/firefly/pull/418 e2e

```
[...]

PR #312, opened 2021-11-01
Move definitions of datatypes and tokens to their proper namespace

nguyer:
Messages that use the `ff_system` namespace:

- Org registration
- Node registration
- Namespace definition

Messages that use whatever namespace is specified in the API request path:

- Everything else


internal/events/aggregator.go
```diff
@@ -428,7 +428,7 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *fftypes.M
 	valid := true
 	eventType := fftypes.EventTypeMessageConfirmed
 	switch {
-	case msg.Header.Namespace == fftypes.SystemNamespace:
+	case msg.Header.Namespace == fftypes.SystemNamespace || msg.Header.Type == fftypes.MessageTypeDefinition:
```
awrichar, line 431: Does syshandler now handle anything that is _not_ a definition?

test/e2e/tokens_test.go
```diff
@@ -50,6 +50,7 @@ func (suite *TokensTestSuite) TestE2EFungibleTokensAsync() {
 	}
 	CreateTokenPool(suite.T(), suite.testState.client1, pool, false)
 
+	<-received1
 	<-received1
```
awrichar, line 54: Could you add comments to notate why we get double events? (one is the pool creation, one is the pool announcement)

internal/events/aggregator.go
nguyer: I think it's _only ever_ handled definitions

awrichar: I wonder if it needs a different name 🙃

awrichar: And maybe the first part of the `if` clause here is irrelevant...

nguyer: Possibly. Probably worth some input from @peterbroadhurst on this one. I'll try to chat with him when he gets a chance. The rest of the code, including the unit tests seem to be written in a way that assumes there could be system broadcasts that are _not_ definitions. Maybe there is a case I'm not thinking about, possibly a future case. Or maybe the way the code was originally written is not relevant anymore.

awrichar (approved):

internal/events/aggregator.go
awrichar: Sounds good. I've approved since I think these changes are good - but probably need to have a chat about the future of what else may be on the ff_system namespace, and if "syshandler" is really the right name for the thing that is handling definitions.

nguyer: @peterbroadhurst I removed the first part of this case in the switch statement, which did change some failure code paths in the tests. I'll highlight those in the places that changed. But would appreciate your eyes on this to make sure the correct thing is still happening here.

internal/events/aggregator_test.go
```diff
@@ -979,11 +979,13 @@ func TestAttemptMessageDispatchFailValidateSystemFail(t *testing.T) {
...
-	msh := ag.syshandlers.(*syshandlersmocks.SystemHandlers)
-	msh.On("HandleSystemBroadcast", mock.Anything, mock.Anything, mock.Anything).Return(false, fmt.Errorf("pop"))
-
 	mdm := ag.data.(*datamocks.Manager)
 	mdm.On("GetMessageData", ag.ctx, mock.Anything, true).Return([]*fftypes.Data{}, true, nil)
+	mdm.On("ValidateAll", ag.ctx, mock.Anything).Return(false, nil)
+
+	dbm := ag.database.(*databasemocks.Plugin)
+	dbm.On("UpdateMessage", ag.ctx, mock.Anything, mock.Anything).Return(nil)
+	dbm.On("InsertEvent", ag.ctx, mock.Anything).Return(errors.Errorf("pop"))
```
nguyer, line 988: @peterbroadhurst This is one of the tests I had to change

internal/events/aggregator_test.go
```diff
@@ -936,12 +937,11 @@ func TestAttemptMessageDispatchFailValidateBadSystem(t *testing.T) {
...
 	defer cancel()
 
-	msh := ag.syshandlers.(*syshandlersmocks.SystemHandlers)
-	msh.On("HandleSystemBroadcast", mock.Anything, mock.Anything, mock.Anything).Return(false, nil)
-
 	mdm := ag.data.(*datamocks.Manager)
 	mdm.On("GetMessageData", ag.ctx, mock.Anything, true).Return([]*fftypes.Data{}, true, nil)
 
+	mdm.On("ValidateAll", ag.ctx, mock.Anything).Return(false, nil)
+
```
nguyer, line 944: @peterbroadhurst This is one of the tests I had to change

internal/events/aggregator.go
```diff
@@ -428,10 +428,10 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *fftypes.M
 	valid := true
 	eventType := fftypes.EventTypeMessageConfirmed
 	switch {
-	case msg.Header.Namespace == fftypes.SystemNamespace:
+	case msg.Header.Type == fftypes.MessageTypeDefinition:
```
peterbroadhurst, line 431: This looks right to me 👍
Think this is the substantial change, matching the package rename.
We went from "If it's on the system namespace, the system handlers handle it"...
To "If it's a definition, then the definition handlers handle it"

internal/events/aggregator_test.go
peterbroadhurst: Just to check, why wouldn't you just change `msh := ag.syshandlers.(*syshandlersmocks.SystemHandlers)` to be the DefinitionHandlers, and return false and set `Header.Type` on the message in `attemptMessageDispatch`?

e.g. so you're testing the same path.

nguyer:
Re-running the build because it failed due to this bug: https://github.com/hyperledger/firefly/issues/265

nguyer:
Re-running the build because it failed due to this bug: https://github.com/hyperledger/firefly/issues/265

internal/definitions/definition_handler.go
```diff
@@ -61,43 +61,43 @@ func NewSystemHandlers(di database.Plugin, dx dataexchange.Plugin, dm data.Manag
...
+		return dh.handleOrganizationBroadcast(ctx, msg, data)
 	case fftypes.SystemTagDefineNode:
-		return sh.handleNodeBroadcast(ctx, msg, data)
+		return dh.handleNodeBroadcast(ctx, msg, data)
 	case fftypes.SystemTagDefinePool:
-		return sh.handleTokenPoolBroadcast(ctx, msg, data)
+		return dh.handleTokenPoolBroadcast(ctx, msg, data)
 	default:
-		l.Warnf("Unknown topic '%s' for system broadcast ID '%s'", msg.Header.Tag, msg.Header.ID)
+		l.Debugf("Unknown topic '%s' for system broadcast or definition ID '%s'", msg.Header.Tag, msg.Header.ID)
```
awrichar, line 95: This is really only for definitions now, right? The "or" seems superfluous. Also we say "topic" but actually log the "tag".

internal/events/aggregator.go
```diff
@@ -428,10 +428,10 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *fftypes.M
...
 		// We handle system events in-line on the aggregator, as it would be confusing for apps to be
```
awrichar, line 432: "system events" is now more accurately "definition events"

awrichar (approved):
New changes look great. Noted a few spelling items in logs/comments.

awrichar:
Looks good, thanks for making those adjustments.

PR #313, opened 2021-11-03
Additional cleanup to support tokens UI

awrichar:
Part of #218


peterbroadhurst (approved):
Looks great @awrichar - does feel more consistent with this use of the UUID

PR #321, opened 2021-11-13
Prometheus Metrics for Mux Router

### LINKED ISSUE #189: Prometheus Instrumentation
We want to start building out robust metrics in FireFly for better operations. As a start, we need to create a `metrics` endpoint that Prometheus can scrape. This should be pretty easy with the HTTP handler built into the Prometheus Go library: https://github.com/prometheus/client_golang.

As a basic demonstration of the new metrics capabilities, we could just add a simple counter for the number of API calls that have been made to the node. Any time an API endpoint is called, it would simply increment a counter.

As a demo of this functionality, it would be great to see the metrics being reported in Prometheus itself. This could simply be a local instance of Prometheus scraping logs from FireFly nodes running locally, set up by the CLI. The FireFly CLI does not need to set up Prometheus.

hfuss:
## TODOs
- [x] Move `/metrics` to separate port / mux router
- [x] Migrate `prometheus.Registry` to its own package `internal/metrics` similar to `internal/log`. Write unit tests
- [ ] E2E test for `/metrics`
- [ ] Relevant FF CLI PR
- [x] Open issue for Helm chart Prometheus support (https://github.com/hyperledger/firefly/issues/322)
- [ ] Docs / scripts / other support for docker-composing Prom and Grafana

## Description

Closes #189

Provides useful histograms for measuring API endpoint latency. for example:
```
 curl http://localhost:5000/metrics
...
# HELP apiserver_rest_request_duration_seconds Histogram of the request duration
# TYPE apiserver_rest_request_duration_seconds histogram
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.005"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.01"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.025"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.05"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.1"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.25"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="0.5"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="1"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="2.5"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="5"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="10"} 0
apiserver_rest_request_duration_seconds_bucket{code="200",host="127.0.0.1:5000",method="POST",route="/api/v1/network/nodes/self",le="+Inf"} 1
```

Really just a prototype PR to get simple metrics working and exposed. Next PR would be to actually try to instrument something like `batchPin` or a plugin and see what refactoring would need to be done around how the Prometheus registry
is handled and provided.

Should there be a test in `server_test.go` that actually asserts the `/metrics` route is present or returns a response?


internal/apiserver/server.go
```diff
@@ -61,6 +65,7 @@ type apiServer struct {
 	maxFilterSkip      uint64
 	apiTimeout         time.Duration
 	apiMaxTimeout      time.Duration
+	metricsRegistery   *prometheus.Registry
```
hfuss, line 68: Long-term this needs to be stored in a global(ish) location so that other parts of the codebase can use it to register custom metrics.

Storing it in the `apiServer` right now so that it can be reused between the REST and admin router, but it feels wrong. If theres a better place for it or if it needs to be stored as a separate module (maybe `internal/metrics`?) happy to do it just need some guidance / examples of how best to organize this.

peterbroadhurst (approved):
This looks great @hfuss 👍 

Feels like the next step would be the first internal metric. I don't know if you have any pointers to @shorsher or others who might want to add in metrics, for how to use the framework you've added.

I see there's `metrics.Registry()` to get the Prometheus registry object. Should code simply use that directly? Or should there be some internal wrapper interface to do things like increment/add statistics/counters?

hfuss:
@peterbroadhurst yep would love to do / help with the first internal metric for something like `batch` or `batchPin` in the next PR. @shorsher lemme know if you wanna sit down and work on this for an hour at some point soon.

> I see there's `metrics.Registry()` to get the Prometheus registry object. Should code simply use that directly? Or should there be some internal wrapper interface to do things like increment/add statistics/counters?

I think we'd only want to create our own wrappers for metrics primitives if we want to support exporting metrics to other formats / publishers like StatsD. Otherwise, in the next PR I'm thinking we'll add functions like `metrics.WithCounter` (similar to logging) to the `internal/metrics` package and that will ensure the Prometheus counter returned is registered properly and so primitives can be fetched via their name.

But figured when we go to add the first few internal metrics that will best inform us how to organize / refactor the metrics package.

peterbroadhurst:
Fantastic - thanks @hfuss.
Please feel free to merge this PR then, and follow-ups can happen on future PRs.

PR #334, opened 2021-11-22
add batchpin counter prometheus metric

shorsher:
Adds `BatchPinCounter` prometheus metric, which will be incremented each time a `batchpin` is submitted

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -16,19 +16,38 @@
 
 package metrics
 
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
 
 var registry *prometheus.Registry
+var BatchPinCounter prometheus.Counter
+
+// MetricsBatchPin is the prometheus metric for total number of batch pins submitted
+var MetricsBatchPin = "ff_batchpin_total"
 
 // Registry returns FireFly's customized Prometheus registry
 func Registry() *prometheus.Registry {
 	if registry == nil {
+		initMetricsCollectors()
 		registry = prometheus.NewRegistry()
-		registry.MustRegister(prometheus.NewGoCollector())
-		registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+		registerMetricsCollectors()
 	}
 
 	return registry
+}
+
+func initMetricsCollectors() {
+	BatchPinCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: MetricsBatchPin,
+		Help: "Number of batch pins submitted",
+	})
+}
+
+func registerMetricsCollectors() {
+	registry.MustRegister(prometheus.NewGoCollector())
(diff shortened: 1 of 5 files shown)


internal/metrics/metrics.go
```diff
@@ -16,16 +16,28 @@
...
 var registry *prometheus.Registry
+var BatchPinCounter prometheus.Counter
+
+// MetricsBatchPin is the prometheus metric for total number of batch pins submitted
+var MetricsBatchPin = "ff_batchPin_total"
 
 // Registry returns FireFly's customized Prometheus registry
 func Registry() *prometheus.Registry {
 	if registry == nil {
+		BatchPinCounter = prometheus.NewCounter(prometheus.CounterOpts{
```
hfuss, line 32: Since there will be a lot of metrics eventually would almost suggest creating a private method like `initMetricCollectors` or something.

internal/metrics/metrics.go
```diff
@@ -16,16 +16,28 @@
...
+			Name: MetricsBatchPin,
+			Help: "Number of batch pins submitted",
+		})
+
 		registry = prometheus.NewRegistry()
 		registry.MustRegister(prometheus.NewGoCollector())
 		registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+		registry.MustRegister(BatchPinCounter)
```
hfuss, line 40: Similarly a separate method `registerMetricCollectors`, a bit preemptive now so not requiring it just a suggestion.

internal/metrics/metrics.go
```diff
@@ -16,16 +16,28 @@
...
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
 
 var registry *prometheus.Registry
+var BatchPinCounter prometheus.Counter
+
+// MetricsBatchPin is the prometheus metric for total number of batch pins submitted
+var MetricsBatchPin = "ff_batchPin_total"
```
hfuss, line 27: Pedantic, but would either go all lowercase `batchpin` or `batch_pin` just based on what I've seen metrics normally named in other projects.

hfuss (approved):
lgtm for the first metric! Thank you for getting this going @shorsher.

Next PR I think we can then start experimenting with metric labels and using histograms to time the total duration of batchpins across the DB and blockchain interactions.

For those interested, here are some good resources for learning more about Prometheus metric types and when to use histograms specifically (TL;DR pretty much always use them instead of summaries if you want to aggregate across multiple processes / pods):
* https://prometheus.io/docs/concepts/metric_types/
* https://prometheus.io/docs/practices/histograms/

internal/metrics/metrics.go
shorsher: `batchpin` works for me, thanks 🙏🏻

internal/metrics/metrics.go
shorsher: yep, I'll add this for both like you suggested, it'll be good to set the precedent now.

nguyer (approved):

PR #335, opened 2021-11-22
Add token URI

### LINKED ISSUE #326: Allow setting a URI for NFTs
It's common for a non-fungible token to have an associated URI which points to an immutable and unique identifier for that digital asset (IPFS CIDs are an example of a commonly used identifier).

FireFly should support setting the URI for a given non-fungible token. This also implies work in the token connector to define the behavior and API at that level.

For reference, ERC1155 has the following behaviors surrounding URIs:
* A [URI event](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/IERC1155.sol#L45) that should be emitted whenever the URI for a token is set/changed
* A [uri() getter](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4c8642b70aadbac89e2b5a55f49890cd4281546e/contracts/token/ERC1155/ERC1155.sol#L59) that can be used to retrieve the URI for a given token (but by default it returns the same URI for all tokens on the contract)
* A private [_setUri() setter](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4c8642b70aadbac89e2b5a55f49890cd4281546e/contracts/token/ERC1155/ERC1155.sol#L245) that sets the URI (again, assuming the same URI for all tokens and leveraging ID substitution)

These behaviors can be changed/extended as needed to support general use.

awrichar:
Depends on https://github.com/hyperledger/firefly-tokens-erc1155/pull/44
Resolves #326

### DIFF
--- a/db/migrations/postgres/000047_add_token_uri_fields.down.sql
+++ b/db/migrations/postgres/000047_add_token_uri_fields.down.sql
@@ -1,0 +1,9 @@
+BEGIN;
+DROP INDEX tokenbalance_pool;
+DROP INDEX tokenbalance_uri;
+
+ALTER TABLE tokenbalance DROP COLUMN uri;
+ALTER TABLE tokentransfer DROP COLUMN uri;
+
+CREATE UNIQUE INDEX tokenbalance_pool ON tokenbalance(key,pool_id,token_index);
+COMMIT;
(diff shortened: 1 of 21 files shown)


peterbroadhurst (approved):
Looks good 👍 
Will leave you to do the `manifestgen.sh head` once the tokens PR is merged, and built (one small comment left on that one).

nguyer (approved):

PR #336, opened 2021-11-23
Leverage new token transfer id from plugin

awrichar:
Depends on https://github.com/hyperledger/firefly-tokens-erc1155/pull/45 (E2E tests will fail without this)


nguyer (approved):
Looks good to me

PR #337, opened 2021-11-24
Updates to handle recursive structures in swagger - depends on kin-openapi change

peterbroadhurst:
- Depends on https://github.com/getkin/kin-openapi/pull/451
- Needs the `go.mod` replace stripping out before it can move out of draft
- Corrects a bug where we weren't honouring the `ffenum` etc. tags on outputs (only inputs)

### DIFF
--- a/internal/apiserver/route_post_new_subscription.go
+++ b/internal/apiserver/route_post_new_subscription.go
@@ -31,7 +31,7 @@ import (
 )
 
 func newSubscriptionSchemaGenerator(ctx context.Context) string {
-	baseSchema, _, _ := openapi3gen.NewSchemaRefForValue(&fftypes.Subscription{})
+	baseSchema, _ := openapi3gen.NewSchemaRefForValue(&fftypes.Subscription{}, nil)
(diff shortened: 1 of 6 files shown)


awrichar:
Looks like upstream PR has been merged 🥳

internal/apiserver/route_post_new_subscription.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 func newSubscriptionSchemaGenerator(ctx context.Context) string {
-	baseSchema, _, _ := openapi3gen.NewSchemaRefForValue(&fftypes.Subscription{})
+	baseSchema, _ := openapi3gen.NewSchemaRefForValue(&fftypes.Subscription{}, nil)
```
awrichar, line 34: Is there any issue with passing nil here? @peterbroadhurst

peterbroadhurst: Think that's fine - would only be a problem if we ever introduce a recursive schema into one of the subscription types 👍

awrichar (approved):

PR #340, opened 2021-11-30
Fix Hyperledger Image Refs and Migration Job's Postgres URL in Helm Chart

hfuss:
Noticed these when trying to deploy FireFly in live environments via the Helm chart

## Fixes

* `psql_url` needs to be a key in the `firefly-config` `Secret` when using the migration job
* uses for `ghcr.io/hyperledger/` for Docker repos
* no longer uses `latest` image tag, and instead uses `v0.9.2` for DX and `.Chart.AppVersion` for FF (which is set to `v0.11.1`)
* error message for when `organization.key` is not set is misleading since it reference the deprecated `organization.identity`

### DIFF
--- a/internal/i18n/en_translations.go
+++ b/internal/i18n/en_translations.go
@@ -133,7 +133,7 @@ var (
 	MsgUnknownDataExchangePlugin   = ffm("FF10213", "Unknown Data Exchange plugin '%s'")
 	MsgParentIdentityNotFound      = ffm("FF10214", "Organization with identity '%s' not found in identity chain for %s '%s'")
 	MsgInvalidSigningIdentity      = ffm("FF10215", "Invalid signing identity")
-	MsgNodeAndOrgIDMustBeSet       = ffm("FF10216", "node.name, org.name and org.identity must be configured first", 409)
+	MsgNodeAndOrgIDMustBeSet       = ffm("FF10216", "node.name, org.name and org.key must be configured first", 409)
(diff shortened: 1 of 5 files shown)


internal/i18n/en_translations.go
```diff
@@ -133,7 +133,7 @@ var (
 	MsgUnknownDataExchangePlugin   = ffm("FF10213", "Unknown Data Exchange plugin '%s'")
 	MsgParentIdentityNotFound      = ffm("FF10214", "Organization with identity '%s' not found in identity chain for %s '%s'")
 	MsgInvalidSigningIdentity      = ffm("FF10215", "Invalid signing identity")
-	MsgNodeAndOrgIDMustBeSet       = ffm("FF10216", "node.name, org.name and org.identity must be configured first", 409)
+	MsgNodeAndOrgIDMustBeSet       = ffm("FF10216", "node.name, org.name and org.key must be configured first", 409)
```
hfuss, line 136: Unrelated to the chart changes, but this error message confused me a bit since this is whats displayed when neither `org.key` nor `org.identity` are set (which was happening w/ my chart installation due to a bug on my end). It should references the new config key instead rather than the deprecated one.

awrichar (approved):
Seems fine to me. Note that I just tagged new versions of dependent repos, so not sure if that implies a change here (ie bumping the dataexchange version).

hfuss:
@awrichar thanks for the review! Pushed an update to use the latest FF and DX releases. This should be good to merge whenever you're ready unless @peterbroadhurst needs to give it a look.

PR #341, opened 2021-12-01
Metrics endpoint to support charting

eberger727:
In order to support histograms, pie charts, and other charting in [Firefly-UI](https://github.com/hyperledger/firefly-ui), an endpoint is needed to fetch an array of row counts in a table, where each index (bucket) of the array is the count for the corresponding time interval between the start and end time.

time interval = `startTime - endTime` / `buckets`

GET `/api/v1/namespaces/default/metrics` with the following params:
`startTime`: unix time stamp to specify the start time of the data you want to fetch (ex. 1638279785)
`endTime`: unix time stamp to specify the end time of the data you want to fetch (ex. 1638366185)
`buckets`: number of buckets to divide the data into between the start and end time
`type`: the table name of the data you want to fetch data for

Typical response times for this endpoint are `50-100ms`. The `buckets` param is hard capped at 100 buckets to prevent expensive queries to the DB


internal/apiserver/route_get_metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/hyperledger/firefly/internal/config"
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/database"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var getMetrics = &oapispec.Route{
+	Name:   "getMetrics",
+	Path:   "namespaces/{ns}/metrics",
```
awrichar, line 32: I think `metrics` is actually a confusing name for this endpoint, especially considering the recently added Prometheus metrics. This is really just performing a count and splitting it into buckets. Based on a quick poll of the desks around me, possibly `histogram` is a good name for it instead?

internal/apiserver/route_get_metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	Method: http.MethodGet,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+	},
+	QueryParams: []*oapispec.QueryParam{
+		{Name: "startTime", Description: i18n.MsgMetricStartTimeParam, IsBool: false},
+		{Name: "endTime", Description: i18n.MsgMetricEndTimeParam, IsBool: false},
+		{Name: "buckets", Description: i18n.MsgMetricBucketsParam, IsBool: false},
+		{Name: "collection", Description: i18n.MsgMetricCollectionParam, IsBool: false},
```
awrichar, line 41: Might consider putting this directly in the path, ie `/metrics/{collection}`. Bonus points if we can actually generate the API in such a way that the valid routes are spelled out (but if that's tough it can be deferred).

internal/apiserver/route_get_metrics.go
awrichar: Not just yet, but let's keep that in mind for the near future.

internal/apiserver/route_get_chart_histogram.go
```diff
@@ -0,0 +1,63 @@
...
+		{Name: "buckets", Description: i18n.MsgMetricBucketsParam, IsBool: false},
+		{Name: "collection", Description: i18n.MsgMetricCollectionParam, IsBool: false},
+	},
+	FilterFactory:   nil,
+	Description:     i18n.MsgTBD,
+	JSONInputValue:  nil,
+	JSONOutputValue: func() interface{} { return []*fftypes.ChartHistogram{} },
+	JSONOutputCodes: []int{http.StatusOK},
+	JSONHandler: func(r *oapispec.APIRequest) (output interface{}, err error) {
+		startTime, err := strconv.ParseInt(r.QP["startTime"], 10, 64)
```
awrichar, line 49: Consider using `ParseString` from `timeutils`, which supports a few different input formats.

awrichar: It returns `FFTime`, which can be converted to Unix time or passed directly into a lot of SQL queries.

internal/i18n/en_translations.go
```diff
@@ -211,4 +211,12 @@ var (
...
 	MsgFieldNotSpecified           = ffm("FF10292", "Field '%s' must be specified", 400)
 	MsgTokenPoolNotConfirmed       = ffm("FF10293", "Token pool is not yet confirmed")
+	MsgMetricStartTimeParam        = ffm("FF10294", "Start time of the metrics to be fetched")
+	MsgMetricEndTimeParam          = ffm("FF10295", "End time of the metrics to be fetched")
+	MsgMetricBucketsParam          = ffm("FF10296", "Number of buckets between start time and end time")
+	MsgMetricCollectionParam       = ffm("FF10297", "Collection to fetch")
+	MsgInvalidNumberOfIntervals    = ffm("FF10298", "Number of time intervals must be between %d and %d")
+	MsgInvalidMetricParam          = ffm("FF10299", "Invalid %s. Must be a number.", 400)
+	MsgErrorFetchingRows           = ffm("FF10300", "Error fetching rows from table")
+	MsgInvalidSQLQuery             = ffm("FF10301", "Invalid SQL query")
```
awrichar, line 221: Is this error used?

eberger727: nope, good catch

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+	case database.CollectionName(database.CollectionMessages):
+		return "messages", nil
+	case database.CollectionName(database.CollectionTransactions):
+		return "transactions", nil
+	case database.CollectionName(database.CollectionOperations):
+		return "operations", nil
+	case database.CollectionName(database.CollectionEvents):
+		return "events", nil
+	default:
+		return "", i18n.NewError(ctx, i18n.MsgUnknownDatabasePlugin, collection)
```
awrichar, line 58: Not sure this is the right error message here

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+func (s *SQLCommon) metricResult(ctx context.Context, rows *sql.Rows, cols []interface{}) ([]interface{}, error) {
```
awrichar, line 62: Can rename this to match the other methods in the file since we dropped the "metric" terminology

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+	results := []interface{}{}
+
+	for i := range cols {
+		results = append(results, &cols[i])
+	}
+	err := rows.Scan(results...)
+	if err != nil {
+		return nil, i18n.NewError(ctx, i18n.MsgDBReadErr, "metrics")
```
awrichar, line 70: Drop "metrics" here as well

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+	return cols, nil
+}
+
+func (s *SQLCommon) GetChartHistogram(ctx context.Context, intervals []fftypes.ChartHistogramInterval, collection database.CollectionName) (metrics []*fftypes.ChartHistogram, err error) {
+	tableName, err := s.getTableNameFromCollection(ctx, collection)
+	if err != nil {
+		return nil, err
+	}
+
+	cols := []interface{}{}
```
awrichar, line 81: Why `[]interface{}`? Isn't this ultimately a list of strings?

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+	qb := sq.Select()
+
+	for i, caseQuery := range s.getCaseQueries(intervals) {
+		query, args, _ := caseQuery.ToSql()
+		col := "case_" + strconv.Itoa(i)
```
awrichar, line 86: Could just use `fmt.Sprintf()` here

internal/database/sqlcommon/chart_sql.go
```diff
@@ -0,0 +1,114 @@
...
+		qb = qb.Column(sq.Alias(sq.Expr("SUM("+query+")", args...), col))
+	}
+
+	rows, _, err := s.query(ctx, qb.From(tableName))
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	if !rows.Next() {
+		return nil, i18n.NewError(ctx, i18n.MsgErrorFetchingRows)
```
awrichar, line 98: This isn't really an error, is it? Maybe it should just return 404 not found?

internal/database/sqlcommon/chart_sql.go
eberger727: To my understanding, this needs to be a key/value map in order to scan the columns into a string variable

eberger727: Optimized [here](https://github.com/hyperledger/firefly/pull/341/files?authenticity_token=JJH9jo6q0vd6rm8ttHLZlxXTqSAXy6T%2FevWoBSsnNz579TDfG%2FUjBQPLqDbiWqYuzWSP077QtXJNcX5TWpctfg%3D%3D&file-filters%5B%5D=.go#diff-876319e361231dfb77a886c9b4461b5c694af76af9e6d5b82efea7357d612643R86)

internal/i18n/en_translations.go
```diff
@@ -211,4 +211,12 @@ var (
...
 	MsgFieldNotSpecified           = ffm("FF10292", "Field '%s' must be specified", 400)
 	MsgTokenPoolNotConfirmed       = ffm("FF10293", "Token pool is not yet confirmed")
+	MsgHistogramStartTimeParam     = ffm("FF10294", "Start time of the data to be fetched")
+	MsgHistogramEndTimeParam       = ffm("FF10295", "End time of the data to be fetched")
+	MsgHistogramBucketsParam       = ffm("FF10296", "Number of buckets between start time and end time")
+	MsgHistogramCollectionParam    = ffm("FF10297", "Collection to fetch")
+	MsgInvalidNumberOfIntervals    = ffm("FF10298", "Number of time intervals must be between %d and %d")
+	MsgInvalidChartNumberParam     = ffm("FF10299", "Invalid %s. Must be a number.", 400)
+	MsgHistogramInvalidTimes       = ffm("FF10300", "Start time must be before end time")
+	MsgUnsupportedCollection       = ffm("FF10301", "%s collection is not supported")
```
awrichar, line 221: I think all of these last 4 can be HTTP 400

awrichar (approved):
Looks good! Noted one small change inline on error messages.

internal/i18n/en_translations.go
eberger727: done

PR #343, opened 2021-12-01
Add sql handling for contract interface events, and adjust database naming

awrichar:
This adds the sql handler for performing database operations on the "events" defined in a contract interface (or FFI).

It also adjusts the naming of "contractevents" and "contractmethods" to "contractinterface_events" and "contractinterface_methods" respectively (in the table names, file names, and method names). The primary reason for this is the need to differentiate between the _format_ of an event that is part of a contract interface, and an _actual_ event that has been emitted and recorded. But it also helps to make clear that these items are generally children of a contractinterface.

However, if this naming is too verbose, we could also consider using "ffi" in more places rather than "contractinterface".


awrichar:
Closing - meant to open this on the feature branch and not on main.

PR #344, opened 2021-12-01
Reduce overhead of gocritic

peterbroadhurst:
For consideration - this seems to reduce the overhead significantly.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,8 +6,6 @@ linters-settings:
     enabled-checks: []
     disabled-checks:
       - regexpMust
-    enabled-tags:
-      - performance
   goheader:
     template: |-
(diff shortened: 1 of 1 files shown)


awrichar:
Guess it's just a cost-benefit discussion. This does remove [some checks](https://go-critic.com/overview#checkers-from-the-performance-group) that have been flagged before (and therefore prompted us to make micro-optimizations). The overall return on these types of optimizations may not be large though.

peterbroadhurst:
> Guess it's just a cost-benefit discussion

+1 to this this assessment. Guess I've placed my vote by raising this, and am interested in the other maintainers thoughts.

nguyer (approved):

PR #346, opened 2021-12-02
Return an empty slice (rather than nil) for zero token accounts

### LINKED ISSUE #345: /tokens/accounts returns null when no accounts are in the network
When sending a GET request to `/tokens/accounts` on a firefly network with no accounts, the response is `null` rather than an expected response of `[]`

Steps to reproduce:
1. Start-up firefly org with no accounts in DB
2. Send GET request to `/tokens/accounts`
3. Response is `null`. Expected response is `[]`

awrichar:
Fixes #345


eberger727 (approved):

peterbroadhurst (approved):
🕳️ 
🪣

PR #350, opened 2021-12-03
Replace UpsertOperation with InsertOperation

awrichar:
No code path ever passes allowExisting=true, and we already have a separate
UpdateOperation method for updates. Therefore, rename this method to
InsertOperation and remove the unneeded branches.

### DIFF
--- a/internal/database/sqlcommon/operation_sql.go
+++ b/internal/database/sqlcommon/operation_sql.go
@@ -50,76 +50,35 @@ var (
 	}
 )
 
-func (s *SQLCommon) UpsertOperation(ctx context.Context, operation *fftypes.Operation, allowExisting bool) (err error) {
-
+func (s *SQLCommon) InsertOperation(ctx context.Context, operation *fftypes.Operation) (err error) {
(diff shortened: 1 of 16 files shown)


internal/database/sqlcommon/operation_sql.go
```diff
@@ -50,76 +50,35 @@ var (
...
+				operation.Plugin,
+				operation.BackendID,
+				operation.Created,
+				operation.Updated,
+				operation.Error,
+				operation.Input,
+				operation.Output,
+			),
+		func() {
+			s.callbacks.UUIDCollectionNSEvent(database.CollectionOperations, fftypes.ChangeEventTypeCreated, operation.Namespace, operation.ID)
```
awrichar, line 78: Notably, there is no path where we emit `ChangeEventTypeUpdated` for operations. The old code gave the impression that there was, but it was in a never-used code branch.

The separate `UpdateOperation` method does not emit this event, presumably because it may not know the namespace.

shorsher:
🚀

peterbroadhurst (approved):
🚤

PR #351, opened 2021-12-03
on transfer/mint/burn fail, set transaction & operation as failed

shorsher:
Previously, if a tokens related transaction failed the transaction and operation in the DB would stay in the pending state. Now, it will correctly be marked as `failed`

### DIFF
--- a/internal/assets/token_transfer.go
+++ b/internal/assets/token_transfer.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/log"
 	"github.com/hyperledger/firefly/internal/sysmessaging"
 	"github.com/hyperledger/firefly/internal/txcommon"
 	"github.com/hyperledger/firefly/pkg/database"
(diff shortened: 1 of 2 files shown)


internal/assets/token_transfer.go
```diff
@@ -298,14 +299,35 @@ func (s *transferSender) sendInternal(ctx context.Context, method sendMethod) er
...
 		panic(fmt.Sprintf("unknown transfer type: %v", s.transfer.Type))
 	}
+
+	// if transaction fails,  mark tx and op as failed in DB
+	if err != nil {
+		s.mgr.database.RunAsGroup(ctx, func(ctx context.Context) (err error) {
+			l := log.L(ctx)
+			tx.Status = fftypes.OpStatusFailed
+			if err = s.mgr.database.UpsertTransaction(ctx, tx, false); err != nil {
+				l.Errorf("SQL upsert failed: %s sql=[ %s ]", err, tx)
```
awrichar, line 317: nit: should the end of this log be `tx=%s`?

internal/assets/token_transfer.go
```diff
@@ -298,14 +299,35 @@ func (s *transferSender) sendInternal(ctx context.Context, method sendMethod) er
...
 	default:
 		panic(fmt.Sprintf("unknown transfer type: %v", s.transfer.Type))
 	}
+
+	// if transaction fails,  mark tx and op as failed in DB
+	if err != nil {
+		s.mgr.database.RunAsGroup(ctx, func(ctx context.Context) (err error) {
+			l := log.L(ctx)
+			tx.Status = fftypes.OpStatusFailed
+			if err = s.mgr.database.UpsertTransaction(ctx, tx, false); err != nil {
```
awrichar, line 316: You could probably use `UpdateTransaction` instead... although we do almost always use Upsert elsewhere.

awrichar (approved):
Noted some minor things; will leave them up to you. Overall looks good!

internal/assets/token_transfer.go
shorsher: yeah probably, it's not an actual sql statement :)

PR #354, opened 2021-12-10
[helm] Adding Storage for DX Blobs and PVC Templating Improvements

hfuss:
We aren't ensuring there's some sort of writeable filesystem to the DX pods for blob storage, might as well offer the ability to persist it too. This refactors how we template the DX PVCs to other more customization (size, storage class, etc.) for future users, as well as offers the ability to disable the PVCs and use `emptyDir` instead.

> **NOTE**: This does include a breaking change in the chart for how DX data is handled: `/data/peers` and `/data/peer-certs` are now apart of the same PVC but given subdirs within the volume. However, we have not made an official release of the chart yet.


drewmarshburn (approved):

peterbroadhurst (approved):
🧊 
💧

PR #356, opened 2021-12-12
[helm] Fabconnect Support

hfuss:
* Templates `blockchain.fabric` config section of `firefly.core` with new Helm values
* Chart CI values broken into two tests: `it-eth` and `it-fab` which tests setting connector values for both Ethereum and Fabric


drewmarshburn (approved):

deploy/charts/firefly/templates/_helpers.tpl
```diff
@@ -124,6 +124,23 @@ blockchain:
...
+  fabric:
+    fabconnect:
+      url: {{ tpl .Values.config.fabconnectUrl . }}
+      {{- if and .Values.config.fabconnectUsername .Values.config.fabconnectPassword }}
+      auth:
+        username: {{ .Values.config.fabconnectUsername | quote }}
+        password: {{ .Values.config.fabconnectPassword | quote }}
+      {{- end }}
+      retry:
+        enable: {{ .Values.config.fabconnectRetry }}
```
peterbroadhurst, line 139: Think this is a great thing to add, but we should add it consistently for `ethconnectRetry` too

peterbroadhurst (approved):
One suggestion @hfuss to consider, with approval

deploy/charts/firefly/templates/_helpers.tpl
hfuss: done, thanks!

hfuss:
Checks are failing and will need to be rerun due to ongoing DockerHub outage.

PR #361, opened 2021-12-20
Fabric connector needs to subscribe from zero

peterbroadhurst:
See conversation in https://github.com/hyperledger/firefly-fabconnect/pull/63


nguyer (approved):
Thanks!

PR #363, opened 2021-12-20
Fix send-to-self for private messages, and add group query URLs

peterbroadhurst:
I noticed that when I was only sending to myself, I was still getting data being pushed by FireFly core to DX.

Looks like when we made the identity change to introduce DIDs, we broke the code here:
https://github.com/hyperledger/firefly/blob/24d075b4ec5b08601780ebac04fefc61aebbd427/internal/privatemessaging/privatemessaging.go#L188-L191

Also @hfuss mentioned to me that we've seen problems where the config has to have the full Key specification exactly as it gets resolved by the blockchain connector (so `0x` prefix for ethereum, and the full `CN=xyz...` syntax for fabric), or things do not work correctly.

This PR:
- Ensures we always use the blockchain connector to get the resolved signing key
- Compares `Owner` of a `Node` to the signing key of the local org, not the DID
- Adds APIs to query groups, which seemed to be missing

### DIFF
--- a/docs/swagger/swagger.yaml
+++ b/docs/swagger/swagger.yaml
@@ -2144,6 +2144,168 @@ paths:
           description: Success
         default:
           description: ""
+  /namespaces/{ns}/groups:
+    get:
+      description: 'TODO: Description'
+      operationId: getGroups
+      parameters:
+      - description: 'TODO: Description'
+        in: path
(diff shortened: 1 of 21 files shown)


nguyer (approved):
Looks good to me!

PR #365, opened 2021-12-22
Validate message topics/tag before sending

awrichar:
Currently these fields are verified only by the receiver, which can lead to
sending a bad message that is rejected on the other end.

### DIFF
--- a/internal/assets/token_pool.go
+++ b/internal/assets/token_pool.go
@@ -29,7 +29,7 @@ func (am *assetManager) CreateTokenPool(ctx context.Context, ns string, pool *ff
 	if err := am.data.VerifyNamespaceExists(ctx, ns); err != nil {
 		return nil, err
 	}
-	if err := fftypes.ValidateFFNameField(ctx, pool.Name, "name"); err != nil {
+	if err := fftypes.ValidateFFNameFieldNoUUID(ctx, pool.Name, "name"); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


awrichar:
Looks like on Fabric, we generate an invalid topic name for org broadcasts. Not sure why it was working before - perhaps we're not doing the receiver-side validation of messages when dealing with org broadcasts? Seems like that should be in effect for all messages, including system definitions.

awrichar:
After further investigation, there are actually a few layers of problems here.
1. We are [ignoring message validation errors](https://github.com/hyperledger/firefly/blob/7065837c64321b70b755556954bb574071df8899/internal/events/persist_batch.go#L200) in many cases and persisting received messages that failed to validate
2. We are generating topic names for Fabric root org broadcasts which violate the 64-char max length (and could therefore overflow db columns)
3. Validation of FF names was [extended](https://github.com/hyperledger/firefly/pull/262) to disallow UUIDs as names - but there are some cases (such as topic) where a UUID is acceptable, and other cases (such as token pools) where a UUID would cause confusion in a get-by-name-or-id scenario
4. Topics and tags are not being validated by the sender before sending a message (ie the original issue noted in this PR)

awrichar:
Pushed additional fixes for the issues noted above. See each commit message for more details.

peterbroadhurst (approved):
Noting this contains a material change, that organizations are now all published on a common `ff_organizations` topic.

PR #367, opened 2021-12-23
Config map length fix, fix bug with missing events by closing the websocket

jebonfig:


internal/blockchain/ethereum/ethereum.go
```diff
@@ -303,6 +303,7 @@ func (e *Ethereum) handleMessageBatch(ctx context.Context, messages []interface{
 
 func (e *Ethereum) eventLoop() {
 	defer close(e.closed)
+	defer e.wsconn.Close()
```
awrichar, line 306: This looks correct to me. I'm now wondering what is the purpose of `close(e.closed)` on the line above this? Perhaps that's a @peterbroadhurst question?

Also, I assume we need this same fix in the Fabric plugin?

peterbroadhurst: The `e.closed` is to account for Unit Tests waiting for the event loop to end before exiting, which is important otherwise you can get spill-over of execution after the end of tests, calling mocks after they are cancelled.

It does feel like the `e.wsconn.Close()` should come before the `close(e.closed)` to me (think the risk of spurious UT failures is low, but I couldn't rule them out)

peterbroadhurst: > Also, I assume we need this same fix in the Fabric plugin?

Agreed. Think we should get the close into both - sorry it's missing, surprised at that and how it hasn't caused problems before 🤷

awrichar (approved):

PR #371, opened 2021-12-30
Fixing Websocket Connections when Prometheus Metrics Enabled

### LINKED ISSUE #370: Websocket: response does not implement http.Hijacker - when metrics enabled
When connecting to a firefly node from a websocket client, connections are being refused with an error message of: `WebSocket upgrade failed: websocket: response does not implement http.Hijacker`

Steps to reproduce:
1. Create Firefly stack
2. Open a websocket client:
    - [Chrome Extension](https://chrome.google.com/webstore/detail/simple-websocket-client/pfdhoblngboilpfeibdedpjgfnlcodoo/related?hl=en)
    - [Node](https://www.npmjs.com/package/websocket)
    - Any other ways you use to make websocket connections
3. Try to connect to `ws://localhost:5000/ws?namespace=default&ephemeral&autoack&filter.events=message_confirmed`
4. If you're able to reproduce, you'll see this error message in the firefly logs: `WebSocket upgrade failed: websocket: response does not implement http.Hijacker`

hfuss:
* Uses my fork https://gitlab.com/hfuss/mux-prometheus (opened a PR against the original: https://gitlab.com/msvechla/mux-prometheus/-/merge_requests/11) so that the necessary `Hijack` method is implemented in the wrapper around `ResponseWriter` in order to fix #370 
* Enables Prometheus metrics when running E2E tests so that any bugs its middleware is introducing are noticed
* Adds the `ff_` prefix to the middleware metrics
* Ran `go mod tidy`


internal/metrics/response_writer_test.go
```diff
@@ -0,0 +1,26 @@
+package metrics
```
hfuss, line 1: will likely need to redo this for code coverage

nguyer (approved):
Thanks @hfuss! Looks good to me after tests and coverage passes.

hfuss:
Thanks @nguyer! @eberger727 and I will ensure E2E is fixed w/ metrics enabled and then I can cleanup the coverage / unit tests.

peterbroadhurst (commented):
Would like to understand why can't just put our own middleware in, that wraps the prometheus middleware, rather than the code copy

internal/metrics/response_writer.go
```diff
@@ -7,31 +7,13 @@ import (
 	"net/http"
 )
 
-// statusResponseWriter captures the response status
-type statusResponseWriter struct {
+// websocketResponseWriter
+type websocketResponseWriter struct {
 	http.ResponseWriter
```
hfuss: @peterbroadhurst this didn't work as expected because the inherited `ResponseWriter` for `websocketResponseWriter` ended up being a `statusResponseWriter` which does not implement `Hijacker`.

I tried climbing the "inheritance stack" with something like the following since `statusResponseWriter` is private:
```golang

type wrapperResponseWriter struct {
   http.ResponseWriter
}
// ...
func (w *websocketResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
	h, ok := w.ResponseWriter.(wrapperResponseWriter).ResponseWriter.(http.Hijacker)
        // ...
}
```

but that didn't work either so decided it was easier to fork the library with the fix and have FireFly use that instead.

PR #374, opened 2022-01-03
Add regex for company name in header

nguyer:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,17 +7,20 @@ linters-settings:
     disabled-checks:
       - regexpMust
   goheader:
+    values:
+      regexp:
+        COMPANY: .*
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -7,17 +7,20 @@ linters-settings:
...
-      
+      Copyright © {{ YEAR }} {{ COMPANY }}
+
       SPDX-License-Identifier: Apache-2.0
-      
+
       Licensed under the Apache License, Version 2.0 (the "License");
       you may not use this file except in compliance with the License.
       You may obtain a copy of the License at
-
```
peterbroadhurst, line 18: Don't think you can change the whitespace here, without flagging a diff on every file

PR #359, opened 2021-12-13
[helm] Publishing Helm Chart to GHCR via OCI

hfuss:
Closes #271.

Helm offers the ability to publish and download charts via OCI registries AKA docker images.

This PR modifies the existing Helm workflow to publishes the latest chart version to GHCR underneath the `ghcr.io/hyperledger/helm/firefly` package (vs. the FireFly docker image being published under `ghcr.io/hyperledger/firefly`.) whenever there is a merge to `main` that changes either the chart files or workflow.

If a tag / GH release is made prefixed w/ `helm-v*`, and it matches the chart's current version, then an official semver release is published to GHCR.

Lastly, this includes some minor changes to the values, has the chart use the latest FireFly release, and overhauls the chart README so that folks can get started using the chart. See https://github.com/kaleido-io/firefly/tree/helm-oci-chart-publishing/deploy/charts/firefly for a preview of the README. The format of the chart README is inspired by other community charts like https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack and https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx.

PR #375, opened 2022-01-03
Have DB creation script account for username

drewmarshburn:
This script previously made incorrect assumptions about the name of the database user and the name of the database. Providing connection strings with no db name initiates a connection using the username as the database name. Instead, the connection should always be made to the default `postgres` database.


deploy/charts/firefly/scripts/ff-db-migrations.sh
```diff
@@ -3,29 +3,34 @@
...
+echo "Provided connection string: '${PSQL_URL}'"
+
 # Extract the database name from the end of the PSQL URL, and check it's there
-DB_NAME=`echo ${PSQL_URL} | sed 's/^.*\///'`
-COLONS=`echo -n $DB_NAME | sed 's/[^:]//g'`
+DB_PARAMS=`echo ${PSQL_URL} | sed 's!^.*/!!'`
+DB_NAME=`echo ${DB_PARAMS} | sed 's!?.*!!'`
 echo "Database name: '${DB_NAME}'"
+USER_NAME=`echo ${PSQL_URL} | sed 's!^.*//!!' | sed 's!:.*$!!'`
+echo "Username: '${USER_NAME}'"
```
hfuss, line 13: Does this still work if the DB doesn't require auth? Know thats a bit unlikely / harder to test.

hfuss (approved):
lgtm otherwise! Thanks for finding this edge case and submitting a fix.

deploy/charts/firefly/scripts/ff-db-migrations.sh
drewmarshburn: Yeah that's a good point. Now, we need connection strings in at least this format:

"postgresql://[username]:[password]/[database]"

Looking at their doc (https://www.postgresql.org/docs/current/libpq-connect.html), supporting all formats of connection strings would be pretty difficult. Wondering what we need to support here? @peterbroadhurst and @hfuss  any insights from what you've previously deployed? It may also be easier to provide the PSQL connection info as separate chart values instead of winding up with all this `sed` `echo fun | sed 's/fun/suffering/'`.

hfuss:
@drewmarshburn please rebase your PR off the latest upstream/main to get the changes from #359 and to retrigger the Go build that broke yesterday.

This should be good to merge after that.

deploy/charts/firefly/scripts/ff-db-migrations.sh
hfuss: I think assuming the PSQL string will have basic auth creds is fine for now. Most users are going to have a setup like that when using the chart.

nguyer (approved):

PR #377, opened 2022-01-03
Set fetch-depth to get the full history in GitHub Actions

nguyer:
This is to fix the linter failing while checking the header copyright date for files that were not changed this year. This change pulls the full git history in the checkout stage, similar to what would happen if a user did a `git clone` on their local machine.

This PR will likely need to be merged without a passing build, due to the fact that all builds are stuck needing this change.


peterbroadhurst (approved):
👍

PR #384, opened 2022-01-06
FIR-4 (rich query updates) and FIR-7 (DX manifests) integration to onchain-logic branch

peterbroadhurst:
Delivery PR for this FIRs into the `onchain-logic` next-release branch:
- https://github.com/hyperledger/firefly-fir/pull/4
- https://github.com/hyperledger/firefly-fir/pull/7

### DIFF
--- a/db/migrations/postgres/000002_create_data_table.up.sql
+++ b/db/migrations/postgres/000002_create_data_table.up.sql
@@ -8,7 +8,7 @@ CREATE TABLE data (
   datatype_version VARCHAR(64)     NOT NULL,
   hash             CHAR(64)        NOT NULL,
   created          BIGINT          NOT NULL,
-  value            BYTEA           NOT NULL,
+  value            TEXT           NOT NULL,
   blobstore        BOOLEAN         NOT NULL
 );
 CREATE UNIQUE INDEX data_id ON data(id);
--- a/db/migrations/postgres/000004_create_batches_table.up.sql
+++ b/db/migrations/postgres/000004_create_batches_table.up.sql
@@ -8,7 +8,7 @@ CREATE TABLE batches (
   group_hash  CHAR(64),
   hash        CHAR(64),
   created     BIGINT          NOT NULL,
-  payload     BYTEA           NOT NULL,
+  payload     TEXT           NOT NULL,
   payload_ref CHAR(64),
   confirmed   BIGINT,
   tx_type     VARCHAR(64)     NOT NULL,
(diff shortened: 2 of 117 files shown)


peterbroadhurst:
Did a bit more improvement work, after exploratory UI testing during the development of https://github.com/hyperledger/firefly-ui/pull/85

- Support for matching partial hash/ID hex values
  - Meant treating the input as string, rather than requiring it to be a full parsable UUID
- Being explicit about whether you are querying for a blank or null value
  - Tweak to the FIR to add a `?` modifier to treat empty as `nil` in the code

peterbroadhurst:
Found that SQLite doesn't support `ILIKE`, and needs a PRAGMA to be case sensitive.
Also that in PSQL ILIKE is slower than `lower(field) LIKE '%search-in-lower%'`

So made some tweaks to allow ILIKE to be disabled

peterbroadhurst:
Need to sort this in E2E that I didn't see in debug:

```
panic: sql: Register called twice for driver sqlite3_ff
```

peterbroadhurst:
That should be solved.

Slightly more awkward revelation - we should be storing our JSON fields as `TEXT` not `BYTEA` I think, otherwise users can't perform a sub-string search. While not something people should rely on in production, it's a big pain if you can't as your operations console to tell you which data JSON contains the business info you need.

peterbroadhurst:
Good news it it seems our SQL layer can happily take `[]byte` values from our `jsonField` type for both `TEXT` and `BLOB` columns. So it should be the case that we can just change the migrations for new envs 🤞

peterbroadhurst:
I may have spoken too soon. Potentially we're now just storing hex in our `TEXT` field

peterbroadhurst:
ok - I've backed out the change from `BYTEA` -> `TEXT`.
Instead think I need to make it an error to query a partial match on a JSON field. This is something that the community might want to come back to (noting it would mean changing the internal type of `fftypes.Byteable` from `[]byte` to `string`, and dealing with DB migration considerations).

peterbroadhurst:
ok - big commit coming, so putting back to draft.

Also discussed with @nguyer that it might make sense for me to redirect this PR to the onchain-logic branch.

peterbroadhurst:
ok - good news https://github.com/hyperledger/firefly/pull/384#issuecomment-1007447847 proved to be true on the database side, once I'd done the big code change on the FireFly side to use Go `string` (instead of `[]byte`) as the type for the JSON types.

New envs will use `TEXT` and hence be happily able to do sub-string queries.

Existing envs will have `BYTEA` containing their JSON, and won't be able to query, but can still retrieve and store data.

peterbroadhurst:
Changing target... closing

PR #385, opened 2022-01-06
[psql-maxconns] default connection limit for postgresql

### LINKED ISSUE #380: Perf-Test: Database begin transaction failed: pq: too many clients already
During a performance test of broadcasts at 50 tx/sec, the following two database errors were encountered ~4.5 hours into the test:
- `ERROR SQL query failed: pq: sorry, too many clients already`
- `FF10114: Database begin transaction failed: pq: sorry, too many clients already`

This causes the `POST /messages/broadcast` requests to not be handled at full capacity and returning `500` status codes

bcast

[broadcastErrors.log.zip](https://github.com/hyperledger/firefly/files/7815806/broadcastErrors.log.zip)

eberger727:
Adding default max connection limit for PSQL to fix #380

### DIFF
--- a/internal/database/postgres/config.go
+++ b/internal/database/postgres/config.go
@@ -1,4 +1,4 @@
-// Copyright © 2021 Kaleido, Inc.
+// Copyright © 2022 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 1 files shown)


peterbroadhurst (approved):
🌈  5️⃣  0️⃣

PR #387, opened 2022-01-06
Prevent send/receive deadlock in wsclient

awrichar:
If a receiver sends a new websocket message while processing a received message, and
the send fails, the code can deadlock with the send and receive loops waiting on each
other.

### DIFF
--- a/pkg/wsclient/wsclient.go
+++ b/pkg/wsclient/wsclient.go
@@ -1,4 +1,4 @@
-// Copyright © 2021 Kaleido, Inc.
+// Copyright © 2022 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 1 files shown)


pkg/wsclient/wsclient.go
```diff
@@ -204,25 +204,19 @@ func (w *wsClient) readLoop() {
...
-
-		// return any error
 		if err != nil {
 			l.Errorf("WS %s closed: %s", w.url, err)
 			return
 		}
 
 		// Pass the message to the consumer
 		l.Tracef("WS %s read (mt=%d): %s", w.url, mt, message)
-		w.receive <- message
```
awrichar, line 225: Previous code could wait indefinitely here, if receiver was trying to perform a send that failed.

shorsher (approved):
💀 🔒

peterbroadhurst (approved):
Sorry I can’t recall why we had the previous order in this code. This change definitely looks correct

PR #391, opened 2022-01-07
Fabric validation of FFI Types

nguyer:


internal/blockchain/fabric/fabric.go
```diff
@@ -323,11 +327,15 @@ func (f *Fabric) handleContractEvent(ctx context.Context, msgJSON fftypes.JSONOb
 	sub := msgJSON.GetString("subId")
 	name := msgJSON.GetString("eventName")
 
+	// TODO: Get timestamp from blockchain once available
+	timestamp := fftypes.Now()
```
peterbroadhurst, line 331: Tracked in https://github.com/hyperledger/firefly-fabconnect/issues/70

internal/blockchain/fabric/fabric.go
```diff
@@ -596,8 +604,65 @@ func parseContractLocation(ctx context.Context, location fftypes.Byteable) (*Loc
...
-	return nil
+	paramDetails, err := f.parseParamDetails(ctx, param.Details)
+	if err != nil {
+		return err
+	}
+	return f.validateParamInternal(ctx, param, paramDetails)
+}
+
+func (f *Fabric) parseParamDetails(ctx context.Context, details fftypes.Byteable) (*paramDetails, error) {
+	ethParam := paramDetails{}
```
peterbroadhurst, line 615: Maybe `fabParam`?

internal/blockchain/fabric/fabric.go
```diff
@@ -596,8 +604,65 @@ func parseContractLocation(ctx context.Context, location fftypes.Byteable) (*Loc
...
+		return nil, i18n.NewError(ctx, i18n.MsgContractParamInvalid, "'type' not set")
+	}
+	return &ethParam, nil
+}
+
+func (f *Fabric) validateParamInternal(ctx context.Context, param *fftypes.FFIParam, paramDetails *paramDetails) error {
+	switch {
+	case len(param.Components) > 0:
+		// object
+		if paramDetails.Type == "object" {
```
peterbroadhurst, line 629: Should this be a constant?

e.g. I'm assuming this means `FabricParamTypeObject`

internal/blockchain/fabric/fabric.go
```diff
@@ -596,8 +604,65 @@ func parseContractLocation(ctx context.Context, location fftypes.Byteable) (*Loc
...
+			for _, childParam := range param.Components {
+				if err := f.ValidateFFIParam(ctx, childParam); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+	case strings.HasPrefix(param.Type, "byte"):
```
peterbroadhurst, line 637: I wonder if a comment/reference on this function would help describe the set of things that are supported.
Reading this line on it's own I'm wondering what the suffix is, as below we have handling of a `[]` suffix.

internal/blockchain/fabric/fabric.go
```diff
@@ -596,8 +604,65 @@ func parseContractLocation(ctx context.Context, location fftypes.Byteable) (*Loc
...
+		// byte (array)
+		if paramDetails.Type == "string" {
```
peterbroadhurst, line 639: Same comment on constant such as `FabricParamTypeString` - I also wonder how I specify bytes in Fabric?
e.g. if I had a non-UTF8 encoded binary input, how would I handle that?

internal/blockchain/fabric/fabric.go
```diff
@@ -596,8 +604,65 @@ func parseContractLocation(ctx context.Context, location fftypes.Byteable) (*Loc
...
+			return nil
+		}
+	case strings.HasSuffix(param.Type, "[]"):
+		// array
+		if strings.Count(param.Type, "[]") == strings.Count(paramDetails.Type, "[]") {
+			param.Type = strings.TrimSuffix(param.Type, "[]")
+			paramDetails.Type = strings.TrimSuffix(paramDetails.Type, "[]")
+			return f.validateParamInternal(ctx, param, paramDetails)
+		}
+	case param.Type == "integer":
```
peterbroadhurst, line 649: Note above comments suggest we need two constants here:
```go
	case param.Type == FireFlyParamTypeInteger:
```

Then below:

```go
		if paramDetails.Type == FabricParamTypeInteger {
```

peterbroadhurst (commented):
Sorry for so many comments here @nguyer - a bunch of my questions are probably as related to code that's in the branch from previous commits, as from this commit.

nguyer:
After a discussion yesterday, decided that this was not correct approach to this problem. Re-evaluating the solution in https://github.com/hyperledger/firefly/issues/405

Issue #393, opened 2022-01-10
Create operation and transaction for custom smart contract calls


nguyer:
Invoking a smart contract should be tracked by an `Operation` and a `Transaction` in FireFly.

awrichar:
This item turns out to be difficult because we have no guarantees about what events may be emitted as a result of a custom smart contract call. This means many of the existing fields of `Transaction` can't be filled in, which means re-thinking the purpose of `Transaction` and how it fits into the landscape in light of all the custom onchain work.

PR #408, opened 2022-01-12
Rectify Transactions and BlockchainEvents

awrichar:
Pre-requisite refactoring for https://github.com/hyperledger/firefly/issues/393


awrichar:
Here's a diagram to describe the new relationships between Transactions, BlockchainEvents, and other items in FireFly:
Firefly___Asset_Trail_-_architecture_whiteboard__Online_Whiteboard_for_Visual_Collaboration

db/migrations/postgres/000054_create_blockchainevents_table.up.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN;
+CREATE TABLE blockchainevents (
```
peterbroadhurst, line 2: Think we need an index on `tx_id` for this one

db/migrations/sqlite/000054_create_blockchainevents_table.up.sql
```diff
@@ -1,12 +1,16 @@
...
-  outputs          BYTEA,
+  name             VARCHAR(256)    NOT NULL,
+  protocol_id      VARCHAR(256)    NOT NULL,
+  timestamp        BIGINT          NOT NULL,
+  subscription_id  UUID,
+  output           BYTEA,
   info             BYTEA,
-  timestamp        BIGINT          NOT NULL
+  tx_type          VARCHAR(64),
+  tx_id            UUID
```
peterbroadhurst, line 13: Same index comment as for PSQL

db/migrations/postgres/000054_create_blockchainevents_table.up.sql
peterbroadhurst: and `id`

pkg/blockchain/plugin.go
```diff
@@ -129,23 +126,45 @@ type BatchPin struct {
...
-	//   - The hashes contain a sender specific nonce that is a monotomically increasing number
+	//   - The hashes contain a sender specific nonce that is a monotonically increasing number
 	//     for batches sent by that sender, within the context (maintined by the sender FireFly node)
 	Contexts []*fftypes.Bytes32
+
+	// Event contains info on the underlying blockchain event for this batch pin
+	Event Event
+}
+
+type Event struct {
```
peterbroadhurst, line 145: ~I think in our current codebase, this object might be better located at `fftypes.BlockchainEvent`~ understand the difference now

pkg/tokens/plugin.go
```diff
@@ -108,4 +109,20 @@ type TokenPool struct {
...
+}
+
+type TokenTransfer struct {
+	fftypes.TokenTransfer
+
+	// PoolProtocolID is the ID assigned to the token pool by the connector
+	PoolProtocolID string
+
+	// Event contains info on the underlying blockchain event for this transfer
+	Event blockchain.Event
```
peterbroadhurst, line 124: It feels slightly odd to me having the tokens plugin refer to the blockchain plugin (rather than exclusively `fftypes`).

However, I understand if `blockchain.Event` moved to `fftypes` it would need to be distinguished from `fftypes.BlockchainEvent`.

peterbroadhurst (commented):
This feels like a simplification throughout the code changes, and I'm very glad you were able to make the time to tackle this now. Thank you!

Couple of comments to consider.

db/migrations/postgres/000054_create_blockchainevents_table.up.sql
awrichar: I see that the `onchain-logic` branch has these indexes now (they are on `contractevents` but I just renamed that table):
```
CREATE INDEX blockchainevents_name ON blockchainevents(namespace,name);
CREATE INDEX blockchainevents_timestamp ON blockchainevents(timestamp);
CREATE INDEX blockchainevents_subscription_id ON blockchainevents(subscription_id);
```

If we add 2 more indexes, is that a bit much for a potentially high-write-volume table?

awrichar: I do think the ones you mentioned are useful... I guess just wondering if I should also prune any others? Possibly `name`?

peterbroadhurst: Think we go with this for now, and test - can't see any obvious candidates to cull.

pkg/tokens/plugin.go
peterbroadhurst: Let's go with this for now - journey makes sense

peterbroadhurst (approved):
Thanks for the extra comments 👍

db/migrations/postgres/000054_create_blockchainevents_table.up.sql
awrichar: In the end I did remove `namespace,name`. It's easy enough to add back later if we want.

PR #411, opened 2022-01-15
Build the URL without the firefly port if it was omitted from stack.json

jebonfig:
Signed-off-by: jebonfig <joe.bonfiglio@kaleido.io>


peterbroadhurst (approved):
👍

PR #419, opened 2022-01-20
JSON Schema validation for FFIs

nguyer:
This PR introduces JSON Schema into FFIs for describing `FFIParam`s. `FFIParam`s now have a `Schema` field that encompasses the previously used `Type` and `Details` fields.

The JSON Schema library used allows for extensions with custom compilers and validators to be added. So FireFly's JSON Schema for FFIs is a slightly more restrictive version of JSON Schema draft 2020-12 with the following requirements:

- The `type` field is always required
- The list of valid types are:
  - `string`
  - `integer`
  - `boolean`
  - `array`
  - `object`

The `type` field is now only used for telling FireFly the _input_ format of a field when a request is made to FireFly's API to invoke a smart contract.

For more specialized fields such as bytes, the input format would be `string` and an additional `contentEncoding` property can be added to the schema to indicate how to treat the contents of that string.

Additionally, blockchain plugins can expose a JSON Schema extension interface specific to that particular blockchain implementation. For example, the Ethereum plugin adds additional requirements (with another meta schema):

- `details` field is always required as an object
- `details` must contain a property called `type` that will always be a string
  - The compiler for the Ethereum JSON Schema extension also checks to make sure that the input type and the Ethereum type are a valid pairing
- An `indexed` boolean property is optional


go.sum
```diff
@@ -1081,13 +1083,8 @@ github.com/xdg-go/pbkdf2 v1.0.0/go.mod h1:jrpuAogTd400dnrH08LKmI/xc1MbPOebTwRqcT
 github.com/xdg-go/scram v1.0.2/go.mod h1:1WAq6h33pAW+iRreB34OORO2Nf7qel3VV3fjBj+hCSs=
 github.com/xdg-go/stringprep v1.0.2/go.mod h1:8F9zXuvzgwmyT5DUm4GUfZGDdT3W+LCvS6+da4O5kxM=
 github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f/go.mod h1:N2zxlSyiKSe5eX1tZViRH5QA0qijqEDrYZiPEAiq3wU=
```
awrichar, line 1085: Wonder why a few refs to this old library stuck around. Transitive dependency?

internal/blockchain/ethereum/ffi_param_validator.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func isEthereumBytesType(input string) bool {
+	matches := bytesRegex.FindStringSubmatch(input)
+	if len(matches) == 2 {
+		if matches[1] == "" {
+			return true
+		}
+		i, err := strconv.ParseInt(matches[1], 10, 0)
+		if err == nil && i >= 1 && i <= 32 {
```
awrichar, line 120: I learned something about Ethereum today. Didn't know they had fixed byte sizes 1-32. 👨‍🎓

internal/blockchain/fabric/fabric.go
```diff
@@ -646,3 +641,8 @@ func (f *Fabric) AddSubscription(ctx context.Context, subscription *fftypes.Cont
 func (f *Fabric) DeleteSubscription(ctx context.Context, subscription *fftypes.ContractSubscription) error {
 	return f.streams.deleteSubscription(ctx, subscription.ProtocolID)
 }
+
+func (f *Fabric) GetFFIParamValidator(ctx context.Context) (fftypes.FFIParamValidator, error) {
+	// Fabconnect does not require any additional validation beyond "JSON Schema correctness" at this time
+	return nil, nil
```
awrichar, line 647: 🚀

pkg/blockchain/plugin.go
```diff
@@ -55,13 +55,16 @@ type Plugin interface {
 
 	ValidateContractLocation(ctx context.Context, location *fftypes.JSONAny) error
 
-	ValidateFFIParam(ctx context.Context, method *fftypes.FFIParam) error
+	// ValidateFFIParam(ctx context.Context, method *fftypes.FFIParam) error
```
awrichar, line 58: Remove?

awrichar (approved):
Looks really really good. Nice work wrangling this into a (nearly) standard schema.

go.sum
nguyer: Yeah I'm not sure

PR #422, opened 2022-01-20
Fix token account listing on Postgres

### LINKED ISSUE #416: Error on tokens -> accounts tab in firefly UI
GET /tokens/accounts/0xe380f179aeb554c7456ae490b48fb1c5489459cd/pools?limit=4
response:
error fetching token pools for account 0xe380f179aeb554c7456ae490b48fb1c5489459cd

No accounts are showing up in the accounts tab because of this failed call

awrichar:
Postgres has much stricter constraints on DISTINCT/GROUP BY queries - specifically,
the fields used in ORDER BY must also be part of SELECT, and therefore may need
an aggregate function applied in order to give deterministic results.

Implement ordering by "seq" or "updated", which seem like the most useful.

Resolves #416

### DIFF
--- a/docs/swagger/swagger.yaml
+++ b/docs/swagger/swagger.yaml
@@ -8371,16 +8371,6 @@ paths:
           type: string
       - description: 'Data filter field. Prefixes supported: > >= < <= @ ^ ! !@ !^'
         in: query
-        name: balance
-        schema:
-          type: string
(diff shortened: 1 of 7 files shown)


peterbroadhurst (approved):
⬆️ 👥 🦄

Issue #400, opened 2022-01-10
Operation resubmission


peterbroadhurst:

peterbroadhurst:
Note duplicate of #221 and needs to be merged

awrichar:
I'm considering this done, and saying that the remaining unchecked items (Resubmit under transaction scoped operations, Button to resubmit on the existing operations panel) can be addressed after 1.0.

Issue #221, opened 2021-10-01
Operation Retry


peterbroadhurst:
The design of FireFly prioirtizes idempotent "at least once" semantics wherever possible, fitting with the modern approaches taken in the underlying technologies it integrates - like blockchain, REST, Kafka, IPFS etc.

We define an "operation" in the database every time an action is instigated by a node in the network.
These objects do not propagate - they are that member's local record of the part of the overall "transaction" they submitted.

Multiple operations roll up into different types of "transaction", and coordinating these actions and events into a programming model is a core function of FireFly (per the "ping pong" diagram https://github.com/hyperledger/firefly/issues/5)

Some examples:

- Broadcast requires an IPFS operation, and a Blockchain (eth/fabric) operation
- Private pinned send requires a set of DX operations to each member, and a Blockchain operation
- Token pool creation is the first higher level transaction, orchestrating:
  - A token operation to create, or find details about, a token on the chain
  - A broadcast transaction, to broadcast that data to the whole network

The reason these `operations` are stored in the database, is so that if they fail, they can be retried.
Each connector should have a level of retry built in, but it is impractical to retry forever. Sometimes failures require.

The retry API/UI mechanisms have not been implemented (you can just query today), and there are some considerations coming out that need to be factored in when it is.

This issue should be used for the architectural discussion on how operations evolve into safely retryable item, as started in some previous PRs - notably https://github.com/hyperledger/firefly/pull/211#discussion_r718960918

Also this item on sophisticated requirements for policy-driven retry in public chain scenarios:
https://github.com/hyperledger/firefly-ethconnect/issues/149#issuecomment-920866244

peterbroadhurst:
Note should be merged with #400

awrichar:
Duplicate of #400

Issue #74, opened 2021-06-14
Resubmit action for operations - such as data exchange send


peterbroadhurst:
Operations can fail, such as a private send, or a pin against a blockchain - even with resilient plugins.
At the point the operation is initiated it's locked in a bunch of locally sequenced messages into a batch, and in the case of a pinned private transfer, those messages could arrive over the blockchain and not over private messaging (or vs. versa).

This means that the `topics` of messages in that batch are stuck - they can't proceed past the blockage.
The operation that failed needs to be re-initiated once the underlying failure is resolved.

So we need to prove a REST API to attempt re-submission, and the UI will be able to add a simple 'retry' for failed operations.

All the data should be available to re-submit, and the processing surrounding the operations is all designed to be idempotent.
So double-sending, or double-pinning, does not actually break the model.
The first sequence to be confirmed on the blockchain wins.

peterbroadhurst:
Duplicate of #221 and #400

PR #435, opened 2022-01-21
Fail transaction when token transfer operation fails

### LINKED ISSUE #434: When transfer operation fails, the transaction is left pending
When a token transfer fails on chain, the operation moves to "failed", but the transaction does not. The transaction remains "pending" indefinitely.

awrichar:
Fixes #434

### DIFF
--- a/internal/events/operation_update.go
+++ b/internal/events/operation_update.go
@@ -1,4 +1,4 @@
-// Copyright © 2021 Kaleido, Inc.
+// Copyright © 2022 Kaleido, Inc.
 //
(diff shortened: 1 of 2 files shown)


peterbroadhurst (approved):
👍  - setting as backport candidate too

peterbroadhurst:
I note that we need to think overall in #74 (/ #221 / #400) about the overall of what it means for a transaction to fail.
No objection to this particular case being a step forwards.

peterbroadhurst:
Removed backport based on https://github.com/hyperledger/firefly/pull/440#pullrequestreview-860404002

Issue #421, opened 2022-01-20
Token transfer messages are delayed in batch assembly


awrichar:
When a token transfer is sent with a message attachment, the message is first created in "staged" state, and then moved to "ready" once the transfer succeeds. On the sender node, this "ready" message is not immediately picked up for processing, but instead waits for the timeout that polls for new messages.

Seems like this could be because Orchestrator only listens for when messages are _created_, but this is actually a message being _updated_ (because it's just changing state to "ready") - so it isn't picked up.
https://github.com/hyperledger/firefly/blob/f56d701fa569de0a8b1597a2bfe17017b0260ecb/internal/orchestrator/persistence_events.go#L38

peterbroadhurst:
Agree description above is spot on. Investigating a solution.

peterbroadhurst:
Thought I was really close with this one, having worked out the shoulder-tap mechanism - then realized a complex aspect to the problem - and actually I'm not sure we would ever send the message in some circumstances.

We keep track of which messages we've assembled into batches, by simply moving a sequence offset forwards.
Unlike with pins we don't update the messages as they're assembed into a batch - rather we update them as the batch seals.

In this case, something is potentially dropping in _behind_ the most recent assembled batch, but if we rewind we might re-read ones we've put into the batch. So I need to think about this a bit more.

peterbroadhurst:
At the moment this is how the batching logic works:
![batch_logic](https://user-images.githubusercontent.com/6660217/150621650-4c3ca62a-91e6-40b0-a73a-d6161e96ca39.jpg)

peterbroadhurst:
This is my current thought of the new batch logic:

![new_batch_logic](https://user-images.githubusercontent.com/6660217/150622173-f2880f99-2226-475a-bc31-aa821fdcdc05.jpg)

- We remove one of the loops
- Persistence only happens once for each batch in the final stage when the batch is sealed
- Dispatching is async to the assembler
- The assembler only allows one batch to be in assembly before it blocks waiting for the dispatch completion
- The batch manager keeps track of all offsets dispatched to each active batch processor
- When dispatch is completed for an offset, that state is passed all the way back to the processor
- The manager updates the *persisted* offset, when the lowest offset dispatched to all processors is complete
- The manager keeps an in-memory offset to use for each DB polling cycle, which might be ahead of the persisted one
- The manager can rewind at any time, because everything will either:
  - Be persisted with a batch ID so skipped
  - Be already in-flight

[...]

PR #438, opened 2022-01-23
Ready state changes require a bump to the message to re-sequence it

peterbroadhurst:
Fixes the core symptom reported in #421 

> Note that this does not implement the full change to batches described in the discussion on that issue.
> However, when that change happens, this fix will still be applicable.


awrichar (approved):

awrichar:
Could also call it simply "ReplaceMessage" - will leave it up to you if you prefer the current name though.

PR #439, opened 2022-01-23
v0.11.x backport: Ready state changes require a bump to the message to re-sequence it

peterbroadhurst:
Backport of #438


awrichar (approved):

awrichar (approved):
Noting that there was a follow-on commit to rename this method (but seems to be missing from this backport).

PR #440, opened 2022-01-23
v0.11.x backport: Fail transaction when token transfer operation fails

peterbroadhurst:
Backport #435


awrichar (approved):
This isn't totally needed (I thought it was a critical fix, but it's not) - and I know there's a wider discussion to be had about transactions + operations. Might still be correct in this case and fine to backport if we want it, but it's not blocking anything I know of at this point.

peterbroadhurst:
understood - cancelling 👍

PR #444, opened 2022-01-24
Restore Transaction reference on TokenTransfer

awrichar:
Although transfers initiated outside of FireFly will NOT have a Transaction, it's
still useful to track Transactions for those transfers initiated by FireFly. This
is needed for tracking of async transfer requests, filtering transfers, etc.

Partially reverts 0e654d2a7b9960e3d2a4946c788f80e5052e056e


peterbroadhurst (approved):
♻️

PR #448, opened 2022-01-25
Add Blockchain Transaction IDs to FireFly Transaction objects and add blockchain IDs to events

peterbroadhurst:
- Adds a string-array field of `blockchainIds` on the `transaction` object
  - Backed by a comma-separated field
  - Alphabetically sorted on each `UpsertTransaction` with lower-case only text, and no duplicates
  - Updated when blockchain operations complete
  - Updated when blockchain events arrive that are correlated with a blockchain transaction
  - In all current FireFly transaction types, should be zero or one blockchain transactions
  - Future extensibility to more
- Adds a structured `blockNumber/transactionIndex/eventIndex/subEventNumber` convention to all event IDs
   - Example `000000654321/000010/00002` would be:
      - Block number: `654321`
      - Transaction index: `10`
      - Log index (ethereum) / event number (fabric): `2`
   - The convention with padding (12/6/6/6 zeros) is intended to provide an alphabetically sortable list
   - For Fabric this feature depends on fixing https://github.com/hyperledger/firefly-fabconnect/issues/74
   - Tokens connectors are going to have tweaks to ensure the same number of zeros / separators for consistency
   - The final `subEventNumber` is reserved for protocols that batch logical events (such as transfers) in an data array emitted by a single event

Transaction example from E2E:

```js
{
  "id": "264200de-188a-4aae-92e7-1e5345604345",
  "namespace": "default",
  "type": "batch_pin",
  "created": "2022-01-25T22:01:08.9671519Z",
  "status": "Succeeded",
  "blockchainIds": [
    "0xb819561e2149c6a77a1de8a971895e486e064f45da26e1c48d2b4263eccffc43"
  ]
}
```

To which the following Blockchain event is associated:

```json
{
  "id": "8b85311c-b51f-4b69-bbbf-083ac4728db4",
  "sequence": 22,
  "source": "ethereum",
  "namespace": "default",
  "name": "BatchPin",
  "protocolId": "000000000031/000000/000000",
  "output": {
    "author": "0xce1fde1c97d3db88dbb83af16ac3b0efefeba80d",
    "batchHash": "0xb605c7d5238dd01fa594cab74db229a997e2c7c5150eeb43288ecdac30b62b97",
    "contexts": [
      "0x2a30597d4aca075c164bfb02c44de63802d66f1422295aa6b7ce8b0daa12c470"
    ],
    "namespace": "default",
    "payloadRef": "QmWC6JQnpoYAsWj6junfwaiq8DvQgoLBMQt2nkoaYmxzaZ",
    "timestamp": "1643148068",
    "uuids": "0x264200de188a4aae92e71e5345604345af60568cd62c48e69cadce0d498d0584"
  },
  "info": {
    "address": "0x3A4e59b55979d0D196aB39471DfA706b0a0CEce9",
    "blockNumber": "31",
    "logIndex": "0",
    "signature": "BatchPin(address,uint256,string,bytes32,bytes32,string,bytes32[])",
    "subId": "sb-51267be5-775c-4d22-6d72-b8a85675187b",
    "timestamp": "1643148068",
    "transactionHash": "0xb819561e2149c6a77a1de8a971895e486e064f45da26e1c48d2b4263eccffc43",
    "transactionIndex": "0x0"
  },
  "timestamp": "2022-01-25T22:01:08Z",
  "tx": {
    "type": "batch_pin",
    "id": "264200de-188a-4aae-92e7-1e5345604345"
  }
}
```

Event example from E2E:


pkg/fftypes/stringarray.go
```diff
@@ -0,0 +1,134 @@
...
+func (sa FFStringArray) Validate(ctx context.Context, fieldName string, isName bool) error {
+	var totalLength int
+	dupCheck := make(map[string]bool)
+	for i, n := range sa {
+		if dupCheck[n] {
+			return i18n.NewError(ctx, i18n.MsgDuplicateArrayEntry, fieldName, i, n)
+		}
+		dupCheck[n] = true
+		totalLength += len(n)
+		if isName {
```
awrichar, line 99: If it's _not_ a name, seems like we don't do any validation? Maybe we should at least be restricting the use of `,` since it's a delimiter (unless I'm missing it).

internal/blockchain/ethereum/ethereum.go
```diff
@@ -205,7 +207,6 @@ func (e *Ethereum) handleBatchPinEvent(ctx context.Context, msgJSON fftypes.JSON
 	}
 
 	if sBlockNumber == "" ||
-		sTransactionIndex == "" ||
```
awrichar, line 208: Should we validate the presence of `txIndex` and `logIndex`?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -284,19 +288,20 @@ func (e *Ethereum) handleContractEvent(ctx context.Context, msgJSON fftypes.JSON
 	}
 	delete(msgJSON, "data")
 
-	event := &blockchain.ContractEvent{
-		Subscription: sub,
+	event := &blockchain.EventWithContext{
+		Subscription:   sub,
+		BlockchainTXID: sTransactionHash,
```
awrichar, line 293: Is this used in the event handler for contract events?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -259,19 +260,22 @@ func (e *Ethereum) handleBatchPinEvent(ctx context.Context, msgJSON fftypes.JSON
...
+			ProtocolID: fmt.Sprintf("%.12d/%.6d/%.6d", blockNumber, txIndex, logIndex),
 			Output:     dataJSON,
 			Info:       msgJSON,
 			Timestamp:  timestamp,
 		},
 	}
 
 	// If there's an error dispatching the event, we must return the error and shutdown
-	return e.callbacks.BatchPinComplete(batch, authorAddress)
+	return e.callbacks.BatchPinComplete(batch, sTransactionHash, authorAddress)
```
awrichar, line 271: Feels slightly odd that for custom contract events, we use a `blockchain.EventWithContext` that contains the `BlockchainTXID` inside it, but for this path, we use a `blockchain.Event` with the blockchain transaction hash passed beside it as a separate parameter. Maybe `BlockchainTXID` should just move into `blockchain.Event`?

internal/blockchain/ethereum/ethereum.go
peterbroadhurst: The code as written will fall back to zero for these if not present (consistently across all paths in eth and fabric).

Would you prefer I move all to rejecting the event if they're not explicitly set to a numeric value?

internal/blockchain/ethereum/ffi_param_validator_test.go
```diff
@@ -108,7 +108,7 @@ func TestSchemaTypeInvalidFFIType(t *testing.T) {
 
 func TestSchemaTypeMissing(t *testing.T) {
 	_, err := NewTestSchema(`{}`)
-	assert.Regexp(t, "missing properties: 'type'", err)
```
awrichar, line 111: 🙏 I'm guessing this is what has been intermittently failing my local test runs...

internal/blockchain/fabric/fabric_test.go
```diff
@@ -573,13 +573,17 @@ func TestHandleMessageBatchPinOK(t *testing.T) {
     "chaincodeId": "firefly",
     "blockNumber": 91,
     "transactionId": "ce79343000e851a0c742f63a733ce19a5f8b9ce1c719b6cecd14f01bcf81fff2",
+		"transactionIndex": 2,
+		"eventIndex": 50,
```
awrichar, line 577: nit: indentation is inconsistent in this file?

internal/blockchain/ethereum/ethereum.go
awrichar: Hmm no, I think zero is OK. Thanks for the clarification.

internal/blockchain/ethereum/ethereum.go
peterbroadhurst: I could certainly do that

awrichar: I think that makes the most sense... Seems like we pull it out of every blockchain event.

internal/blockchain/ethereum/ethereum.go
awrichar: Ok, good for now. Will need to have a think about whether we should be putting it somewhere on that path.

internal/database/sqlcommon/transaction_sql.go
```diff
@@ -49,23 +51,27 @@ func (s *SQLCommon) UpsertTransaction(ctx context.Context, transaction *fftypes.
...
 	)
 	if err != nil {
 		return err
 	}
 	existing := transactionRows.Next()
-	transactionRows.Close()
 
 	if existing {
+		var existingBlockchainIDs fftypes.FFStringArray
+		_ = transactionRows.Scan(&existingBlockchainIDs)
```
awrichar, line 65: Should we check for error?

internal/events/token_pool_created.go
```diff
@@ -41,21 +41,22 @@ func addPoolDetailsFromPlugin(ffPool *fftypes.TokenPool, pluginPool *tokens.Toke
...
+		ID:            pool.TX.ID,
+		Status:        status,
+		Namespace:     pool.Namespace,
+		Type:          pool.TX.Type,
+		BlockchainIDs: fftypes.NewFFStringArray(blockchainTXID),
 	}
 }
 
-func (em *eventManager) confirmPool(ctx context.Context, pool *fftypes.TokenPool, ev *blockchain.Event) error {
+func (em *eventManager) confirmPool(ctx context.Context, pool *fftypes.TokenPool, ev *blockchain.Event, blockchainTXID string) error {
```
awrichar, line 54: I think the blockchain ID is now inside the `blockchain.Event`, right? so no need to pass it separately also

pkg/tokens/plugin.go
```diff
@@ -101,6 +101,9 @@ type TokenPool struct {
 	// Not guaranteed to be set for pool creation events triggered outside of FireFly
 	TransactionID *fftypes.UUID
 
+	// BlockchainTXID is the blockchain transaction hash that resulted in the event that created the pool
+	BlockchainTXID string
```
awrichar, line 105: This should not be needed (here or on TokenTransfer below) since it's been included in `blockchain.Event`

awrichar (approved):
Left a couple more comments, but marking approved now so you can merge after looking through those.

Thanks for taking this on; I think it alleviates a big chunk of the concerns/confusion over Transactions!

internal/database/sqlcommon/transaction_sql.go
peterbroadhurst: ... I took the approach that a problem here would be detected below, but it might have been lazy

internal/events/token_pool_created.go
peterbroadhurst: Sorry I missed this one

internal/database/sqlcommon/transaction_sql.go
awrichar: Yea, actually seems ok

Issue #445, opened 2022-01-25
Websocket test: close of closed channel


awrichar:
Websocket unit tests are intermittently failing with "close of closed channel".

https://github.com/hyperledger/firefly/runs/4940262645?check_suite_focus=true

```
panic: close of closed channel

goroutine 74 [running]:
github.com/hyperledger/firefly/pkg/wsclient.NewTestWSServer.func1.1(0xc0003fc7e0, 0xc000363600, 0xc000010128)
	/home/runner/work/firefly/firefly/pkg/wsclient/wstestserver.go:45 +0x10e
created by github.com/hyperledger/firefly/pkg/wsclient.NewTestWSServer.func1
	/home/runner/work/firefly/firefly/pkg/wsclient/wstestserver.go:40 +0x16c
FAIL	github.com/hyperledger/firefly/pkg/wsclient	0.071s
```

This is potentially a side-effect (or an alteration of the underlying original bug) introduced by #420.

awrichar:
https://github.com/hyperledger/firefly/runs/4972849059?check_suite_focus=true

awrichar:
Is this yet another flavor of a similar issue?
https://github.com/hyperledger/firefly/runs/5058705964?check_suite_focus=true

peterbroadhurst:
Think this is something sufficiently different, to be tracked in a new issue.
Think it will naturally become redundant when batch logic is revamped (next week) - but still good to have.

peterbroadhurst:
Moved to https://github.com/hyperledger/firefly/issues/445

Closing this one out.

PR #458, opened 2022-01-28
Avoid potential double-close of channels due to reconnects after tests complete

peterbroadhurst:
Fix for #445 

I wasn't able to recreate, but the test runs showed fast timeout heartbeating and connections... so I think in a slow environment a test is reconnecting the websocket after the test had finished.

Our little test websocket server would re-use the channels in that case, even though they were close


awrichar (approved):
Worth a try 👍

PR #459, opened 2022-01-28
Fixing E2E Tests for Multi-Member Networks

hfuss:
Noticed if I use the E2E tests to validate a multi-member (2+ members) network that it'd get confused about which members it was testing because the order of the members in the stack file can never be guaranteed to match the order of the orgs returned by the API of one of the nodes.


peterbroadhurst (approved):
🍅 🍎

PR #460, opened 2022-01-28
Allow synchronous 200 response for token pool creation

awrichar:


db/migrations/postgres/000058_remove_tokenpool_key.up.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE tokenpool DROP COLUMN key;
```
peterbroadhurst, line 2: Glad to see this drop 👍

peterbroadhurst (approved):

PR #462, opened 2022-02-01
Add DefinitionBatchActions to execute after a batch of system definitions

awrichar:


peterbroadhurst (approved):
This is great - thanks for finding what looks like a pretty elegant solution to cracking this tough nut.

My only comments, are on your comments - and about being able to understand how to use the feature in the future within the code.

Issue #469, opened 2022-02-03
Docker build fails on Apple Silicon


nguyer:
Running a Docker build on an M1 based Mac with the following command:
```
docker build -t hyperledger/firefly .
```
Results in the following:

```
------
 > [fabric-builder 10/10] RUN ./bin/peer lifecycle chaincode package /firefly/smart_contracts/fabric/firefly-go/firefly_fabric.tar.gz --path /firefly/smart_contracts/fabric/firefly-go --lang golang --label firefly_1.0:
#36 0.251 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
------
```

The dockerfile pulls x86_64 specific binaries for Fabric to build the chain code right now. My guess is that because there is an arm64 version of Go's Alpine image (which is what we're using for building), Docker pulls that, then tries to run x86_64 binaries on it, which obviously won't work.

Fabric doesn't currently provide arm64 binaries, so I'll see if I can force this build step to run on a x86_64 container

PR #473, opened 2022-02-03
Fix Docker build on ARM

nguyer:
Resolves https://github.com/hyperledger/firefly/issues/469

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,8 @@ RUN mkdir /firefly/frontend \
 ADD . .
 RUN make build
 
-FROM golang:1.16-alpine3.13 AS fabric-builder
-RUN apk add  libc6-compat
+FROM golang@sha256:bab81aadc644eee23bafdf479e9bd60c418d405dbc79370ababbb2a2f88d8034 AS fabric-builder
+RUN apk add libc6-compat
 WORKDIR /firefly/smart_contracts/fabric/firefly-go
 ADD smart_contracts/fabric/firefly-go .
 RUN GO111MODULE=on go mod vendor
@@ -21,6 +21,7 @@ RUN touch core.yaml
 RUN ./bin/peer lifecycle chaincode package /firefly/smart_contracts/fabric/firefly-go/firefly_fabric.tar.gz --path /firefly/smart_contracts/fabric/firefly-go --lang golang --label firefly_1.0
 
 FROM node:14-alpine3.11 AS solidity-builder
+RUN apk add python alpine-sdk
 WORKDIR /firefly/solidity_firefly
 ADD smart_contracts/ethereum/solidity_firefly/package*.json .
 RUN npm install
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,23 @@
-FROM golang:1.16-alpine3.13 AS firefly-builder
+ARG FIREFLY_BUILDER_TAG
+ARG FABRIC_BUILDER_TAG
+ARG FABRIC_BUILDER_PLATFORM
+ARG SOLIDITY_BUILDER_TAG
+ARG BASE_TAG
+
+FROM $FIREFLY_BUILDER_TAG AS firefly-builder
+ARG UI_TAG
+ARG UI_RELEASE
 RUN apk add make gcc build-base curl git
 WORKDIR /firefly
 ADD go.mod go.sum ./
 RUN go mod download
-ENV UI_RELEASE "https://github.com/hyperledger/firefly-ui/releases/download/v0.5.0/v0.5.0_8cb358c.tgz"
+ENV UI_RELEASE https://github.com/hyperledger/firefly-ui/releases/download/$UI_TAG/$UI_RELEASE.tgz
 RUN mkdir /firefly/frontend \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -9,8 +9,8 @@ RUN mkdir /firefly/frontend \
 ADD . .
 RUN make build
 
-FROM golang:1.16-alpine3.13 AS fabric-builder
-RUN apk add  libc6-compat
+FROM golang@sha256:bab81aadc644eee23bafdf479e9bd60c418d405dbc79370ababbb2a2f88d8034 AS fabric-builder
```
peterbroadhurst, line 12: I'm really worried about how deep we're burying dependency information that needs to be regularly updated in the Dockerfiles at the moment.

Specifically:
- UI release version
- This pretty obscure tag that links to a particular (I assume ARM friendly) image of Golang

Can these be rolled up to the Manifest?

nguyer: It's a very good point. I think it's possible to pull these into the manifest if we are okay with moving beyond just a basic `docker build ...` command. Perhaps we can add a target to the Makefile for `docker` so the command just becomes `make docker`. And that script takes care of passing the versions from the manifest to here or wherever else they need to go.

nguyer: As an aside, this is just the sha for the x86_64 version of this image (which runs just fine on an M1 Mac). It is the same image that this Dockerfile would have picked if it were running on an x86_64 host.

peterbroadhurst: Think `make docker` sounds like a good approach.

Other option which would move forwards with this PR as is, is to put in the comments a repeatable instruction to find (without owning an M1 Mac) the reverse lookup to "what Go image is this?" and most importantly the next x86_64 image to apply a Go patch release (common security update requirement).

peterbroadhurst:
```suggestion
FROM --platform=linux/x86_64 golang:1.16-alpine3.13 AS fabric-builder
```

This seemed to work for this step for me @nguyer - I think it's doing the same thing as your change

peterbroadhurst:
@nguyer - I've submitted https://github.com/hyperledger/firefly/pull/484 as a PR to this branch, with the above discussion/suggestion, and a few other changes and updates to dependencies necessary for the build to work for me

peterbroadhurst:
I note after my PR, I'm still seeing a failure from IPFS:

```
pbroadhurst@macbook-pro firefly_e2e % docker compose logs ipfs_0
firefly_e2e_ipfs_0  | Changing user to ipfs
firefly_e2e_ipfs_0  | ipfs version 0.11.0
firefly_e2e_ipfs_0  | generating ED25519 keypair...done
firefly_e2e_ipfs_0  | peer identity: 12D3KooWPgtHDht3oR2gmRRYzGzhH5UvS62NUN4fdtkdTBrLdBXK
firefly_e2e_ipfs_0  | initializing IPFS node at /data/ipfs
firefly_e2e_ipfs_0  | to get started, enter:
firefly_e2e_ipfs_0  | 
firefly_e2e_ipfs_0  | 	ipfs cat /ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc/readme
firefly_e2e_ipfs_0  | 
firefly_e2e_ipfs_0  | Copying swarm key from variable...
firefly_e2e_ipfs_0  | Initializing daemon...
firefly_e2e_ipfs_0  | go-ipfs version: 0.11.0-67220ed
firefly_e2e_ipfs_0  | Repo version: 11
firefly_e2e_ipfs_0  | System version: amd64/linux
firefly_e2e_ipfs_0  | Golang version: go1.16.7
firefly_e2e_ipfs_0  | 2022/02/07 01:37:09 [zeroconf] no suitable IPv4 interface: udp4: failed to join any of these interfaces: [{169 1500 eth0 02:42:ac:12:00:02 up|broadcast|multicast}]
firefly_e2e_ipfs_0  | 2022/02/07 01:37:09 [zeroconf] no suitable IPv6 interface: udp6: failed to join any of these interfaces: [{169 1500 eth0 02:42:ac:12:00:02 up|broadcast|multicast}]
firefly_e2e_ipfs_0  | 2022-02-07T01:37:09.907Z	ERROR	cmd/ipfs	ipfs/daemon.go:430	error from node construction: no supported interface
firefly_e2e_ipfs_0  | 
firefly_e2e_ipfs_0  | Error: no supported interface
```

peterbroadhurst:
Workaround listed here: https://github.com/ipfs/go-ipfs/issues/8645#issuecomment-1024944637

nguyer:
Yep I've run into the same thing and found that same thread. This is a firefly-cli issue. Shouldn't be too bad to implement but it's more than one line, since we're currently not writing a custom config file for IPFS. I'm going to look into this more tomorrow though.

nguyer:
I pulled in changes from https://github.com/hyperledger/firefly/pull/484 as well

peterbroadhurst:
Think it might be worth the script checking if `jq` exists:

```
 if [[ ! -x `which jq` ]]; then echo 'Please install "jq" to continue'; exit 1; fi
```

peterbroadhurst (approved):
Looks great!

PR #474, opened 2022-02-03
Make the factories act like factories

### LINKED ISSUE #467: When multiple token connectors are configured, FireFly only sends requests to one
If you configure multiple token connectors and pass a `connector` param to an API such as `/tokens/pools`, it appears to be ignored. FireFly always routes the request to the last connector in the config.

awrichar:
Instead of instantiating all plugins up front, instantiate them on demand.

Fixes #467


awrichar:
Looks like the events factory actually does want singletons and not a true factory...

peterbroadhurst (approved):
This makes sense. I feel like we could at some point spend some more cycles to make it feel cleaner, but I don't think it justifies further distraction now.

PR #475, opened 2022-02-03
Use PreFinalize/Finalize in all definition handlers

awrichar:
Follow-up to #462


peterbroadhurst (approved):
This is great. The architecture seemed to work well for all the cases, and be clear to read.

PR #477, opened 2022-02-03
[new-prom-metrics] prometheus metrics

eberger727:
to track broadcasts, private messages, and token mint/burn/transfers

Signed-off-by: David Echelberger <david.echelberger@kaleido.io>


awrichar:
Will the "start time" values be accurate if there are lots of things in flight (both from me and from others)? Seems like there's not enough correlation to determine the actual time taken by one particular message, transfer, etc.

eberger727:
Closing in favor of #503

PR #478, opened 2022-02-03
Store all token transfer inputs on operation

awrichar:
Since the transfer is only recorded upon success, all inputs must be stored
on the operation in order to support retries or accurately report failures.

Needed for #400 and #316


peterbroadhurst (approved):
📥 📤

PR #480, opened 2022-02-04
Address some inconsistencies in Operation behavior

awrichar:
- Operations should all correspond to an external plugin action (only announce_pool needed a change)
- Name all operations as "plugin_action" (only contract_invoke needed a change)
- Mark operations successful after the new [synchronous forms](https://github.com/hyperledger/firefly/pull/460) of token pool creation and activation

### DIFF
--- a/docs/swagger/swagger.yaml
+++ b/docs/swagger/swagger.yaml
@@ -4661,13 +4661,13 @@ paths:
                   type:
                     enum:
                     - blockchain_batch_pin
+                    - blockchain_invoke
                     - publicstorage_batch_broadcast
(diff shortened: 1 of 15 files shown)


peterbroadhurst (approved):
... reminds me we really need to get the reference section filled in with a bunch of the work you're doing, before V1.0

awrichar:
Yea - this was a direct result of attempting to make a concise picture of what operations mean, and seeing that a few things didn't line up. It feels very clean now.

FireFly_contributor_whiteboarding_-_public_access__Online_Whiteboard_for_Visual_Collaboration

PR #482, opened 2022-02-05
Remove extra "rejected" events

awrichar:
Only "message_rejected" will be emitted for rejected definition messages -
no special per-type event will be emitted.

This requires one extra lookup in syncasync to find rejected token pools,
but seems better for overall consistency.


awrichar:
Note that this change was prompted after drawing this picture of events:

FireFly_contributor_whiteboarding_-_public_access__Online_Whiteboard_for_Visual_Collaboration

Issue #481, opened 2022-02-04
Perf-Test: Websocket events not being sent


eberger727:
The FireFly performance tests use go routines to send multiple requests to an API at the same time. These routines then wait for their specific websocket event to confirm their transaction was completed before moving on to its next task. 

In v0.11.5 of FireFly, these websocket events were being sent correctly, allowing for workers to proceed as expected.

In v0.12.0 of FireFly, one websocket event is received, then no more websocket events come through, resulting in workers not being able to confirm that their transaction completed and moving on to their next task.

[ff-perf v0-11-5.log](https://github.com/hyperledger/firefly/files/8006353/v0-11-5.log)
[ff-perf v0-12-0.log](https://github.com/hyperledger/firefly/files/8006354/v0-12-0.log)
[v0-12-0-ff_core.log](https://github.com/hyperledger/firefly/files/8006409/v0-12-0-ff_core.log)

peterbroadhurst:
This is an interesting log entry:

```
[2022-02-04T21:43:17.643Z] DEBUG Message 627e73f3-cee3-4ff1-93cf-049fb03b9989 pinned at sequence 15 blocked by earlier context 37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f at sequence 14 dbtx=rI1QGY1p pid=1 role=aggregator
```

eberger727:
Fresh logs with timestamps
[v0-12-0-ff_core.log](https://github.com/hyperledger/firefly/files/8006470/v0-12-0-ff_core.log)
[v0-12-0-ff_eth.log](https://github.com/hyperledger/firefly/files/8006471/v0-12-0-ff_eth.log)
[v0-12-0.log](https://github.com/hyperledger/firefly/files/8006472/v0-12-0.log)

peterbroadhurst:
fyi @awrichar - investigating, but looks like we might not be resolving the pins any more in the aggregator.
Possibly this fell out with the aggregator `PreFinalize`/`Finalize` logic.
Digging more.

peterbroadhurst:
The only `UpdateNextPin` I can see in the code is in a masked (private transfer) branch

peterbroadhurst:
https://github.com/hyperledger/firefly/blob/f262c5c30c06dbc8d9c282a5620906763155ce00/internal/events/aggregator.go#L356-L374

peterbroadhurst:
[...]

PR #483, opened 2022-02-06
Update aggregator batch processing to maintain in-memory pin state until OnFinalize

peterbroadhurst:
Fix for #481 

We were failing to process multiple private messages on the same topic, within a single batch of pins that are read by the aggregator. This was a side-effect of the #462 changes - because the logic that evaluates the `NextPin` state on each message no longer had up-to-date information to read from the database.

The core fix implemented for that, was to move to in-memory processing for all state that can change during processing of a page of pins in the aggregator. So the first time we access the state on a context (context is a particular `topic`, scoped to a group if it's a private context), we read the data we need in memory and from that point on we update it in memory until the `Finalize` phase is called at the end. At that point everything is flushed.

Fixing this turned into quite a big change to the internals of the aggregator, and in doing so I did find two other less serious issues:
1. When there were multiple pins in a message (multiple topics) we would leave some of those pins with `dispatched=false` when the message was confirmed. This means potential re-processing on rewind.
2. If the pins within a message with multiple topics, spanned a page of reads from the aggregator, then when we came to the next page we could immediately re-process the message.

The fix for ^^^ was that we always mark _all_ pins associated with a message as dispatched. That can include pins outside of the page that was just read, so we calculate the start+end index of the pins within the batch for that message, and do an update in the DB scoped to that start+end range (and the batch ID).

> While enhancing the E2E I found another bug - in the automatic reply generation for webhooks, we're not setting the `topics` in the reply to match the request.

### DIFF
--- a/internal/database/sqlcommon/pin_sql.go
+++ b/internal/database/sqlcommon/pin_sql.go
@@ -149,12 +149,12 @@ func (s *SQLCommon) UpdatePins(ctx context.Context, filter database.Filter, upda
 	}
 	defer s.rollbackTx(ctx, tx, autoCommit)
 
-	query, err := s.buildUpdate(sq.Update("pins"), update, msgFilterFieldMap)
+	query, err := s.buildUpdate(sq.Update("pins"), update, pinFilterFieldMap)
 	if err != nil {
 		return err
 	}
 
-	query, err = s.filterUpdate(ctx, "", query, filter, opFilterFieldMap)
+	query, err = s.filterUpdate(ctx, "", query, filter, pinFilterFieldMap)
 	if err != nil {
 		return err
 	}
--- a/internal/database/sqlcommon/pin_sql.go
+++ b/internal/database/sqlcommon/pin_sql.go
@@ -1,4 +1,4 @@
-// Copyright © 2021 Kaleido, Inc.
+// Copyright © 2022 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -141,7 +141,7 @@ func (s *SQLCommon) GetPins(ctx context.Context, filter database.Filter) (messag
 
 }
 
-func (s *SQLCommon) SetPinDispatched(ctx context.Context, sequence int64) (err error) {
+func (s *SQLCommon) UpdatePins(ctx context.Context, filter database.Filter, update database.Update) (err error) {
 
 	ctx, tx, autoCommit, err := s.beginOrUseTx(ctx)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


internal/database/sqlcommon/pin_sql.go
```diff
@@ -141,24 +141,25 @@ func (s *SQLCommon) GetPins(ctx context.Context, filter database.Filter) (messag
...
+	if err != nil {
+		return err
+	}
+
+	query, err = s.filterUpdate(ctx, "", query, filter, pinFilterFieldMap)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.updateTx(ctx, tx, query, nil /* no change events filter based update */)
```
awrichar, line 162: So this type does reliably emit ChangeEventTypeCreated. It theoretically emits ChangeEventTypeDeleted, although I don't know of anywhere we call DeletePin. It now never emits ChangeEventTypeUpdated though.

I wonder if having change events at all is useful, or if it's confusing to emit them inconsistently.

internal/events/aggregator_batch_state.go
```diff
@@ -0,0 +1,387 @@
...
+// At the end we flush the changes to the database
+type nextPinGroupState struct {
+	groupID           *fftypes.Bytes32
+	topic             string
+	nextPins          []*fftypes.NextPin
+	new               bool
+	identitiesChanged map[string]bool
+}
+
+type NextPinState struct {
```
awrichar, line 56: Any reason this one is public/exported? Seems to be internal to the package.

internal/events/aggregator_batch_state.go
```diff
@@ -0,0 +1,387 @@
...
+// so that all pins associated to the message can be marked dispatched at the end of the batch.
+type dispatchedMessage struct {
+	batchID       *fftypes.UUID
+	msgID         *fftypes.UUID
+	firstPinIndex int64
+	lastPinIndex  int64
+}
+
+// batchState are synchronous actions to be performed while processing system messages, but which must happen after reading the whole batch
+type batchState struct {
```
awrichar, line 77: The description here doesn't fully cover the expanded purpose of this class

awrichar (approved):
I like where this has landed. The pin logic is still pretty complex, but as far as I can tell it looks right - and I think it's slightly more efficient now? Thanks for adding E2E coverage as well.

No major notes, so marking approved and we can always circle back on anything that doesn't need to be addressed immediately.

internal/database/sqlcommon/pin_sql.go
peterbroadhurst: This is common across the codebase for these update-multiple events.

We don't have a need for them where we use them currently, and it's hard in SQL without detailed DB-specific coding to find out what rows have been updated. There's also a performance overhead

internal/events/aggregator_batch_state.go
peterbroadhurst: I was thinking of the boundary of the BatchState and its interface with Aggregator.

Agree now looking back it's confusing, as these aren't on the external interface. So while this doesn't _actually_ make them public outside the package, it is confusing.

internal/events/aggregator_batch_state.go
peterbroadhurst: 👍  - I've added a bunch of extra text here - might be good for you to validate my understanding @awrichar

test/e2e/onchain_offchain_test.go
```diff
@@ -49,25 +49,41 @@ func (suite *OnChainOffChainTestSuite) TestE2EBroadcast() {
...
+		require.NoError(suite.T(), err)
+		assert.Equal(suite.T(), 202, resp.StatusCode())
+	}
 
-	waitForMessageConfirmed(suite.T(), received2, fftypes.MessageTypeBroadcast)
-	<-changes2 // also expect database change events
-	val2 := validateReceivedMessages(suite.testState, suite.testState.client2, fftypes.MessageTypeBroadcast, fftypes.TransactionTypeBatchPin, 1, 0)
-	assert.Equal(suite.T(), data.Value, val2.Value)
+	for i := 0; i < totalMessages; i++ {
+		// Wait for all thel message-confirmed events, from both participants
```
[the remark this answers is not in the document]
peterbroadhurst: Not an intentional change - I'll fix that

internal/events/aggregator_batch_state.go
awrichar: Looks very good, thanks 🤩

awrichar (approved):

PR #491, opened 2022-02-08
Fix fabric event timestamp parsing

jimthematrix:
Signed-off-by: Jim Zhang <jim.zhang@kaleido.io>


peterbroadhurst (approved):
Reviewed the FabConnect code, and this change is because the JSON payload contains a JSON `number` (Go `float64`) rather than a `string`. The `GetInt64` handles various types (number/string, in decimal or hex).

PR #494, opened 2022-02-08
Compile directly with solc to avoid full Node.js npm install of Truffle

peterbroadhurst:
Saves a lot of time and I/O activity during the docker build.

The `jq` command converts the `combined.json` output from `solc` in this syntax of JSON required by the CLI (which is a subset of the Truffle output):
https://github.com/hyperledger/firefly-cli/blob/680665495e45074b250e5ed5a74e8f152774254c/pkg/types/ethconnect.go#L19-L23


nguyer (approved):

nguyer:
The issue in this build was not caused by these changes, and should be fixed in https://github.com/hyperledger/firefly/pull/486 so I'll allow this to be merged as-is.

Issue #493, opened 2022-02-08
FF-Perf: Websocket events appear to be dropped


eberger727:
Test: 25 broadcasts, 25 private messages, one subscription

In this test, 25 workers were kicked off for both broadcasts and private messages. Shortly after the test is kicked off, workers are not receiving websocket events, preventing them from proceeding. The `ff-perf.log` has logs of message IDs and event IDs that can be used to correlate to the firefly logs that are also attached. Every minute, the perf test checks for messages that have not been confirmed in > 1 minute, do a `Ctrl+F` for `Delinquent Messages` and you will see the messageID that has not been confirmed yet, along with the time it was sent at.

<img width="1655" alt="image" src="https://user-images.githubusercontent.com/17093033/153029391-86ea67da-340c-4122-a559-b6248fb041b8.png">

[log_ethconnect_0.log.zip](https://github.com/hyperledger/firefly/files/8025588/log_ethconnect_0.log.zip)
[log_firefly_core_0.log.zip](https://github.com/hyperledger/firefly/files/8025589/log_firefly_core_0.log.zip)
[ff-perf.log](https://github.com/hyperledger/firefly/files/8025590/ff-perf.log)

eberger727:
Logs for commit `d90d176` to main
[log_firefly_core_0.log.zip](https://github.com/hyperledger/firefly/files/8025407/log_firefly_core_0.log.zip)
[log_ethconnect_0.log.zip](https://github.com/hyperledger/firefly/files/8025408/log_ethconnect_0.log.zip)
[ff-perf.log](https://github.com/hyperledger/firefly/files/8025409/ff-perf.log)

eberger727:
Run after commit `82f55af`. Looks like events are not being sent back at all, or deadlock is occuring.
<img width="1663" alt="image" src="https://user-images.githubusercontent.com/17093033/153051090-a48ea93e-4ab2-497e-8df7-b27615a03688.png">
[ff-perf.log](https://github.com/hyperledger/firefly/files/8026453/ff-perf.log)
[log_ethconnect_0.log](https://github.com/hyperledger/firefly/files/8026454/log_ethconnect_0.log)
[log_firefly_core_0.log](https://github.com/hyperledger/firefly/files/8026455/log_firefly_core_0.log)

peterbroadhurst:
Quick script to extract how many times we see each msgid in the list:
```sh
[...]

PR #496, opened 2022-02-08
Rewind when messages appear behind offset

peterbroadhurst:
Fix for #493 

Observed reviewing the logs for #493 that the messages were never being picked up by the batch processor, and the perf tool is submitting them in a highly parallel way. My analysis is that with PSQL as used in this test (or any production ready database) parallel writing of messages to the DB could result in the commit order of the transactions, being mismatched with the sequence allocation to the rows. 

The code in the batch manager did not handle this - it would simply do a shoulder tap to re-read from the DB when a message was created, but not check if that message had appeared behind the offset and it needed to do a rewind.

This code change implements a simple rewind (like we have for the event aggregator).
Note this means that in the case of parallel REST API submission of messages, the final delivery order might not match the local sequence order. However, that's in-line with the assurances FireFly provides - as application should only be able to depend on.

- The local order they submitted in (in this case there is parallel submission, so the order is not deterministic) - but they should only rely on this for their own participation, and cannot consider this the global order across participants.
- The final order as ordered by the blockchain, when pinning is used.

### DIFF
--- a/internal/batch/batch_manager.go
+++ b/internal/batch/batch_manager.go
@@ -225,7 +225,11 @@ func (bm *batchManager) assembleMessageData(msg *fftypes.Message) (data []*fftyp
 
 func (bm *batchManager) markRewind(rewindTo int64) {
 	bm.rewindMux.Lock()
-	bm.rewindTo = rewindTo
+	// Make sure we only rewind backwards - as we might get multiple shoulder taps
+	// for different message sequences during a single poll cycle.
+	if bm.rewindTo < 0 || rewindTo < bm.rewindTo {
+		bm.rewindTo = rewindTo
+	}
 	bm.rewindMux.Unlock()
 }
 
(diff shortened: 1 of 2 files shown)


internal/batch/batch_manager.go
```diff
@@ -219,7 +223,29 @@ func (bm *batchManager) assembleMessageData(msg *fftypes.Message) (data []*fftyp
 	return data, nil
 }
 
+func (bm *batchManager) markRewind(rewindTo int64) {
+	bm.rewindMux.Lock()
+	bm.rewindTo = rewindTo
```
awrichar, line 228: Should we check if `rewindTo` is already set?

awrichar (changes requested):
Marked one item that I think requires a change.

awrichar (approved):

PR #497, opened 2022-02-08
Pass requestId to DX and remove backend_id from Operation

awrichar:
Follow the convention of all other plugins by passing a request ID into the
connector, rather than reading one from the connector response. This will be
enforced as a requirement for future connectors, so backend_id is removed from
the operations table.

Depends on https://github.com/hyperledger/firefly-dataexchange-https/pull/54

### DIFF
--- a/internal/dataexchange/dxhttps/dxhttps.go
+++ b/internal/dataexchange/dxhttps/dxhttps.go
@@ -44,17 +44,17 @@ type HTTPS struct {
 }
 
 type wsEvent struct {
-	Type      msgType `json:"type"`
-	Sender    string  `json:"sender"`
-	Recipient string  `json:"recipient"`
-	RequestID string  `json:"requestId"`
-	Path      string  `json:"path"`
-	Message   string  `json:"message"`
-	Hash      string  `json:"hash"`
-	Size      int64   `json:"size"`
-	Error     string  `json:"error"`
-	Manifest  string  `json:"manifest"`
-	Info      string  `json:"info"`
+	Type      msgType            `json:"type"`
+	Sender    string             `json:"sender"`
+	Recipient string             `json:"recipient"`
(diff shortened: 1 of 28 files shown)


peterbroadhurst (approved):
Looks good. Noting it has a co-req with https://github.com/hyperledger/firefly-dataexchange-https/pull/54 meaning a hard dependency on upgrading DX with this.

awrichar:
Yep - good news is it won't fail anything without the change (as testified by the E2E tests here), but it will result in a bunch of DX operations being left "pending".

Planning to merge this with a manifest.json change after the DX PR goes in.

internal/dataexchange/dxhttps/dxhttps.go
```diff
@@ -44,17 +44,17 @@ type HTTPS struct {
...
+	Sender    string             `json:"sender"`
+	Recipient string             `json:"recipient"`
+	RequestID string             `json:"requestId"`
+	Path      string             `json:"path"`
+	Message   string             `json:"message"`
+	Hash      string             `json:"hash"`
+	Size      int64              `json:"size"`
+	Error     string             `json:"error"`
+	Manifest  string             `json:"manifest"`
+	Info      fftypes.JSONObject `json:"info"`
```
awrichar, line 57: @peterbroadhurst @gabriel-indik I wanted to call special attention to this change. I don't actually see any dataexchange events that include "info" in their payload, so I don't think it breaks anything. Was this just added as a possible future plug point?

gabriel-indik: @awrichar the Info field is actually expected. After reviewing the latest changes I realized this is something that is going to be coming from Data Exchange. This is, instead of having to deal with each individual type of response Data Exchange produces, the information will always be encapsulated under the "info" JSON object.

PR #499, opened 2022-02-08
Fix batch pin index calculation logic and improve logging

peterbroadhurst:
Next step on #493

### DIFF
--- a/internal/batch/batch_manager.go
+++ b/internal/batch/batch_manager.go
@@ -338,7 +338,7 @@ func (bm *batchManager) newEventNotifications() {
 				return
 			}
 			l.Debugf("New message sequence notification: %d", m)
-			bm.markRewind(m)
+			bm.markRewind(m - 1)
 		case <-bm.ctx.Done():
 			l.Debugf("Exiting due to cancelled context")
 			return
--- a/internal/batch/batch_manager.go
+++ b/internal/batch/batch_manager.go
@@ -51,7 +51,7 @@ func NewBatchManager(ctx context.Context, ni sysmessaging.LocalNodeInfo, di data
 		startupOffsetRetryAttempts: config.GetInt(config.OrchestratorStartupAttempts),
 		dispatchers:                make(map[fftypes.MessageType]*dispatcher),
 		shoulderTap:                make(chan bool, 1),
-		newMessages:                make(chan int64, readPageSize),
+		newMessages:                make(chan int64),
 		sequencerClosed:            make(chan struct{}),
 		retry: &retry.Retry{
 			InitialDelay: config.GetDuration(config.BatchRetryInitDelay),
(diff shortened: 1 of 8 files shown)


internal/batch/batch_manager.go
```diff
@@ -51,7 +51,7 @@ func NewBatchManager(ctx context.Context, ni sysmessaging.LocalNodeInfo, di data
 		startupOffsetRetryAttempts: config.GetInt(config.OrchestratorStartupAttempts),
 		dispatchers:                make(map[fftypes.MessageType]*dispatcher),
 		shoulderTap:                make(chan bool, 1),
-		newMessages:                make(chan int64, readPageSize),
```
peterbroadhurst, line 54: Seemed arbitrary that we had a page size here, as we have a dedicated channel reading from it

internal/batch/batch_manager.go
```diff
@@ -219,18 +219,20 @@ func (bm *batchManager) assembleMessageData(msg *fftypes.Message) (data []*fftyp
...
+	log.L(bm.ctx).Infof("Detected new batch-pinned message %s sequence=%d", msg.Header.ID, msg.Sequence)
 	return data, nil
 }
 
 func (bm *batchManager) markRewind(rewindTo int64) {
 	bm.rewindMux.Lock()
 	// Make sure we only rewind backwards - as we might get multiple shoulder taps
 	// for different message sequences during a single poll cycle.
-	if bm.rewindTo < 0 || rewindTo < bm.rewindTo {
+	previousRewind := bm.rewindTo
```
peterbroadhurst, line 230: This is just to allow logging

internal/events/aggregator.go
```diff
@@ -183,6 +183,20 @@ func (ag *aggregator) getPins(ctx context.Context, filter database.Filter) ([]ff
 	return ls, err
 }
 
+func (ag *aggregator) extractBatchMessagePin(batch *fftypes.Batch, requiredIndex int64) (totalBatchPins int64, msg *fftypes.Message, msgBaseIndex int64) {
```
peterbroadhurst, line 186: The re-write of this logic, is the main thing here.

We trigger from the `pin` - which references the `batch` and the `index`, but we need two other bits of information:
- Which message within the batch lines up with this index
   - Noting that each message might have multiple topics
- What the first index is of this message in the batch
   - This is to account for the new logic from #483 that ensures all pins associated with a message get marked `dispatched=true` together

internal/events/aggregator.go
```diff
@@ -206,23 +219,14 @@ func (ag *aggregator) processPins(ctx context.Context, pins []*fftypes.Pin, stat
 		}
 
 		// Extract the message from the batch - where the index is of a topic within a message
-		var msg *fftypes.Message
-		var i int64 = -1
```
peterbroadhurst, line 210: In the case of `pin.Index == 0` this logic was broken. `msg` and `msgBaseIndex` were incorrect.

peterbroadhurst:
This whole thing as I'm chipping away at it, feels like I'm just coming to the big re-work needed in https://github.com/hyperledger/firefly/issues/421#issuecomment-1019026240 from the other end.

I'm thinking I will just go full steam for that and deal with the consequences.

peterbroadhurst:
Closing this out - in favor of the full rework.

PR #500, opened 2022-02-09
New DX events

gabriel-indik:

### DIFF
--- a/pkg/fftypes/operation.go
+++ b/pkg/fftypes/operation.go
@@ -48,6 +48,8 @@ const (
 	OpStatusPending OpStatus = "Pending"
 	// OpStatusSucceeded the infrastructure runtime has returned success for the operation.
 	OpStatusSucceeded OpStatus = "Succeeded"
+	// OpStatusSucceeded the infrastructure runtime has returned success for the operation.
+	OpStatusAcknowledged OpStatus = "Acknowledged"
 	// OpStatusFailed happens when an error is reported by the infrastructure runtime
 	OpStatusFailed OpStatus = "Failed"
 )
--- a/internal/dataexchange/dxhttps/dxhttps.go
+++ b/internal/dataexchange/dxhttps/dxhttps.go
@@ -65,12 +65,14 @@ const (
 type msgType string
 
 const (
-	messageReceived  msgType = "message-received"
-	messageDelivered msgType = "message-delivered"
-	messageFailed    msgType = "message-failed"
-	blobReceived     msgType = "blob-received"
-	blobDelivered    msgType = "blob-delivered"
-	blobFailed       msgType = "blob-failed"
+	messageReceived     msgType = "message-received"
+	messageDelivered    msgType = "message-delivered"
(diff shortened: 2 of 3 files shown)


gabriel-indik:
Something that was previously discussed but not yet implemented: currently there are only 3 operation status:
1. Pending
2. Succeeded
3. Failed

When sending a message through Data Exchange, the message would go through the following stages:
1. Sent
2. Received
3. Acknowledged
(and Error if there is a communication issue)

Note that a recipient may choose not to acknowledge a message. This is why keeping the status in "pending" after the message is sent/received doesn't seem like the right approach. Instead we could incorporate a new state called either "Acknowledged" or "Confirmed".

pkg/fftypes/operation.go
```diff
@@ -48,6 +48,8 @@ const (
 	OpStatusPending OpStatus = "Pending"
 	// OpStatusSucceeded the infrastructure runtime has returned success for the operation.
 	OpStatusSucceeded OpStatus = "Succeeded"
+	// OpStatusSucceeded the infrastructure runtime has returned success for the operation.
+	OpStatusAcknowledged OpStatus = "Acknowledged"
```
awrichar, line 52: I'm not sure a 4th state is the way to go here. If something "succeeded" but was not "acknowledged", is the operation complete or not? It makes things a bit fuzzy when looking at a UI view of all the operations. What icons do we show?
FireFly_Explorer

Keep in mind that operation statuses also roll up to the transaction status. So whatever we change here needs to also involve a conversation about how transaction status is aggregated - is some operations within the transaction are "succeeded" and others are "acknowledged", what is the overall state of the transaction? Does it have an "acknowledged" state also?

gabriel-indik: An operation that succeeded is considered complete. An operation can only become acknowledged after it succeeded. If an operation becomes acknowledged, its completion doesn't change.

@awrichar I agree with your comments and I too have similar concerns, this is why I brought this item up as needing some conversation. The 4th state is something that was briefly covered in the past with @peterbroadhurst although we didn't settle at that time on what that new state should be.

There are other alternatives we could also consider, such as keeping only 3 states ("pending", "succeeded" and "failed") and adding a boolean "acknowledged". This would be a tradeoff to maintain the current 3 states while allowing to check if an operation has been acknowledged as it would open the possibility for inconsistencies e.g. acknowledged being true while the state is "pending" or "failed".

gabriel-indik: @awrichar a recipient may choose not to acknowledge. From the point of view of the sender, if the message made it to the recipient that should be considered complete. If in addition the recipient chooses to acknowledge the message, then that is additional information on that operation.

gabriel-indik: Ultimately I guess we could leave the operations as-is and allow plug-in specific code to surface acknowledgement information in the user experience. The downside of this approach is that within the context of FireFly one would not be able to check which operations have been acknowledged.

PR #501, opened 2022-02-09
Re-work batch logic for simplicity, efficiency, and restart recovery

peterbroadhurst:
This leads on from the investigation in #421 and then #499

At the moment this is how the batching logic works:

## OLD

batch_logic

- The assembly loop of each batch dispatcher is able to keep accepting work as long as the last batch has been successfully dispatched.
- The assembly loop passes work to the persistence loop, which has as many slots in its channel input as there are slots in the batch.
- The persistence loop keeps continually recording data to the database.
- The offset on the batch manager that's reading from the messages queue, is updated as soon as the assembly loop picks up the message

This design evolved from a similar design in the previous generation of Asset Trail technology.

Given this most recent issue, and some inefficiency and crash recovery challenges that exist in the design after it's port over to FireFly and evolution, I'm planning an updated design.

## NEW

This PR updates it to simplify it (even beyond the previous proposal in #421):

batching_v2

- We remove two of the loops
- Persistence only happens once for each batch in the final stage when the batch is sealed
- Dispatching is synchronous to the assembler
- The assembler only allows one batch to be in assembly before it blocks waiting for the dispatch completion
- The batch manager just has in-memory offset state, and on restart it looks from time-zero again
- We introduce a new `sent` state for messages that go into a batch
- Anything that doesn't reach that state is eligible for re-send after restart (at least once delivery)
- The manager can rewind at any time, because everything will either:
  - Be persisted with a batch ID so skipped
  - Be already in-flight
- The processor keeps track of sequences its flushing, and does duplicate detection
- The processor orders the assembly batch in sequence order

Note this means if things are appearing at once across a batch assembly window of 0.5s, it's very likely all messages will be sent in DB sequence order (even though the DB doesn't guarantee that's the order they'll become visible in).


awrichar:
Explanation seems sound. I'll wait for more of the code to take shape before reviewing though.

pkg/fftypes/transaction.go
```diff
@@ -19,8 +19,10 @@ package fftypes
 type TransactionType = FFEnum
 
 var (
-	// TransactionTypeNone indicates no transaction should be used for this message/batch
+	// TransactionTypeNone deprecreated - replaced by TransactionTypeUnpinned
```
awrichar, line 22: typo: deprecated

pkg/fftypes/transaction.go
```diff
@@ -19,8 +19,10 @@ package fftypes
...
 	TransactionTypeNone TransactionType = ffEnum("txtype", "none")
+	// TransactionTypeUnpinned indicates no transaction should be used for this message/batch
```
awrichar, line 24: I think the wording should indicate that the message will be sent without pinning any evidence to the blockchain.

There _is_ a FireFly transaction (although there's no blockchain transaction) - so want to avoid confusion there.

internal/batch/batch_manager.go
```diff
@@ -132,280 +133,228 @@ func (bm *batchManager) NewMessages() chan<- int64 {
...
+	dispatcherKey := bm.getDispatcherKey(txType, msgType)
+	dispatcher, ok := bm.dispatchers[dispatcherKey]
 	if !ok {
-		return nil, i18n.NewError(bm.ctx, i18n.MsgUnregisteredBatchType, batchType)
+		return nil, i18n.NewError(bm.ctx, i18n.MsgUnregisteredBatchType, dispatcherKey)
 	}
-	dispatcher.mux.Lock()
-	key := fmt.Sprintf("%s:%s:%s[group=%v]", namespace, identity.Author, identity.Key, group)
-	processor, ok := dispatcher.processors[key]
+	name := fmt.Sprintf("%s|%s|%v", namespace, identity.Author, group)
```
awrichar, line 145: Small thing, but maybe there should be a `getProcessorKey` next to `getDispatcherKey`.

internal/batch/batch_manager.go
```diff
@@ -132,280 +133,228 @@ func (bm *batchManager) NewMessages() chan<- int64 {
...
+		msg:  msg,
+		data: data,
 	}
 	processor.newWork <- work
 	return nil
 }
 
-func (bm *batchManager) WaitStop() {
-	<-bm.sequencerClosed
+func (bm *batchManager) reapQuiescing() {
```
awrichar, line 302: Award for the best function name here 🏅

internal/batch/batch_manager.go
```diff
@@ -132,280 +133,228 @@ func (bm *batchManager) NewMessages() chan<- int64 {
...
+			newMessages = true
 		default:
+			checkingMessages = false
 		}
 	}
+	return newMessages
 }
 
-func (bm *batchManager) waitForShoulderTapOrPollTimeout() {
+func (bm *batchManager) waitForShoulderTapOrPollTimeout() (done bool) {
```
awrichar, line 269: Just a spelling thing - but "shoulder tap" has been replaced by "new message", right?

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	bp.assemblyQueue = append([]*batchWork{}, initalWork...)
+	bp.assemblyQueueBytes = batchSizeEstimateBase
+}
+
+// addWork adds the work to the assemblyQueue, and calculates if we have overflowed with this work.
+// We check for duplicates, and add the work in sequence order.
+// This helps in the case for parallel REST APIs all committing to the DB at a similar time.
+// With a sufficient batch size and batch timeout, the batch will still dispatch the messages
+// in DB sequence order (although this is not guaranteed).
+func (bp *batchProcessor) addWork(newWork *batchWork) (full, overflow bool) {
```
awrichar, line 154: I would probably make these `bool` returns into an enum but you know how I roll 🤓

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	newQueue := make([]*batchWork, 0, len(bp.assemblyQueue)+1)
+	added := false
+	skip := false
+	// Check it's not in the recently flushed lish
```
awrichar, line 158: `lish`

internal/batch/batch_manager.go
```diff
@@ -132,280 +133,228 @@ func (bm *batchManager) NewMessages() chan<- int64 {
...
-			l.Debugf("New message sequence notification: %d", m)
-			bm.markRewind(m)
-		case <-bm.ctx.Done():
-			l.Debugf("Exiting due to cancelled context")
-			return
-		}
-		// Do not block sending to the shoulderTap - as it can only contain one
+func (bm *batchManager) newMessageNotification(seq int64) {
+	log.L(bm.ctx).Debugf("Notification of message %d", seq)
+	if (seq - 1) < bm.readOffset {
```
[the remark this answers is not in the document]
peterbroadhurst: > just add 1 to `readOffset` everywhere

Worried this is a big risky change. Going for the comment

internal/batch/batch_processor.go
```diff
@@ -33,63 +34,91 @@ import (
...
 
-type batchProcessorConf struct {
-	Options
-	namespace          string
-	identity           fftypes.Identity
-	group              *fftypes.Bytes32
-	dispatch           DispatchHandler
-	processorQuiescing func()
+// FlushStatus is an object that can be returned on REST queries to understand the status
+// of the batch processor
```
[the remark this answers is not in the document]
peterbroadhurst: Great point - I meant to add the API. Will do that

internal/batch/batch_processor.go
peterbroadhurst: ... yeah... in my opinion with the code as it's structure they are independent things, so we'd end up with 4 states.
So I'm going to stick with the bools here unless you feel strongly.

internal/batch/batch_processor.go
peterbroadhurst: `lush` 👍  🏴󠁧󠁢󠁷󠁬󠁳󠁿

pkg/fftypes/transaction.go
peterbroadhurst: Looking forward to the day we `dep-recreate` something

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	skip := false
+	// Check it's not in the recently flushed lish
+	for _, flushedSequence := range bp.flushedSequences {
+		if newWork.msg.Sequence == flushedSequence {
+			log.L(bp.ctx).Debugf("Ignoring add of recently flushed message %s sequence=%d to in-flight batch assembly %s", newWork.msg.Header.ID, newWork.msg.Sequence, bp.assemblyID)
+			skip = true
+			break
+		}
+	}
+	// Build the new sorted work list, checking there for duplicates too
```
awrichar, line 166: If `skip` is true by this point, we're going to do a bunch of looping to make a copy of `bp.assemblyQueue` and then replace it with a copy of itself. Feels like if `skip` is set at any point, we can just abandon `newQueue` altogether and keep the existing queue intact, right?

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+		if newWork.msg.Sequence == work.msg.Sequence {
+			log.L(bp.ctx).Debugf("Ignoring duplicate add of message %s sequence=%d to in-flight batch assembly %s", newWork.msg.Header.ID, newWork.msg.Sequence, bp.assemblyID)
+			skip = true
+		}
+		if !added && !skip && newWork.msg.Sequence < work.msg.Sequence {
+			newQueue = append(newQueue, newWork)
+			added = true
+		}
+		newQueue = append(newQueue, work)
+	}
```
awrichar, line 177: It also feels like much of this loop could be replaced by a single `append` of the new work and then a call to `sort.Slice()`. Would still need a check for duplicates ahead of that though, so unsure of the ultimate performance impact. Just a thought.

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	if !added && !skip {
+		newQueue = append(newQueue, newWork)
+		added = true
+	}
+	if added {
```
awrichar, line 182: `added` will _always_ be true here unless `skip` was set. Again just feels like a lot of the logic should be skipped early if `skip` gets flipped to true.

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	bp.assemblyQueue = newQueue
+	full = len(bp.assemblyQueue) >= int(bp.conf.BatchMaxSize) || (bp.assemblyQueueBytes >= bp.conf.BatchMaxBytes)
+	overflow = len(bp.assemblyQueue) > 1 && (bp.assemblyQueueBytes > bp.conf.BatchMaxBytes)
+	return full, overflow
+}
+
+func (bp *batchProcessor) startFlush(overflow bool) (id *fftypes.UUID, flushAssembly []*batchWork, byteSize int64) {
+	bp.statusMux.Lock()
+	defer bp.statusMux.Unlock()
+	// Star the clock
```
awrichar, line 195: `Start`

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	// We keep all of the last batch, and up to twice the batch size over time (noting our channel
+	// size is our batch size - so the batch manager cannot get further than that ahead).
+	newFlushedSeqLen := len(flushAssembly) + len(bp.flushedSequences)
+	maxFlushedSeqLen := int(2 * bp.conf.BatchMaxSize)
+	if newFlushedSeqLen > maxFlushedSeqLen && maxFlushedSeqLen > len(bp.flushedSequences) {
+		newFlushedSeqLen = maxFlushedSeqLen
+	}
+	newFlushedSequnces := make([]int64, newFlushedSeqLen)
+	for i, fs := range flushAssembly {
+		newFlushedSequnces[i] = fs.msg.Sequence
```
awrichar, line 218: `newFlushedSequences`

internal/batch/batch_processor.go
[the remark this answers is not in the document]
peterbroadhurst: `GET` `/api/v1/status/batchmanager` now added

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+	}
+	for i := 0; i < newFlushedSeqLen-len(flushAssembly); i++ {
```
awrichar, line 220: Can you just iterate `i := len(flushAssembly); i < newFlushedSeqLen`?

internal/batch/batch_processor.go
peterbroadhurst: So I think this ends up as more code in the end of the day. We have to have all the branches that we have today, but _also_ create a custom function do pass into the `sort.Slice()` to do the comparison.

As for whether it's more performant in practice to do a single traversal and allocate memory for a new pointer array each time, vs only re-allocate the array when we add and do a sort, I'm not sure.

My gut was re-allocating the array each time and doing a single pass optimizes for the case where we're adding work - which is the most common case.

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+		newFlushedSequnces[i+len(flushAssembly)] = bp.flushedSequences[i]
```
awrichar, line 224: Are we intentionally keeping the _first_ elements in `bp.flushedSequences` and dropping ones off the end? Or do we actually want to preserve the tail and drop the earliest elements?

internal/batch/batch_processor.go
peterbroadhurst: this is a good point 👍

peterbroadhurst: That's not actually true, if (as is the most common case) we're adding to the end of the list.

peterbroadhurst: I will do a simplification experiment, to see if I can make the add logic based on if the item is > than the current one... which I think would always add

internal/batch/batch_processor.go
awrichar: Also if you initialize the slice with `make([]int64, len(flushAssembly), newFlushedSeqLen)`, then I think maybe you could get fancy here by replacing the loop with an `append`... not sure what ultimately reads cleaner though.

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,249 @@ func (bw *batchWork) estimateSize() int64 {
...
+			if overflow {
+				batchTimeout = time.NewTimer(bp.conf.BatchTimeout)
+			}
 
-		// Don't include the sealing time in the duration
-		batchFull := overflowedWork != nil || batchSize >= bp.conf.BatchMaxSize
-		l.Debugf("Assembly batch loop: Size=%d Full=%t Bytes=%.2fkb (est) Overflow=%t", batchSize, batchFull, float64(batchPayloadEstimate)/1024, overflowedWork != nil)
+			err := bp.flush(overflow)
+			if err != nil {
+				l.Tracef("Batch processor shutting down: %s", err)
```
awrichar, line 331: Only a trace log here? Should flush errors be more visible?

internal/batch/batch_processor.go
awrichar: Yea, in the common case `added = true` will be hit up above on line 180, so it will be true by this point.

peterbroadhurst: ... that went badly - just leaving the skip part

peterbroadhurst: ok - early return was the solution 👍  (`overflow` and `full` can be assumed to be `false` if we didn't add anything)

internal/batch/batch_processor.go
peterbroadhurst: 🌟

internal/batch/batch_processor.go
awrichar: Yea, I think with the other optimizations added to abort early in the case of a duplicate, this now reads pretty clean and doesn't need any further tweaking.

internal/database/sqlcommon/event_sql.go
```diff
@@ -49,6 +49,14 @@ func (s *SQLCommon) InsertEvent(ctx context.Context, event *fftypes.Event) (err
 	}
 	defer s.rollbackTx(ctx, tx, autoCommit)
 
+	// This is special to events - we take the cost of a full table lock on the events table, so
+	// that nobody can add rows that increment the sequence, until our transaction has committed.
+	// This allows us to rely on the sequence to always be increasing, even when writing events
+	// concurrently (it does not guarantee we won't get a gap in the sequences).
```
awrichar, line 55: Wonder if it's worth noting this in the database `plugin.go` as well, since we are now requiring any database plugin to provide a stable commit ordering for `InsertEvent` specifically.

internal/batch/batch_processor.go
peterbroadhurst: Because the existing `flushedSequences` might be a bunch of small chunks, the right change here seemed to be to reverse the order of the array elements as we add them (rather than as we remove them).

That way the chunks are ordered newest to oldest, and the elements are ordered newest to oldest.

internal/batch/batch_processor.go
peterbroadhurst: The only way errors can percolate out of `flush` without infinite retry, are if we're shutting down. But will bump to warn.

internal/privatemessaging/message.go
```diff
@@ -107,7 +99,12 @@ func (s *messageSender) setDefaults() {
...
 		s.msg.Header.Type = fftypes.MessageTypePrivate
 	}
-	if s.msg.Header.TxType == "" {
+	switch s.msg.Header.TxType {
+	case fftypes.TransactionTypeUnpinned, fftypes.TransactionTypeNone:
+		// "unpinned" used to be called "none" (before we introduced batching + a TX on unppinned sends)
+		s.msg.Header.TxType = fftypes.TransactionTypeUnpinned
+	default:
+		// the only other valid option is "batch_pin"
 		s.msg.Header.TxType = fftypes.TransactionTypeBatchPin
```
awrichar, line 108: Should we yell at them for setting an invalid txtype? Or just quietly set to the default like this?

Also might consider adding similar pre-validation to `broadcast/message.go` - can't remember if/when that case actually fails due to an invalid txtype.

peterbroadhurst: Going to go for the default approach, but do the same for Broadcast

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,245 @@ func (bw *batchWork) estimateSize() int64 {
...
+	// size is our batch size - so the batch manager cannot get further than that ahead).
+	newFlushedSeqLen := len(flushAssembly) + len(bp.flushedSequences)
+	maxFlushedSeqLen := int(2 * bp.conf.BatchMaxSize)
+	if newFlushedSeqLen > maxFlushedSeqLen && maxFlushedSeqLen > len(bp.flushedSequences) {
+		newFlushedSeqLen = maxFlushedSeqLen
+	}
+	newFlushedSequences := make([]int64, newFlushedSeqLen)
+	for i := 0; i < len(flushAssembly); i++ {
+		// Add in reverse order - so we can trim the end of it off later (in the next block) and keep the newest
+		newFlushedSequences[len(flushAssembly)-i-1] = flushAssembly[i].msg.Sequence
```
awrichar, line 214: So... we are adding to the front of a list, in reverse order, and then ultimately trimming the end of the list.

It feels like we could add to the end of the list in order, and then trim the front, and it might be more understandable. However, I do think the current logic works, so it's not mandatory to re-spin this again if we want to defer for now.

awrichar (approved):
Marking approved with one more optional thought for a possible simplification.

Thanks for this; feels like a huge step forward for the message batching.

internal/batch/batch_processor.go
```diff
@@ -101,103 +130,254 @@ func (bw *batchWork) estimateSize() int64 {
...
+
+	// We keep as much of the END of the existing set as we can, and shift it forwards.
+	// Then we add the whole of the current flush set after that
+	combinedLen := len(flushAssembly) + len(bp.flushedSequences)
+	newLength := combinedLen
+	if combinedLen > maxFlushedSeqLen {
+		newLength = maxFlushedSeqLen
+	}
+	dropLength := combinedLen - newLength
+	retainLength := len(bp.flushedSequences) - dropLength
```
awrichar, line 202: Thanks, personally I feel like the extra vars and comments have made this MUCH easier to follow.

PR #503, opened 2022-02-09
Prometheus metrics

eberger727:
Prometheus metrics for the following events:

Broadcast & Private Message:
- Submitted
- Confirmed
- Rejected
- Histogram

Token Mint, Transfer, and Burn:
- Submitted
- Confirmed
- Rejected
- Histogram


internal/privatemessaging/privatemessaging.go
```diff
@@ -93,6 +96,8 @@ func NewPrivateMessaging(ctx context.Context, di database.Plugin, im identity.Ma
 		},
 		opCorrelationRetries:  config.GetInt(config.PrivateMessagingOpCorrelationRetries),
 		maxBatchPayloadLength: config.GetByteSize(config.PrivateMessagingBatchPayloadLimit),
+		metricsEnabled:        config.GetBool(config.MetricsEnabled),
+		metrics:               mm,
```
awrichar, line 100: Maybe the metrics manager should be responsible for exposing the "are metrics enabled" flag, instead of each component reading it directly from the config?

internal/events/operation_update.go
```diff
@@ -42,6 +42,7 @@ func (em *eventManager) operationUpdateCtx(ctx context.Context, operationID *fft
 	// Special handling for OpTypeTokenTransfer, which writes an event when it fails
 	if op.Type == fftypes.OpTypeTokenTransfer && txState == fftypes.OpStatusFailed {
 		event := fftypes.NewEvent(fftypes.EventTypeTransferOpFailed, op.Namespace, op.ID)
+		// TODO: Need to determine transfer type and track metric
```
awrichar, line 45: I think this is now available in `op.Input.getString("type")`

awrichar: Or you can use the helpers in `token_inputs.go` to extract all transfer details from the Input. Maybe that's better...

internal/orchestrator/orchestrator_test.go
```diff
@@ -454,6 +459,12 @@ func TestInitDataComponentFail(t *testing.T) {
 	assert.Regexp(t, "FF10128", err)
 }
 
+func TestInitMetricsComponentFailGracefully(t *testing.T) {
+	or := newTestOrchestrator()
+	or.metrics = nil
+	or.initComponents(context.Background())
```
awrichar, line 465: Maybe should check for `NoError` on this return?

awrichar: The title of the test seems wrong too. This isn't a failure test right? There is no way metrics init can fail. It just succeeds.

internal/assets/token_transfer_test.go
```diff
@@ -91,8 +95,10 @@ func TestMintTokensSuccess(t *testing.T) {
 	mti.On("MintTokens", context.Background(), mock.Anything, "F1", &mint.TokenTransfer).Return(nil)
 	mth.On("SubmitNewTransaction", context.Background(), "ns1", fftypes.TransactionTypeTokenTransfer).Return(fftypes.NewUUID(), nil)
 	mdi.On("InsertOperation", context.Background(), mock.Anything).Return(nil)
-
+	mmi.On("IsMetricsEnabled").Return(true)
```
peterbroadhurst, line 98: Just as an idea for the future, might be worth adding `enableMetrics` to the `newTestAssets` call to avoid this boilerplate on so many tests

peterbroadhurst (approved):
Just one thought - not a requirement at all for you to change the PR, but just something that might make a change like this easier in the future.

internal/assets/token_transfer_test.go
eberger727: Good idea. There is now a `newTestAssets` and `newTestAssetsWithMetrics`

hfuss (approved):
Very quickly glanced and lgtm, talked with @eberger727 and it would be nice to include tags on the messages and token transfer counters to keep track of number of messages / transfers btwn pairs of members i.e. `to` and `from` tags on each counter.

Not a requirement for this massive improvement around metrics, but will open a follow up issue for it.

PR #505, opened 2022-02-10
Rename data exchange plugin to "ffdx"

### LINKED ISSUE #259: Rename dx "https" plugin
As discussed on #202 in relation to tokens, the name "https" is too generic for a plugin.
A plugin not only implies a particular protocol (or protocols), but also a very specific set
of methods and events implemented on that protocol.

The dx plugin should also be renamed - possibly something like "ffdx" or "ffdata"?

awrichar:
Resolves #259

Open to other names here, but we landed on "fftokens" for the tokens plugin, so this seemed reasonable.
The name just needs to be globally unique in FireFly across all plugins, and identifiable as a data exchange plugin
vs. some other type of plugin.

### DIFF
--- a/internal/dataexchange/dxfactory/factory.go
+++ b/internal/dataexchange/dxfactory/factory.go
@@ -20,18 +20,24 @@ import (
 	"context"
 
 	"github.com/hyperledger/firefly/internal/config"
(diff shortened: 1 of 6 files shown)


peterbroadhurst (approved):
Thanks for including the migration here

PR #507, opened 2022-02-10
Add "initEnabled" config to DX

awrichar:
When enabled, FireFly will post to `/api/v1/init` on the DX connector with a list of all current nodes,
before initial connect of the websocket and before each reconnect.


internal/dataexchange/ffdx/config.go
```diff
@@ -24,9 +24,12 @@ import (
 const (
 	// DataExchangeManifestEnabled determines whether to require+validate a manifest from other DX instances in the network. Must be supported by the connector
 	DataExchangeManifestEnabled = "manifestEnabled"
+	// DataExchangeReInitEnabled instructs FireFly to always post all current nodes to the /init API before connecting or reconnecting to the connector
+	DataExchangeReInitEnabled = "reInitEnabled"
```
peterbroadhurst, line 28: I think this should be `initEnabled` - as you should always set this whenever you have a `ffdx` plugin that supports init. So we don't need to indicate the `re` (which is a bit confusing)

peterbroadhurst: Another spelling could be:
`"apiVersion": "1.1"`

peterbroadhurst (commented):
Would like to get your spelling thoughts before we merge @awrichar 
However, the code looks great

internal/dataexchange/ffdx/config.go
awrichar: I think `initEnabled` is appropriate. Personally I think it should be independently toggleable regardless of the plugin version.

pkg/wsclient/wsclient.go
```diff
@@ -204,6 +209,15 @@ func (w *wsClient) connect(initial bool) error {
 		if w.closed {
 			return false, i18n.NewError(w.ctx, i18n.MsgWSClosing)
 		}
+
+		retry = !initial || attempt < w.initialRetryAttempts
```
awrichar, line 213: Want to highlight since it's not obvious in the diff - the logic around `initialRetryAttempts` appears to have been broken in the original code (so that it would only try once to connect on startup before failing). It's tangential to the bulk of the changes in this PR, but it should be working now (with unit test added to demonstrate).

peterbroadhurst (approved):
👍

PR #511, opened 2022-02-11
Add endpoint to automatically generate an FFI

nguyer:
This PR adds an endpoint to automatically generate an FFI. The implementation is handled by the blockchain plugin. Right now, only the Ethereum plugin implements this functionality, but other blockchain plugins in the future could as well, if they have their own DSL to describe a smart contract.

Resolves https://github.com/hyperledger/firefly/issues/397

### DIFF
--- a/internal/apiserver/route_post_contract_interface_generate.go
+++ b/internal/apiserver/route_post_contract_interface_generate.go
@@ -1,0 +1,47 @@
+// Copyright © 2022 Kaleido, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apiserver
+
+import (
+	"net/http"
+
+	"github.com/hyperledger/firefly/internal/config"
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var postContractInterfaceGenerate = &oapispec.Route{
+	Name:   "postGenerateContractInterface",
+	Path:   "namespaces/{ns}/contracts/interfaces/generate",
+	Method: http.MethodPost,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+	},
+	QueryParams: []*oapispec.QueryParam{
+		{Name: "confirm", Description: i18n.MsgConfirmQueryParam, IsBool: true, Example: "true"},
+	},
+	FilterFactory:   nil,
+	Description:     i18n.MsgTBD,
+	JSONInputValue:  func() interface{} { return fftypes.JSONAnyPtr("{}") },
+	JSONInputMask:   nil,
+	JSONOutputValue: func() interface{} { return &fftypes.FFI{} },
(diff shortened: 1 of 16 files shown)


internal/apiserver/route_post_contract_interface_generate.go
```diff
@@ -0,0 +1,47 @@
...
+
+var postContractInterfaceGenerate = &oapispec.Route{
+	Name:   "postGenerateContractInterface",
+	Path:   "namespaces/{ns}/contracts/interfaces/generate",
+	Method: http.MethodPost,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+	},
+	QueryParams: []*oapispec.QueryParam{
+		{Name: "confirm", Description: i18n.MsgConfirmQueryParam, IsBool: true, Example: "true"},
```
awrichar, line 36: Seems like this is unused?

nguyer: Yep. Good catch, thanks.

internal/contracts/manager.go
```diff
@@ -53,6 +53,7 @@ type Manager interface {
 	GetContractSubscriptionByNameOrID(ctx context.Context, ns, nameOrID string) (*fftypes.ContractSubscription, error)
 	GetContractSubscriptions(ctx context.Context, ns string, filter database.AndFilter) ([]*fftypes.ContractSubscription, *database.FilterResult, error)
 	DeleteContractSubscriptionByNameOrID(ctx context.Context, ns, nameOrID string) error
+	GenerateFFI(ctx context.Context, ns, name, version string, input *fftypes.JSONAny) (*fftypes.FFI, error)
```
awrichar, line 56: Any reason we pass name and version here? Seems like the only usage of this method always sets them to empty strings.

test/data/assetcreator/chaincode/smartcontract.go
```diff
@@ -24,3 +24,11 @@ func (s *SmartContract) CreateAsset(ctx contractapi.TransactionContextInterface,
 	ctx.GetStub().SetEvent("AssetCreated", assetJSON)
 	return ctx.GetStub().PutState(name, assetJSON)
 }
+
+func (s *SmartContract) GetAsset(ctx contractapi.TransactionContextInterface, name string) (string, error) {
```
awrichar, line 28: No objections to this being added to the chaincode... but was it intended to be used in a test or something? Seems unused at this point.

awrichar (approved):

test/data/assetcreator/chaincode/smartcontract.go
nguyer: Ah this is being used in my other branch for additional tests :) It shouldn't have been in this commit. I'll remove it now for clarity.

internal/contracts/manager.go
nguyer: Yeah... maybe worth a quick chat about this one... I wrote the backend first and thought that it would make sense to pass in the name, version, etc. so it could be filled in on the generated FFI. Then I created the API endpoint and couldn't think of a good place to specify them when calling the API, because the payload of that request right now is entirely an ABI array.

One option would be to modify the endpoint to wrap the ABI elements in another object so that request would look like:

```json
{
  "name": "...",
  "version": "...",
  "abi": [...]
}
```

internal/contracts/manager.go
```diff
@@ -592,3 +593,10 @@ func (cm *contractManager) checkParamSchema(ctx context.Context, input interface
 	}
 	return nil
 }
+
+func (cm *contractManager) GenerateFFI(ctx context.Context, ns string, generationRequest *fftypes.FFIGenerationRequest) (*fftypes.FFI, error) {
+	if generationRequest.Namespace == "" {
+		generationRequest.Namespace = ns
```
awrichar, line 599: I think in general we ignore any nested Namespace value in an input body (ie list it in "ignored" fields so that Swagger won't generate it), and always write/overwrite with the one from the path.

awrichar (approved):
Looks good; I think the new input body format is a lot better (and more future-proof). One minor comment left above.

internal/contracts/manager.go
nguyer: Ah I didn't realize that. I did the exact opposite here. I'll switch it around to match the existing pattern. Thanks!

PR #517, opened 2022-02-14
Add Operations Manager, wrap all operations, and add retry functionality

awrichar:
First step toward https://github.com/hyperledger/firefly-fir/pull/10


internal/operations/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	om := &operationsManager{
+		ctx:      ctx,
+		database: di,
+		txHelper: txcommon.NewTransactionHelper(di),
+		tokens:   ti,
+	}
+	return om, nil
+}
+
+func (om *operationsManager) StartOperation(ctx context.Context, op *fftypes.Operation) error {
```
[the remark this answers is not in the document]
peterbroadhurst: Agree we can't afford to have repeated reads on the critical path, if the calling code already has all this in hand.

I also agree that a suite of E2E tests would be the most durable way of testing these code paths, but clearly that comes with an initial cost of introducing some systematic way of introducing failures at the right phase.

I wonder if a cost-effective compromise would be a REST API call for `GET` `/{ns}/transactions/{txid}operations/{opid}/inputs` that the E2E test uses to confirm the JSON matches what's expected. This wouldn't need any guards around it.

Even typing this, I understand it's complicated - because this would mean only immutable parts of data objects are used to construct the inputs. e.g. if the operation succeeds, and the state changes, it's still possible to generate the _original_ inputs of the operation. This feels "right" in a blockchain centric system, but I understand in practice there could still be complexities.

A rambling thought on the code structure, that I don't feel strongly about but might be useful in your ideas:
- An defined input structure for each type of operation, containing the source types that are used to aggregate together the input
- The assembly code that does the mapping is shared between submit and retry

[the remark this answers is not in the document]
peterbroadhurst: After I wrote this, I realized that today we have a DB persisted `inputs` field on the operation itself.

I suddenly feel like I'm missing something obvious. Why are the inputs not always available on re-submission, simply because at the submission time they were stored in the DB?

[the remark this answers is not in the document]
peterbroadhurst: Makes sense. Particularly for something like an IPFS batch, where we're currently trying to avoid writing the batch to the DB once... and this might end up with us doing it twice if we're not careful.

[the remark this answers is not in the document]
awrichar: Had a bit of a further chat with @peterbroadhurst 1:1, so for the record:
Noted that there definitely are two phases here ("loading/preparing" the needed data from the operation input and any related objects, and then actually "running" the operation). The loading part can be skipped on the initial trigger if the caller already has all the needed information pre-built. Already agreed on this part.

We further identified that the "loading" and "preparing" could be split apart, and we should define some standard format for the "prepared" operation. It should be a serializable object that expresses something at the granularity of "this wraps an HTTP GET to /xyz with params A, B, C". Exact formatting may be at the discretion of the plugin, but each operation should have a "prepare" and a "run" action. Defining an intermediate format here will be useful for debugging, E2E testing, and eventually for more robust retry flows such as "prepare-modify-run".

awrichar:
I've continued working this so it now encompasses all operations and actually adds the `/retry` route for operations. Unit tests and basic manual verification look good. This now fully implements https://github.com/hyperledger/firefly-fir/pull/10, except for the unresolved questions listed there and the E2E tests.

db/migrations/postgres/000061_add_operation_retry.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE operations DROP COLUMN retry_id;
```
peterbroadhurst, line 2: Just a note that you'll need to get a different slot (think currently 70 is next after PRs in the pipe)

internal/operations/manager.go
```diff
@@ -0,0 +1,145 @@
...
+	"context"
+
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/log"
+	"github.com/hyperledger/firefly/pkg/database"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+	"github.com/hyperledger/firefly/pkg/tokens"
+)
+
+type OperationHandler interface {
```
peterbroadhurst, line 29: It's a bit vague feedback, as I can't point out a specific need - but I wonder if a `Name()` function would be useful for debugging / error logging.

internal/operations/manager.go
```diff
@@ -0,0 +1,145 @@
...
+		database: di,
+		tokens:   ti,
+		handlers: make(map[fftypes.OpType]OperationHandler),
+	}
+	return om, nil
+}
+
+func (om *operationsManager) RegisterHandler(handler OperationHandler, ops []fftypes.OpType) {
+	for _, opType := range ops {
+		om.handlers[opType] = handler
```
peterbroadhurst, line 63: Related to other comment above, maybe a `Debugf("OpType=%s registered to handler %s", opType, handler.Name())` is the first use we could make of a `Name()` function for the manager.

internal/operations/manager.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (om *operationsManager) writeOperationSuccess(ctx context.Context, opID *fftypes.UUID) {
+	if err := om.database.ResolveOperation(ctx, opID, fftypes.OpStatusSucceeded, "", nil); err != nil {
+		log.L(ctx).Errorf("Failed to update operation %s: %s", opID, err)
+	}
+}
+
+func (om *operationsManager) writeOperationFailure(ctx context.Context, opID *fftypes.UUID, err error) {
+	if err := om.database.ResolveOperation(ctx, opID, fftypes.OpStatusFailed, err.Error(), nil); err != nil {
+		log.L(ctx).Errorf("Failed to update operation %s: %s", opID, err)
```
peterbroadhurst, line 143: As above: Think it would be good to write the full data of the operation here - particularly the outputs - as they would have been lost.

pkg/database/plugin.go
```diff
@@ -199,6 +199,9 @@ type iOperationCollection interface {
 	// ResolveOperation - Resolve operation upon completion
 	ResolveOperation(ctx context.Context, id *fftypes.UUID, status fftypes.OpStatus, errorMsg string, output fftypes.JSONObject) (err error)
```
peterbroadhurst, line 200: Now we have a generic `UpdateOperation` - I wonder if it would be more consistent to remove the `ResolveOperation` from the DB layer as that's just a thin wrapper:
https://github.com/hyperledger/firefly/blob/60e00ff8041a81b5e038ec53f23a7a430f058a09/internal/database/sqlcommon/operation_sql.go#L178-L186

peterbroadhurst (changes requested):
Really think this worked out well. A very clear template for what an operation means in the codebase, that's easy to extend 👍 

Some minor comments to consider @awrichar

db/migrations/postgres/000061_add_operation_retry.down.sql
awrichar: ➡️ ➡️ ➡️ ➡️ ➡️

internal/contracts/operations.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+type blockchainInvokeData struct {
+	Request *fftypes.ContractCallRequest `json:"request"`
+}
+
+func addBlockchainInvokeInputs(op *fftypes.Operation, req *fftypes.ContractCallRequest) (err error) {
+	var reqJSON []byte
+	if reqJSON, err = json.Marshal(req); err == nil {
```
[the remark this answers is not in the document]
awrichar: Agreed. I think we should get this and other pending items merged, and then I'm planning to take another look at that piece of work.

internal/operations/manager.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return op, om.RunOperation(ctx, po)
+}
+
+func (om *operationsManager) writeOperationSuccess(ctx context.Context, opID *fftypes.UUID) {
+	if err := om.database.ResolveOperation(ctx, opID, fftypes.OpStatusSucceeded, "", nil); err != nil {
+		log.L(ctx).Errorf("Failed to update operation %s: %s", opID, err)
```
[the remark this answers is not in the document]
awrichar: We don't have any path that writes operation outputs here. I'm not sure there's any more data to capture beyond what is in the log. Need to think further to be sure there's no time we _should_ be capturing outputs here...

internal/operations/manager.go
awrichar: Same as above - the only "outputs" from a synchronous failure are in the form of a Go `error` which is logged here. I'm not sure there's anything else useful to log... unless it's a sign that we're not returning enough info from some other layer.

internal/privatemessaging/operations.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return nodeID, blobHash, err
+}
+
+func addBatchSendInputs(op *fftypes.Operation, nodeID *fftypes.UUID, groupHash *fftypes.Bytes32, batchID *fftypes.UUID, manifest string) {
+	op.Input = fftypes.JSONObject{
+		"node":     nodeID.String(),
+		"group":    groupHash.String(),
+		"batch":    batchID.String(),
+		"manifest": manifest,
```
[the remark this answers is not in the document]
awrichar: This is the operation input, not the output. I tried to preserve it alongside the other fields here, since it's already used in a few places.

pkg/database/plugin.go
awrichar: Yea, was thinking the same thing. I can do this.

internal/operations/manager.go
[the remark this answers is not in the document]
peterbroadhurst: Sorry @awrichar - I simply meant _in_ the log itself. With a `%+v` insert.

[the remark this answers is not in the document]
peterbroadhurst: Ok - I understand now the `ResolveOperation` call is only modifying one field on the operation, it's not the call that does the full update with the outputs etc.

peterbroadhurst (approved):
I've marked approval @awrichar as I think you were proposing this goes in at this point, and the comments are taken as input for future work in other changes.

pkg/database/plugin.go
awrichar: After looking at this again, there are a lot of places where we invoke `ResolveOperation`, and this change would take all of those from 1 line to 5 lines. So it would be more consistent, but more verbose as well. I slightly prefer keeping the concise helper.

Could also move the helper somewhere other than the database layer. For instance, it could be on the Operations Manager (but would then need to add a dependency from Event Manager on Operations Manager).

internal/operations/cache.go
```diff
@@ -0,0 +1,84 @@
...
+
+func getCacheKey(op *fftypes.Operation) (string, error) {
+	opCopy := &fftypes.Operation{
+		Namespace:   op.Namespace,
+		Transaction: op.Transaction,
+		Type:        op.Type,
+		Plugin:      op.Plugin,
+		Input:       op.Input,
+	}
+	key, err := json.Marshal(opCopy)
```
peterbroadhurst, line 49: This feels like extra evidence to the point that we should treat the inputs as references, rather than full data wherever possible. As this is going to be a large string cache key that we are marshaling.

peterbroadhurst: ... as an aside I still think serialized JSON would be a more efficient cache key than a SHA etc.

peterbroadhurst (approved):
Thanks for adding the caching. Like that implementation, and think it's great we'll have de-dup on retry within the batch processor.

internal/operations/cache.go
awrichar: Cool, that was my inkling as well but wanted to have someone else back me up. But agreed it's yet another reason to minimize the size of "Input" as much as is practical.

PR #521, opened 2022-02-14
Provide Migration / Registration / Debugging Utilities in Docker Image

hfuss:
This would allow us remove the bits in the [helm chart](https://github.com/hyperledger/firefly-helm-charts/blob/main/charts/firefly/scripts/ff-register-contracts.sh#L21) that download `apk` packages at runtime. This is a best practice for Docker images and would allow enterprises using the chart / images to use them within their networks where they do not allow for public internet access.

Obviously the chart could simply use different images (however for migrations we take advantage of the fact that the migration are bundled into the firefly image), however I wanted to get the community's feedback on taking the hit in the image size being slightly larger for the sake of operational simplicity.


peterbroadhurst (approved):
Thanks @hfuss - I agree these utilities seem to tip clearly over the usefulness vs. bloat compromise.

I could imagine use maintaining two tag prefixes at some point in FireFly's future, for a runtime vs. utility image. However, I don't think that's justified yet.

PR #525, opened 2022-02-14
Add new DX events

gabriel-indik:


internal/dataexchange/ffdx/ffdx.go
```diff
@@ -347,6 +349,8 @@ func (h *FFDX) eventLoop() {
 			case messageFailed:
 				err = h.callbacks.TransferResult(msg.RequestID, fftypes.OpStatusFailed, fftypes.TransportStatusUpdate{Error: msg.Error})
 			case messageDelivered:
+				err = h.callbacks.TransferResult(msg.RequestID, fftypes.OpStatusSucceeded, fftypes.TransportStatusUpdate{})
```
peterbroadhurst, line 352: If this is received when `manifest` mode is `true` - should this not still be `OpStatusPending`?

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -366,6 +370,11 @@ func (h *FFDX) eventLoop() {
 				} else {
 					err = h.callbacks.BLOBReceived(msg.Sender, *hash, msg.Size, msg.Path)
 				}
+			case blobAcknowledged:
+				err = h.callbacks.TransferResult(msg.RequestID, fftypes.OpStatusSucceeded, fftypes.TransportStatusUpdate{
```
peterbroadhurst, line 374: I don't think we have any handling for `blob` generation of a `manifest` - so the change to fully support this might be a bit larger

peterbroadhurst (commented):
Couple of questions here

internal/dataexchange/ffdx/ffdx.go
gabriel-indik: Good point, the logic has been updated to keep the operation status as "Pending" at this stage.

internal/dataexchange/ffdx/ffdx.go
gabriel-indik: Added the logic to store the blob hash as the operation input. This information is later used to check the data coming back from the acknowledgement, similar to how manifests are checked for batch messages.

PR #529, opened 2022-02-14
Support for token approvals

shorsher:
Token approvals are now first class in FireFly. The tokens plugin has
been extended to support submitting & querying token approvals within your
network.

Approvals must be associated with a token pool. In the case of the ERC1155
connector, which only supports setApprovalForAll, an approval on one token
pool will be propagated to every pool.

* Queryable via GET /approval & /approval/{id}
* Submitted via POST /approval


db/migrations/postgres/000059_create_tokenapproval_table.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+DROP TABLE IF EXISTS tokenapproval;
```
awrichar, line 2: Note that your up/down migration filenames should have the same number prefix

internal/apiserver/route_get_token_approvals.go
```diff
@@ -0,0 +1,55 @@
...
+
+var getTokenApprovals = &oapispec.Route{
+	Name:   "getTokenApprovals",
+	Path:   "namespaces/{ns}/tokens/approvals",
+	Method: http.MethodGet,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+	},
+	QueryParams: []*oapispec.QueryParam{
+		{Name: "fromOrTo", Description: i18n.MsgTBD},
```
awrichar, line 37: I don't think this is a valid query param for this route

internal/syncasync/sync_async_bridge.go
```diff
@@ -59,6 +61,7 @@ const (
 	messageReply
 	tokenPoolConfirm
 	tokenTransferConfirm
+	tokenApproveConfirm
```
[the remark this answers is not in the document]
shorsher: good catch, thank you!

internal/apiserver/route_get_token_approvals.go
awrichar: Still think this needs to be removed?

test/e2e/tokens_test.go
```diff
@@ -69,6 +69,21 @@ func (suite *TokensTestSuite) TestE2EFungibleTokensAsync() {
...
+		},
+	}
+	approvalOut := TokenApproval(suite.T(), suite.testState.client1, approval, false)
+
+	waitForEvent(suite.T(), received1, fftypes.EventTypeApprovalConfirmed, approvalOut.LocalID)
+	approvals := GetTokenApprovals(suite.T(), suite.testState.client1, poolID)
+	assert.Equal(suite.T(), 1, len(approvals))
+	assert.Equal(suite.T(), "erc1155", approvals[0].Connector)
+	assert.Equal(suite.T(), true, approvals[0].Approved)
+
```
awrichar, line 86: Would be good to actually exercise the token approval, ie transfer a token on someone else's behalf - if it's not too difficult to add/extend a test that does it.

awrichar (approved):
Overall this looks great. Marking approved with just a few suggestions inline above.

Also please double-check on the reported coverage drop (might just need another sync from main). You'll need a manifest bump to pick up the erc1155 connector changes too.

internal/apiserver/route_get_token_approvals.go
shorsher: :)

Issue #489, opened 2022-02-07
Rename "Public Storage" to "Shared Storage"


awrichar:
One of the main plugins is titled "Public Storage", but this name continues to cause confusion, as it implies that the data will be available publicly (ie outside of the FireFly network).

Suggest renaming to "Shared Storage" to more accurately describe the nature of this component, as being used for off-chain data that is to be shared with all participants of the network.

PR #538, opened 2022-02-21
Rename Public Storage to Shared storage

nguyer:
Resolves https://github.com/hyperledger/firefly/issues/489

This PR globally renames "public storage" to "shared storage" in all parts of the code. If the orchestrator cannot find a valid "sharedstorage" plugin, it falls back and looks for a "publicstorage" plugin for now, so existing config files should still work with this version.


pkg/fftypes/operation.go
```diff
@@ -26,8 +26,8 @@ var (
 	OpTypeBlockchainBatchPin OpType = ffEnum("optype", "blockchain_batch_pin")
 	// OpTypeBlockchainInvoke is a smart contract invoke
 	OpTypeBlockchainInvoke OpType = ffEnum("optype", "blockchain_invoke")
-	// OpTypePublicStorageBatchBroadcast is a public storage operation to store broadcast data
-	OpTypePublicStorageBatchBroadcast OpType = ffEnum("optype", "publicstorage_batch_broadcast")
+	// OpTypeSharedStorageBatchBroadcast is a shared storage operation to store broadcast data
+	OpTypeSharedStorageBatchBroadcast OpType = ffEnum("optype", "sharedstorage_batch_broadcast")
```
nguyer, line 30: @peterbroadhurst @awrichar I do question whether renaming this will cause problems with existing deployments that already have operations of type `publicstorage_batch_broadcast` in the DB. Should we do something differently here?

README.md
```diff
@@ -192,7 +192,7 @@ Directories:
   │       │
   │       │   ┌───────────────┐  - Broadcast of data to all parties in the network
   │       ├───┤ broadcast [Bm]│    * Implements dispatcher for batch component
-  │       │   │ manager       |    * Integrates with public storage interface [Ps] plugin
+  │       │   │ manager       |    * Integrates with shared storage interface [Ps] plugin
```
awrichar, line 195: Do we change the "Ps" abbreviation as well?

internal/config/config.go
```diff
@@ -206,7 +206,9 @@ var (
 	OrgDescription = rootKey("org.description")
 	// OrchestratorStartupAttempts is how many time to attempt to connect to core infrastructure on startup
 	OrchestratorStartupAttempts = rootKey("orchestrator.startupAttempts")
-	// PublicStorageType specifies which public storage interface plugin to use
+	// SharedStorageType specifies which shared storage interface plugin to use
+	SharedStorageType = rootKey("sharedstorage.type")
+	// PublicStorageType specifies which shared storage interface plugin to use
```
awrichar, line 211: Might want to notate that this is deprecated?

internal/contracts/manager.go
```diff
@@ -77,7 +77,7 @@ func NewContractManager(ctx context.Context, database database.Plugin, publicSto
 	return &contractManager{
 		database:          database,
 		txHelper:          txcommon.NewTransactionHelper(database),
-		publicStorage:     publicStorage,
+		sharedStorage:     sharedStorage,
```
awrichar, line 80: I'm surprised there are no renames outside of the constructor. Is this even used?

internal/data/data_manager.go
```diff
@@ -50,26 +50,26 @@ type dataManager struct {
...
 
-func NewDataManager(ctx context.Context, di database.Plugin, pi publicstorage.Plugin, dx dataexchange.Plugin) (Manager, error) {
+func NewDataManager(ctx context.Context, di database.Plugin, pi sharedstorage.Plugin, dx dataexchange.Plugin) (Manager, error) {
 	if di == nil || pi == nil || dx == nil {
 		return nil, i18n.NewError(ctx, i18n.MsgInitializationNilDepError)
 	}
 	dm := &dataManager{
 		database:          di,
-		publicstorage:     pi,
+		sharedstorage:     pi,
```
awrichar, line 65: Is this used? I see that blobstore (initialized below) uses its reference to sharedstorage, but I'm not sure this one actually gets used.

internal/orchestrator/orchestrator.go
```diff
@@ -359,14 +364,24 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
 		return err
 	}
 
-	if or.publicstorage == nil {
-		psType := config.GetString(config.PublicStorageType)
-		if or.publicstorage, err = psfactory.GetPlugin(ctx, psType); err != nil {
-			return err
+	if or.sharedstorage == nil {
+		ssType := config.GetString(config.SharedStorageType)
```
awrichar, line 368: You might have a look at the `initDataExchange` method, which does a similar backwards-compatible plugin initialization. In particular, I'd suggest simply checking if `ssType` is set or empty. If it's set, it needs to be valid. If it's empty, fallback to check the old config name. I don't think it's worth calling `GetPlugin` with an empty string and then looking for errors.

internal/orchestrator/orchestrator.go
```diff
@@ -359,14 +364,24 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
...
+				}
+			} else {
+				return err
+			}
 		}
 	}
-	if err = or.publicstorage.Init(ctx, publicstorageConfig.SubPrefix(or.publicstorage.Name()), or); err != nil {
-		return err
+	if err = or.sharedstorage.Init(ctx, sharedstorageConfig.SubPrefix(or.sharedstorage.Name()), or); err != nil {
+		if err = or.sharedstorage.Init(ctx, publicstorageConfig.SubPrefix(or.sharedstorage.Name()), or); err != nil {
```
awrichar, line 382: I think you should also track which config was used above and invoke the matching prefix, rather than simply trying both.

README.md
peterbroadhurst: Yeah - I need to go through and give this whole thing a bit of love at some point soon 

```
       ,
      /(  ___________
     |  >:===========`
      )(
      ""
```

peterbroadhurst:
Copying comment from https://github.com/hyperledger/firefly/pull/548#pullrequestreview-895823364

> Note I think we should:
> - Coordinate a merge of #549 with this, as anyone who up-migrates will be unable to apply that after (it contains the missing gaps)
> - Consider shipping a `v0.13.1` before we merge this (or #549 or #517) as I think all those would need 2nd digit bump

README.md
nguyer: It took me entirely too long to read what that meant 😆

awrichar (approved):

PR #548, opened 2022-02-24
Rename contract subscriptions to contract listeners

nguyer:
This PR includes several noteworthy changes:

- Contract Subscriptions have now been renamed to Contract Listeners everywhere, to reduce confusion. This includes the `/contracts/subscriptions` endpoint. It is now `/contracts/listeners`
- The `/subscribe` endpoints on Contract Interfaces and APIs have been dropped
- The `/contracts/listeners` endpoint can now take a interface reference to point it at an existing FFI, rather than putting the full FFI for the event inline
- `blockchain_event` has been renamed to `blockchain_event_received`
- The FFI generator endpoint for Ethereum now expects the ABI nested one level deeper. The payload should now look like:

```json
{
  "input": {
    "abi": []
  }
}
```


peterbroadhurst (commented):
Note I think we should:
- Coordinate a merge of #549 with this, as anyone who up-migrates will be unable to apply that after (it contains the missing gaps)
- Consider shipping a `v0.13.1` before we merge this (or #549 or #517) as I think all those would need 2nd digit bump

internal/contracts/manager.go
```diff
@@ -44,15 +44,15 @@ type Manager interface {
 	GetContractAPI(ctx context.Context, httpServerURL, ns, apiName string) (*fftypes.ContractAPI, error)
 	GetContractAPIs(ctx context.Context, httpServerURL, ns string, filter database.AndFilter) ([]*fftypes.ContractAPI, *database.FilterResult, error)
 	BroadcastContractAPI(ctx context.Context, httpServerURL, ns string, api *fftypes.ContractAPI, waitConfirm bool) (output *fftypes.ContractAPI, err error)
-	SubscribeContract(ctx context.Context, ns, eventPath string, req *fftypes.ContractSubscribeRequest) (*fftypes.ContractSubscription, error)
-	SubscribeContractAPI(ctx context.Context, ns, apiName, eventPath string, req *fftypes.ContractSubscribeRequest) (*fftypes.ContractSubscription, error)
+	// SubscribeContract(ctx context.Context, ns, eventPath string, req *fftypes.ContractSubscribeRequest) (*fftypes.ContractListener, error)
```
peterbroadhurst, line 47: Any reason to keep these commented out rather than just delete them (I assume they are unused)?

internal/contracts/manager.go
```diff
@@ -480,6 +480,14 @@ func (cm *contractManager) AddContractSubscription(ctx context.Context, ns strin
 			sub.Event = &fftypes.FFISerializedEvent{
 				FFIEventDefinition: event.FFIEventDefinition,
 			}
+		} else if sub.Event.Name != "" && sub.Interface != nil {
```
peterbroadhurst, line 483: What about other combinations, such as `sub.Event.Name == "" && sub.Interface != nil` - is that an error combination handled in `UpsertContractListener`?

internal/contracts/manager.go
```diff
@@ -526,58 +534,58 @@ func (cm *contractManager) GetContractSubscriptionByNameOrID(ctx context.Context
...
-	}
-
-	req.Interface = api.Interface.ID
-	if api.Location != nil {
-		req.Location = api.Location
-	}
-
-	return cm.SubscribeContract(ctx, ns, eventPath, req)
-}
+// func (cm *contractManager) SubscribeContract(ctx context.Context, ns, eventPath string, req *fftypes.ContractSubscribeRequest) (*fftypes.ContractListener, error) {
```
peterbroadhurst, line 554: Note earlier comment on just zapping these if they're un-used

internal/i18n/en_translations.go
```diff
@@ -229,7 +229,7 @@ var (
 	MsgContractInternalType         = ffm("FF10309", "Input '%v' of type '%v' is not compatible blockchain internalType of '%v'", 400)
 	MsgContractLocationInvalid      = ffm("FF10310", "Failed to validate contract location: %v", 400)
 	MsgContractParamInvalid         = ffm("FF10311", "Failed to validate contract param: %v", 400)
-	MsgContractSubscriptionExists   = ffm("FF10312", "A contract subscription already exists in the namespace: '%s' with name: '%s'", 409)
+	MsgContractListenerExists       = ffm("FF10312", "A contract subscription already exists in the namespace: '%s' with name: '%s'", 409)
```
peterbroadhurst, line 232: Think it's worth changing the message text as well as the name

pkg/fftypes/event.go
```diff
@@ -47,8 +47,8 @@ var (
 	EventTypeContractInterfaceConfirmed EventType = ffEnum("eventtype", "contract_interface_confirmed")
 	// EventTypeContractAPIConfirmed occurs when a new contract API has been confirmed
 	EventTypeContractAPIConfirmed EventType = ffEnum("eventtype", "contract_api_confirmed")
-	// EventTypeBlockchainEvent occurs when a new event has been recorded from the blockchain
-	EventTypeBlockchainEvent EventType = ffEnum("eventtype", "blockchain_event")
+	// EventTypeBlockchainEvent occurs when a new event has been received from the blockchain
+	EventTypeBlockchainEvent EventType = ffEnum("eventtype", "blockchain_event_received")
```
peterbroadhurst, line 51: Probably worth `EventTypeBlockchainEvent` -> `EventTypeBlockchainEventReceived` for consistency

peterbroadhurst (changes requested):
Changes look great. Some very minor suggestions, and a comment on coordinating the merge.

internal/contracts/manager.go
nguyer: I did a little more testing on this to try different combinations of these things, and tweaked this just a bit. The case you were asking about was already covered by the `validateFFIEvent` function. I added some additional tests as well. I think we're covered here now.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -134,6 +134,10 @@ type EthconnectMessageHeaders struct {
 	ID   string `json:"id,omitempty"`
 }
 
+type FFIGenerationInput struct {
```
peterbroadhurst, line 137: Is this change described in the PR description?

It has migration implications.

peterbroadhurst (approved):
@nguyer - thanks for the updates. One additional comment to make sure we document the `abi` nesting change.

internal/blockchain/ethereum/ethereum.go
nguyer: Thanks I will note that. This is a breaking change which should be called out in the release notes as well.

PR #514, opened 2022-02-11
Flatten the DXInfo back to the orginal payload on the ffdx API to connectors

peterbroadhurst:
... details in the comments of the change

PR #549, opened 2022-02-24
FIR-9: Identity enhancements

peterbroadhurst:
See https://github.com/hyperledger/firefly-fir/pull/9

@awrichar - the E2E is passing on these changes, but leaving PR in review state as I'd like to add some more E2E tests and do migration testing before pulling it out of draft.

I will put comments in-line in the code to help orient your review


Makefile
```diff
@@ -31,43 +31,43 @@ mocks-$(strip $(1))-$(strip $(2)): ${MOCKERY}
...
+$(eval $(call makemock, internal/batchpin,         Submitter,            batchpinmocks))
+$(eval $(call makemock, internal/sysmessaging,     SystemEvents,         sysmessagingmocks))
+$(eval $(call makemock, internal/sysmessaging,     MessageSender,        sysmessagingmocks))
+$(eval $(call makemock, internal/sysmessaging,     LocalNodeInfo,        sysmessagingmocks))
+$(eval $(call makemock, internal/syncasync,        Bridge,               syncasyncmocks))
+$(eval $(call makemock, internal/data,             Manager,              datamocks))
+$(eval $(call makemock, internal/batch,            Manager,              batchmocks))
+$(eval $(call makemock, internal/broadcast,        Manager,              broadcastmocks))
+$(eval $(call makemock, internal/privatemessaging, Manager,              privatemessagingmocks))
+$(eval $(call makemock, internal/contracts,        Manager,              contractmocks))
```
peterbroadhurst, line 61: Re-ordered to match the dependency order

db/migrations/postgres/000063_create_identities_table.up.sql
```diff
@@ -0,0 +1,137 @@
...
+  namespace      VARCHAR(64)     NOT NULL,
+  value          TEXT            NOT NULL,
+  created        BIGINT          NOT NULL
+);
+
+CREATE UNIQUE INDEX verifiers_id ON verifiers(id);
+CREATE UNIQUE INDEX verifiers_value ON verifiers(vtype, namespace, value);
+CREATE UNIQUE INDEX verifiers_identity ON verifiers(identity);
+
+INSERT INTO identities (
```
peterbroadhurst, line 38: This is obviously more complex than all our previous migrations, as we merge two tables into one.

I have managed to make it so we delete the original tables, and the down migration re-creates them - but only with data that exists prior to the up-migration.

db/migrations/sqlite/000065_pin_signer.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE pins ADD COLUMN signer TEXT;
+UPDATE pins SET signer = "";
```
peterbroadhurst, line 2: This was something significant I found as I was verifying the code against this table.

If the `pins` doesn't have the signing key reference in there, then for private messages we cannot verify the signing key against the `key` field in the message.

I did consider avoiding the full Eth address / Fabric DN string in the pins table - but a join table seemed like a lot of complexity for an unknown optimization benefit.

internal/apiserver/route_get_net_node.go
```diff
@@ -35,7 +35,7 @@ var getNetworkNode = &oapispec.Route{
 	FilterFactory:   nil,
 	Description:     i18n.MsgTBD,
 	JSONInputValue:  nil,
-	JSONOutputValue: func() interface{} { return &fftypes.Node{} },
+	JSONOutputValue: func() interface{} { return &fftypes.Identity{} },
```
peterbroadhurst, line 38: Note that `Identity` doesn't have exactly the same field set as the old `fftypes.Organization` or `fftypes.Node` objects. However, they are really close and it felt wrong to arbitrarily do mapping.

So we need to describe the difference in the JSON structures (`did` added, `identity` removed, `dx.endpoint` now just `profile`) in the migration notes for the next release.

internal/assets/token_approval.go
```diff
@@ -151,13 +151,6 @@ func (am *assetManager) validateApproval(ctx context.Context, ns string, approva
...
-	if approval.Key == "" {
-		org, err := am.identity.GetLocalOrganization(ctx)
-		if err != nil {
-			return err
-		}
-		approval.Key = org.Identity
-	}
-
-	return nil
+	approval.Key, err = am.identity.ResolveInputSigningKeyOnly(ctx, approval.Key)
```
peterbroadhurst, line 154: This change is made across all the token connectors, which were actually not previously doing the resolve function on the key.

... which makes me realize the keys are not guaranteed to be resolvable by the blockchain connector in this case. So I need to tweak this before we merge, so they use a function that only does the default-setting, and does **not** do the resolve against the blockchain plugin.

internal/broadcast/message.go
```diff
@@ -135,8 +134,8 @@ func (s *broadcastSender) resolveAndSend(ctx context.Context, method sendMethod)
 
 func (s *broadcastSender) resolve(ctx context.Context) ([]*fftypes.DataAndBlob, error) {
 	// Resolve the sending identity
-	if !s.isRootOrgBroadcast(ctx) {
```
peterbroadhurst, line 138: I was pleased to be able to remove this function, and simply push the detailed validation here down to the `IdentityClaim` specific confirmation code.

internal/database/sqlcommon/identity_sql.go
```diff
@@ -0,0 +1,250 @@
...
+		"itype",
+		"namespace",
+		"name",
+		"description",
+		"profile",
+		"messages_claim",
+		"messages_verification",
+		"messages_update",
+		"created",
+		"updated",
```
peterbroadhurst, line 44: Note the updated is due to the fact that there's a `PATCH` on these, with the new `IdentityUpdate` flow.

internal/database/sqlcommon/identity_sql.go
```diff
@@ -0,0 +1,250 @@
...
+				identity.Created,
+				identity.Updated,
+			),
+		func() {
+			s.callbacks.UUIDCollectionNSEvent(database.CollectionIdentities, fftypes.ChangeEventTypeCreated, identity.Namespace, identity.ID)
+		})
+	return err
+}
+
+func (s *SQLCommon) UpsertIdentity(ctx context.Context, identity *fftypes.Identity, optimization database.UpsertOptimization) (err error) {
```
peterbroadhurst, line 105: I went straight to the optimization enabled upsert, as it seems like a good pattern.
I hope we'll roll this out universally across the code over time.

internal/database/sqlcommon/message_sql_test.go
```diff
@@ -49,7 +49,7 @@ func TestUpsertE2EWithDB(t *testing.T) {
 			ID:   msgID,
 			CID:  nil,
 			Type: fftypes.MessageTypeBroadcast,
-			Identity: fftypes.Identity{
+			SignerRef: fftypes.SignerRef{
```
peterbroadhurst, line 52: Note this is just an internal structure name change for grouping of fields - the JSON field names are still the same (`key` - the signing key, and `author` - the DID)

internal/database/sqlcommon/nextpin_sql.go
```diff
@@ -155,7 +155,7 @@ func (s *SQLCommon) UpdateNextPin(ctx context.Context, sequence int64, update da
 	}
 	defer s.rollbackTx(ctx, tx, autoCommit)
 
-	query, err := s.buildUpdate(sq.Update("nextpins"), update, nodeFilterFieldMap)
+	query, err := s.buildUpdate(sq.Update("nextpins"), update, pinFilterFieldMap)
```
peterbroadhurst, line 158: Existing bug fixed here

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -45,7 +45,7 @@ type FFDX struct {
 	needsInit    bool
 	initialized  bool
 	initMutex    sync.Mutex
-	nodes        []fftypes.DXInfo
+	nodes        []fftypes.JSONObject
```
peterbroadhurst, line 48: After the discussion in recent change https://github.com/hyperledger/firefly/pull/514 I found I could just standardize the whole code on expecting `id` to exist in this JSON structure, and pull that out. That meant I could remove the complexity of the `DXInfo` object.

However, the peer ID is still a first class object in the new model. Instead of being in the `Identity` object - it's a new type of `Verifier` that's associated with it.

This means it's an indexed field that we can search by, but we don't have to have special stuff in the identity object. It's also true that we're using it as a verifier for messages - we trust private exchanges, as long as they come from a `peer` that is registered to the org claimed in the message (see table referenced above).

You'll notice there is some special handling of how we broadcast `nodes` as an identity - because the broadcast needs to be self-verifying at the blockchain layer. e.g. the `claim` is signed by the `parent` blockchain verifier of a node (rather than the `claim` being signed by the child, and the `verification` being signed by the parent).

internal/definitions/definition_handler_identity_claim.go
```diff
@@ -0,0 +1,211 @@
...
+		verifier.VerifierRef.Type = fftypes.VerifierTypeFFDXPeerID
+		verifier.VerifierRef.Value = identity.Profile.GetString("id")
+	default:
+		verifier.VerifierRef.Type = dh.blockchain.VerifierType()
+		verifier.VerifierRef.Value = msg.Header.Key
+	}
+	return verifier
+}
+
+func (dh *definitionHandlers) confirmVerificationForClaim(ctx context.Context, state DefinitionBatchState, msg *fftypes.Message, identity, parent *fftypes.Identity) (*fftypes.UUID, error) {
```
peterbroadhurst, line 80: Note this function has to work in two cases, because we cannot guarantee the batch-pin order of two transactions signed by different blockchain identities (they will be batched separately, and ordered into blocks separately):

1. The claim comes in first before the verification
    - In this case, the claim message is `confirmed` and this function will be run again when the `verification` comes in
2. The verification comes in first
    - The verification is `confirmed` when it arrives, so that we can search for it as confirmed here when we (later) process the verification

Note we handle the case where the claim and the verification end up in the same definition-batch, which required an update to the `DefinitionBatchState` interface (more later on that).

internal/definitions/definition_handler_identity_claim.go
```diff
@@ -0,0 +1,211 @@
...
+func (dh *definitionHandlers) verifyClaimSignature(ctx context.Context, msg *fftypes.Message, identity *fftypes.Identity, parent *fftypes.Identity) (valid bool) {
+
+	author := msg.Header.Author
+	if author == "" {
+		return false
+	}
+
+	var expectedSigner *fftypes.Identity
+	switch {
+	case identity.Type == fftypes.IdentityTypeNode:
```
peterbroadhurst, line 48: Special case I mentioned above where the parent org signs the node claim directly (no separate verification flow)

internal/definitions/definition_handler_identity_update.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return ActionRetry, err
+	}
+	if identity == nil {
+		log.L(ctx).Warnf("Invalid identity update message %s - not found: %s", msg.Header.ID, update.Identity.ID)
+		return ActionReject, nil
+	}
+
+	// Check the author matches
+	if identity.DID != msg.Header.Author {
```
peterbroadhurst, line 52: Updates are simply-self signed by the identity that already is established. The parent does not need to sign-off on profile updates.

internal/definitions/definition_handler_identity_verification.go
```diff
@@ -0,0 +1,85 @@
...
+	// At this point, this is a valid verification, but we don't know if the claim has arrived.
+	// It might be being processed in the same pin batch as us - so we can't
+
+	// See if the message has already arrived, if so we need to queue a rewind to it
+	claimMsg, err := dh.database.GetMessageByID(ctx, verification.Claim.ID)
+	if err != nil {
+		return ActionRetry, err
+	}
+	if claimMsg == nil || claimMsg.State != fftypes.MessageStateConfirmed {
+		claimMsg = state.GetPendingConfirm()[*verification.Claim.ID]
```
peterbroadhurst, line 63: This is the update to the definition batch state, so we know if a message has been confirmed in the same batch as us (as we won't find it from the database search - it only gets marked confirmed after the finalizers have run).

internal/events/aggregator.go
```diff
@@ -88,14 +94,14 @@ func newAggregator(ctx context.Context, di database.Plugin, sh definitions.Defin
 }
 
 func (ag *aggregator) start() {
-	go ag.offchainListener()
+	go ag.batchRewindListener()
```
peterbroadhurst, line 97: This rename is not strictly required. I started down a path of looking at using the rewind functionality, to leave messages pending between claim and verification. However, as that would block any other claims to the identity on the same topic, and have other problems, I abandoned the approach.

I left the rename, as I think the new more generic naming is better.

internal/events/aggregator.go
```diff
@@ -249,6 +255,46 @@ func (ag *aggregator) processPins(ctx context.Context, pins []*fftypes.Pin, stat
 	return err
 }
 
+func (ag *aggregator) checkOnchainConsistency(ctx context.Context, msg *fftypes.Message, pin *fftypes.Pin) (valid bool, err error) {
```
peterbroadhurst, line 258: The introduction of this new function, and its use in a different phase of the verification, is what I mentioned above about introducing the `signer` on the `pin`.

Previously we verified the signing identity for batches when storing them to the DB, after querying the data from IPFS - because we had the signing identity in-hand at that point. For private messages, no such code path existed.

Now we defer the check to each individual message (which retains the signer information individually) at the time we are attempting the dispatch of the pin in the aggregator. This has multiple benefits:

- We do it consistently for private and broadcast
- We do the identity lookup at the point in the sequence that the pin is being processed consistently - so that means checks on identity etc. should be validly ordered.

Note that the batch persist logic still ensures that every message in the batch, has the same signing identity.

internal/events/aggregator.go
```diff
@@ -249,6 +255,46 @@ func (ag *aggregator) processPins(ctx context.Context, pins []*fftypes.Pin, stat
...
+		return false, err
+	}
+	if resolvedAuthor == nil {
+		if msg.Header.Type == fftypes.MessageTypeDefinition &&
+			(msg.Header.Tag == fftypes.SystemTagIdentityClaim || msg.Header.Tag == fftypes.DeprecatedSystemTagDefineNode || msg.Header.Tag == fftypes.DeprecatedSystemTagDefineOrganization) {
+			// We defer detailed checking of this identity to the system handler
+			return true, nil
+		} else if msg.Header.Type != fftypes.MessageTypePrivate {
+			// Only private messages, or root org broadcasts can have an unregistered key
+			l.Errorf("Invalid message '%s'. Author '%s' cound not be resolved: %s", msg.Header.ID, msg.Header.Author, err)
```
peterbroadhurst, line 286: This is the special case that allows an anonymous key to be used to pin private messages (per the table).

Note that the key must **not** be registered to any identity to use this path. If the key is registered, then it must be registered to the `author` identity in the private message.

internal/events/dx_callbacks.go
```diff
@@ -64,51 +64,49 @@ func (em *eventManager) MessageReceived(dx dataexchange.Plugin, peerID string, d
 	return string(manifestBytes), err
 }
 
-func (em *eventManager) checkReceivedIdentity(ctx context.Context, peerID, author, signingKey string) (node *fftypes.Node, err error) {
+// the data came from) matches the org listed in the batch. The on-chain identity check
+// is performed by the aggregator, across broadcast and private consistently.
+func (em *eventManager) checkReceivedOffchainIdentity(ctx context.Context, peerID, author string) (node *fftypes.Identity, err error) {
```
peterbroadhurst, line 69: Renamed to make it clear that this only verifies the off-chain identity at this point.

internal/events/persist_batch.go
```diff
@@ -18,75 +18,20 @@ package events
...
-		l.Errorf("Invalid batch '%s'. Hash in batch '%s' does not match transaction hash '%s'", batch.ID, batch.Hash, onchainHash)
+		log.L(ctx).Errorf("Invalid batch '%s'. Hash in batch '%s' does not match transaction hash '%s'", batch.ID, batch.Hash, onchainHash)
 		return false, nil // This is not retryable. skip this batch
 	}
 
-	valid, err = em.persistBatch(ctx, batch)
-	return valid, err
-}
-
-func (em *eventManager) isRootOrgBroadcast(batch *fftypes.Batch) bool {
```
peterbroadhurst, line 64: The second copy of this logic I was able to remove :)

internal/events/token_pool_created.go
```diff
@@ -165,7 +165,7 @@ func (em *eventManager) TokenPoolCreated(ti tokens.Plugin, pool *tokens.TokenPoo
 		// Initiate a rewind if a batch was potentially completed by the arrival of this transaction
 		if batchID != nil {
 			log.L(em.ctx).Infof("Batch '%s' contains reference to received pool '%s'", batchID, pool.ProtocolID)
-			em.aggregator.offchainBatches <- batchID
+			em.aggregator.rewindBatches <- batchID
```
peterbroadhurst, line 168: This is an example of why I left the rename - as the `offchain` felt a bit odd here.

internal/i18n/errors.go
```diff
@@ -29,5 +29,8 @@ func NewError(ctx context.Context, msg MessageKey, inserts ...interface{}) error
 
 // WrapError wraps an error
 func WrapError(ctx context.Context, err error, msg MessageKey, inserts ...interface{}) error {
+	if err == nil {
```
peterbroadhurst, line 32: This kept catching me out in the code - where we'd return a `nil` error, if we accidentally used `WrapError` instead of `NewError`. So I've made it so you can safely use `WrapError` for a case where `err` is nil - and it will just fall back to `NewError`.

internal/identity/identitymanager.go
```diff
@@ -33,36 +34,44 @@ import (
...
 
 type Manager interface {
-	ResolveInputIdentity(ctx context.Context, identity *fftypes.Identity) (err error)
-	ResolveSigningKey(ctx context.Context, inputKey string) (outputKey string, err error)
-	ResolveSigningKeyIdentity(ctx context.Context, signingKey string) (author string, err error)
-	ResolveLocalOrgDID(ctx context.Context) (localOrgDID string, err error)
-	GetLocalOrgKey(ctx context.Context) (string, error)
-	OrgDID(org *fftypes.Organization) string
-	GetLocalOrganization(ctx context.Context) (*fftypes.Organization, error)
+	ResolveInputSigningIdentity(ctx context.Context, namespace string, msgSignerRef *fftypes.SignerRef) (err error)
```
peterbroadhurst, line 37: Pretty much all newly rebuilt (as well as renamed) functions in this class, trying to address some previous sources of confusion, as well as introducing a better level of caching, and migrating to the new `Identity` object model.

internal/definitions/definition_handler_network_node_test.go
```diff
@@ -24,418 +24,170 @@ import (
...
-		DX: fftypes.DXInfo{
-			Peer:     "peer1",
-			Endpoint: fftypes.JSONObject{"some": "info"},
-		},
-	}
-	b, err := json.Marshal(&node)
-	assert.NoError(t, err)
-	data := &fftypes.Data{
-		Value: fftypes.JSONAnyPtrBytes(b),
+const oldNodeExample = `
```
peterbroadhurst, line 34: I took sample payloads from a current FireFly (v0.13) environment to build these tests. Hopefully that will help protect against future issues that could stop replay of existing organizations/nodes for a member joining a network.

Note that there is **no support for migration of organizations with a parent org**.

My justification for this is:
- We do not believe the function is adopted
- The child identities do not prove ownership of their keys - which is required in the new model - so any migration would have to include some kind of sophisticated "fork" support to prevent new identities exploiting the old parent-key-only verification model.

internal/networkmap/update_identity.go
```diff
@@ -0,0 +1,68 @@
...
+package networkmap
+
+import (
+	"context"
+
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+func (nm *networkMap) UpdateIdentity(ctx context.Context, ns, uuidStr string, dto *fftypes.IdentityUpdateDTO, waitConfirm bool) (identity *fftypes.Identity, err error) {
```
peterbroadhurst, line 26: The new code takes the approach that an identity update is a completely separate flow to an identity claim/verification. This seemed to make sense from lots of angles, particularly the fact that an identity update only needs to be signed by the owner of the identity.

internal/privatemessaging/groupmanager.go
```diff
@@ -204,7 +204,7 @@ func (gm *groupManager) getGroupNodes(ctx context.Context, groupHash *fftypes.By
 //
 // Errors are only returned for database issues. For validation issues, a nil group is returned without an error.
 func (gm *groupManager) ResolveInitGroup(ctx context.Context, msg *fftypes.Message) (*fftypes.Group, error) {
-	if msg.Header.Tag == string(fftypes.SystemTagDefineGroup) {
+	if msg.Header.Tag == fftypes.SystemTagDefineGroup {
```
peterbroadhurst, line 207: I got rid of the special enum type for the system tags - didn't seem to add anything, apart from a bunch of this kind of casting all over the place. They are just string constants now.

internal/privatemessaging/privatemessaging.go
```diff
@@ -245,7 +245,7 @@ func (pm *privateMessaging) sendData(ctx context.Context, tw *fftypes.TransportW
 		}
 
 		// Send the payload itself
-		err := pm.exchange.SendMessage(ctx, op.ID, node.DX.Peer, payload)
+		err := pm.exchange.SendMessage(ctx, op.ID, node.Profile.GetString("id"), payload)
```
peterbroadhurst, line 248: This is the concrete example of what I mentioned above - where it is necessary (and enforced) for all DX profiles to have the special field `id` in there. Note this code could have looked up the `verifier` chained off the node (used for the reverse lookup when verifying the off-chain identity of something received), but it seemed easier to just access it directly here rather than taking an extra cached-DB-lookup complexity onboard.

pkg/blockchain/plugin.go
```diff
@@ -85,7 +88,7 @@ type Callbacks interface {
 	// submitted by us, or by any other authorized party in the network.
 	//
 	// Error should will only be returned in shutdown scenarios
-	BatchPinComplete(batch *BatchPin, signingIdentity string) error
+	BatchPinComplete(batch *BatchPin, signingKey *fftypes.VerifierRef) error
```
peterbroadhurst: A bit of faff throughout this PR in dealing with the fact that we don't just have a string for a verifier any more, you have to describe its type. This is necessary to allow a single DB table, but as we have separate types for Fabric/Ethereum we have to pass it around a bit.

pkg/fftypes/batch.go
```diff
@@ -30,7 +30,7 @@ type Batch struct {
 	Namespace string      `json:"namespace"`
 	Type      MessageType `json:"type"`
 	Node      *UUID       `json:"node,omitempty"`
-	Identity
+	SignerRef
```
peterbroadhurst: Per previous comments, this code level spelling change is not visible on the external JSON.

pkg/fftypes/constants.go
```diff
@@ -22,31 +22,38 @@ const (
 	SystemNamespace = "ff_system"
 )
 
-type SystemTag string
```
peterbroadhurst: As mentioned above, this was removed as it wasn't adding value.

pkg/fftypes/jsonany.go
```diff
@@ -135,11 +136,18 @@ func (h *JSONAny) JSONObjectNowarn() JSONObject {
 	return jo
 }
 
+// Value ensures we write null to the DB for null values
+func (h *JSONAny) Value() (driver.Value, error) {
+	if h.IsNil() {
+		return nil, nil
```
peterbroadhurst: Previously when we stored `null` data to the DB, we were storing the string `"null"`. That's a pain as the query REST filtering doesn't let you check for that using our standard null-checking.

I hit confusion in the code too about storing `nil` and getting back `"null"` for other paths.

I did a good hunt around for any implications of this change on other stored objects, and didn't find any. We were using `JSONObject` in a lot more cases, which already stores `null` as `nil` (not `"null"`)

pkg/fftypes/verifier.go
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fftypes
+
+// VerifierType is the type of an identity berifier. Where possible we use established DID verifier type strings
+type VerifierType = FFEnum
+
+var (
```
peterbroadhurst: I realize I still need to add the `GET` `/api/v1/namespaces/{ns}/identities/{idid}/did` export for a DID JSON object that is mentioned in the FIR. This code here was written with that in mind, but I used snake-case consistent lower-case values here as it's important in how we treat FFEnum values (these can be translated into the DID ones that are camel-case in the JSON mapping layer).

internal/assets/token_approval.go
peterbroadhurst: ok - I've addressed this.
Rather than switching address resolution off completely (which would be a big pain for the 90% case of Eth tokens + Eth blockchain), I've added a setting (default true) that allows it to be switched off for asset manager scenarios.

This might need to evolve in the future, if a FireFly use case requires advanced Ethereum resolution of keys (with things like HD wallet alias support etc.), in a mixed blockchain (Fabric core, Eth tokens) environment. I suggest the community addresses that issue if/when encountered.

db/migrations/postgres/000065_pin_signer_event_correlator.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+ALTER TABLE pins ADD COLUMN signer TEXT;
+UPDATE pins SET signer = "";
+
+ALTER TABLE events ADD COLUMN cid UUID;
```
peterbroadhurst, line 5: See the sync-async changes below

internal/syncasync/sync_async_bridge.go
```diff
@@ -44,6 +44,8 @@ type Bridge interface {
 	WaitForReply(ctx context.Context, ns string, id *fftypes.UUID, send RequestSender) (*fftypes.MessageInOut, error)
 	// WaitForMessage waits for a message with the supplied ID
 	WaitForMessage(ctx context.Context, ns string, id *fftypes.UUID, send RequestSender) (*fftypes.Message, error)
+	// WaitForIdentity waits for an identity with the supplied ID
+	WaitForIdentity(ctx context.Context, ns string, id *fftypes.UUID, send RequestSender) (*fftypes.Identity, error)
```
peterbroadhurst, line 48: Realized I needed to add full sync-async support for `IdentityClaim` because either message could be confirmed first.

internal/syncasync/sync_async_bridge.go
```diff
@@ -225,18 +242,23 @@ func (sa *syncAsyncBridge) getOperationFromEvent(event *fftypes.EventDelivery) (
 }
 
 func (sa *syncAsyncBridge) handleMessageConfirmedEvent(event *fftypes.EventDelivery) error {
+
+	// See if the CID marks this as a reply to an inflight message
+	inflight := sa.getInFlight(event.Namespace, messageConfirm, event.Reference)
+	inflightReply := sa.getInFlight(event.Namespace, messageReply, event.Correlator)
+
+	if inflightReply == nil && inflight == nil {
```
peterbroadhurst, line 250: I found as I was extending, that the sync async manager was needing to perform a DB lookup for any event emitted in the system - regardless of if anything was in-flight. This didn't feel right, and as I was working on the code here I thought I'd tweak it to check the in-flight list first.

This did mean adding a `correlator` field to the event, alongside the `reference`, that contains the `cid` from a reply message, or for definitions it's filled in by the definition handler appropriately.

internal/apiserver/route_get_net_node.go
awrichar: I agree no need to do extra data mapping, but this will be a breaking change for some apps built against the current version of FireFly.

Are we certain we want to keep the old `/nodes` and `/organizations` routes at all, or are the redundant with the new `/identities` routes?

db/migrations/postgres/000063_create_identities_table.up.sql
```diff
@@ -0,0 +1,137 @@
...
+);
+
+CREATE UNIQUE INDEX identities_id ON identities(id);
+CREATE UNIQUE INDEX identities_did ON identities(did);
+CREATE UNIQUE INDEX identities_name ON identities(itype, namespace, name);
+
+CREATE TABLE verifiers (
+  seq            SERIAL          PRIMARY KEY,
+  id             UUID            NOT NULL,
+  identity       UUID            NOT NULL,
```
awrichar, line 27: What is the difference between "id" and "identity" here? My first guess was that "id" is a generated ID unique to the verifier row, and that "identity" is a foreign key reference to the identities table. But I saw in the migrations logic below, these columns are set to the same value.

peterbroadhurst: `identity` is a link to the `Identity` resource that owns this.
`id` is the UUID of this `Verification` object.

internal/apiserver/route_get_identity_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/hyperledger/firefly/internal/config"
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var getIdentityByID = &oapispec.Route{
+	Name:   "getIdentityByID",
+	Path:   "namespaces/{ns}/identities/{idid}",
```
awrichar, line 30: minor thing, but `idid` does look a bit silly 😅 and also has DID in the name despite not being a DID endpoint. Would just `id` or `iid` be any better?

internal/config/config.go
```diff
@@ -224,6 +224,8 @@ var (
 	AssetManagerRetryMaxDelay = rootKey("asset.manager.retry.maxDelay")
 	// AssetManagerRetryFactor the backoff factor to use for retry of database operations
 	AssetManagerRetryFactor = rootKey("asset.manager.retry.factor")
+	// AssetManagerResolveKeysViaBlockchainPlugin whether to verify/resolve keys via the blockchain plugin (supporing facilities like key aliases), but requiring the same backing blockchain connector
+	AssetManagerResolveKeysViaBlockchainPlugin = rootKey("asset.manager.resolveKeysViaBlockchainPlugin")
```
awrichar, line 228: If this is false, does that mean keys are not resolved at all? Or there is some "other" way to resolve keys?

I'm probably just trying to turn a `bool` into an `enum` again :) just trying to be sure the implications of true vs false are clear here.

internal/broadcast/definition.go
```diff
@@ -25,28 +25,31 @@ import (
...
 	}
 
 	return bm.broadcastDefinitionCommon(ctx, ns, def, signingIdentity, tag, waitConfirm)
 }
 
-func (bm *broadcastManager) BroadcastRootOrgDefinition(ctx context.Context, def *fftypes.Organization, signingIdentity *fftypes.Identity, tag fftypes.SystemTag, waitConfirm bool) (msg *fftypes.Message, err error) {
+func (bm *broadcastManager) BroadcastDefinitionResolveKeyOnly(ctx context.Context, ns string, def fftypes.Definition, signingIdentity *fftypes.SignerRef, tag string, waitConfirm bool) (msg *fftypes.Message, err error) {
 
-	signingIdentity.Author = bm.identity.OrgDID(def)
+	signingIdentity.Key, err = bm.identity.ResolveInputSigningKeyOnly(ctx, signingIdentity.Key, true)
```
awrichar, line 44: Should we clear `signingIdentity.Author` here? Or perhaps we should just pass in a key, instead of passing in a `SignerRef` that is implied to have a key with no author.

internal/identity/identitymanager.go
```diff
@@ -33,36 +34,44 @@ import (
...
+	ResolveInputSigningKeyOnly(ctx context.Context, namespace string, resolveViaBlockchainPlugin bool) (signingKey string, err error)
```
awrichar, line 38: I think the var names here are wrong - second param is the key, not the namespace, right?

I would also say we can drop the "Only" and call it `ResolveInputSigningKey`, but maybe that's just me.

internal/config/config.go
awrichar: I see you explained [here](https://github.com/hyperledger/firefly/pull/549#discussion_r814465236), so sounds like:
* `resolveKeysViaBlockchainPlugin=true` means keys passed to asset/token routes will be validated via the blockchain plugin
* `resolveKeysViaBlockchainPlugin=false` means keys passed to asset/token routes will not be validated at all (which was actually the preexisting behavior)

I guess just wondering how this might be spelled if it's more extensible down the road...

internal/database/sqlcommon/nextpin_sql.go
awrichar: 😱

db/migrations/postgres/000063_create_identities_table.up.sql
awrichar: OK, that's what I'd guessed after a few minutes staring. Why are they set to the same value on the migration path, instead of generating a unique `id`? Is that purely to support the down-migration strategy of separating migrated from newly-created entries?

peterbroadhurst:
I've completed some extra migration testing now, using the following process:
- Set `DATABASE_TYPE=postgres`
- Checkout `main` and do `make e2e`
- Do a `docker compose down` on the env
- Checkout `identityv2` and do a `make docker`
- Remove the following volumes to reset the DB and event streams on one of the nodes:
   - `docker volume rm firefly_e2e_ethconnect_events_1 firefly_e2e_postgres_1`
- Do a `docker compose up -d`
- Check the `firefly_core_0` container comes up ok, and `organization` `org_0` seems ok
- Patch the admin endpoints to force a start of the `firefly_core_1` node:
    - `docker compose exec firefly_core_1 sh`
    - `apk add curl jq`
    - `curl -H 'Content-Type: application/json' -d 'true' -v -X PUT 'http://localhost:5201/admin/api/v1/config/records/admin.preInit'`
    - `curl -H 'Content-Type: application/json' -d 'true' -v -X POST  'http://localhost:5201/admin/api/v1/config/reset'`
 - Check the node starts ok
 - Check `organization` `org_0` seems ok, and matches the one on `firefly_core_0`
    - The only variation found was the `verifier` ID - which is independently generated anyway

db/migrations/postgres/000063_create_identities_table.up.sql
peterbroadhurst: 1. to support the down migration strategy
2. to avoid needing to have functions in PSQL/SQLite to generate UUIDs

peterbroadhurst: I realized actually after doing the migration work, that the `id` of a verifier is really a `localId` (not an `id`) as it won't be applied universally to all nodes. I'm trying to think in the context of a DID document if this is a problem.... it might be.

awrichar: Yea, if we can deterministically compute something like a `hash`, that seems better. Generally we only use `id` when the ID can be explicitly shared in a definition message.

internal/definitions/definition_handler.go
```diff
@@ -58,33 +60,43 @@ const (
 
 // DefinitionBatchActions are actions to be taken at the end of a definition batch
```
awrichar, line 61: Comment doesn't match the rename below

internal/events/aggregator.go
```diff
@@ -387,10 +439,15 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *fftypes.M
...
+		event.Correlator = msg.Header.CID
+		if customCorrelator != nil {
+			// Definition handlers can set a custom event correlator (such as a token pool ID)
+			event.Correlator = customCorrelator
+		}
 		if err = ag.database.InsertEvent(ctx, event); err != nil {
 			return err
 		}
-		log.L(ctx).Infof("Emitting %s %s for message %s:%s", eventType, event.ID, msg.Header.Namespace, msg.Header.ID)
+		log.L(ctx).Infof("Emitting %s %s for message %s:%s (c=%v)", eventType, event.ID, msg.Header.Namespace, msg.Header.ID, msg.Header.CID)
```
awrichar, line 450: probably should spell out `correlator=` or at least `cid=`, because I'll definitely forget what this log means

awrichar: and should we actually log `event.Correlator` here, which may or may not be the same as `msg.Header.CID`?

internal/definitions/definition_handler_identity_claim.go
```diff
@@ -0,0 +1,218 @@
...
+		// In the special case of a node, the parent signs it directly
+		expectedSigner = parent
+	default:
+		expectedSigner = identity
+	}
+
+	valid = author == expectedSigner.DID ||
+		(expectedSigner.Type == fftypes.IdentityTypeOrg || author == fmt.Sprintf("%s%s", fftypes.FireFlyOrgDIDPrefix, expectedSigner.ID))
```
awrichar, line 56: Is this logic correct? Feels like maybe one `||` should be a `&&`

internal/config/config.go
peterbroadhurst: I shall have a think and make a proposal. Overall I'd like to find a better word than `resolve` for the thing we do here, but one hasn't sprung to mind yet.

internal/broadcast/definition.go
peterbroadhurst: The intention of this function, is that it doesn't mess with the `Author` - because that must in the case of an `IdentityClaim` match the identity being created. So it's the caller's responsibility to have that set before calling in here. However, this function still gets the identity plugin to "resolve" the key (see comment above about spelling) - such as getting Fabric to look up the full MSP Key Identifier for a shortname.

Think I'll just add comments if that's ok, but please feel free to come back around if that doesn't help.

internal/identity/identitymanager.go
peterbroadhurst: Yeah, really fought with the function name spelling here. The key (no pun intended) thing about the name here is to state it's the unusual case. Most code paths when resolving a signing key, you want to resolve it in FireFly with an associated identity. In a couple of cases it want's to perform the special case of just "resolving" a key via the blockchain plugin, without resolving what identity might/might-not be associated with it.

internal/events/aggregator.go
```diff
@@ -342,21 +393,20 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *fftypes.M
 
 	// Validate the message data
 	valid := true
+	var customCorrelator *fftypes.UUID
 	switch {
 	case msg.Header.Type == fftypes.MessageTypeDefinition:
 		// We handle definition events in-line on the aggregator, as it would be confusing for apps to be
 		// dispatched subsequent events before we have processed the definition events they depend on.
-		msgAction, batchAction, err := ag.definitions.HandleDefinitionBroadcast(ctx, msg, data, tx)
+		state.correlator = nil
```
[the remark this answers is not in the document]
peterbroadhurst: 😄  - has just removed the multiple returns, then this one edge case came back.

If I shift it to a return, I'd like to make it an extensible structure again like before - because this one item doesn't feel significant enough to justify its own parameter in an ever increasing list.

internal/config/config.go
peterbroadhurst: ... considering `Normalize` unless something better springs to mind. Then it can be `keyNormalization: "blockchain"|"none"|...tbd`

db/migrations/postgres/000063_create_identities_table.up.sql
peterbroadhurst: I'm going to have to do a post migration during startup for this though sadly :(

peterbroadhurst: actually I might cheat and put the identity ID in there for migrations, and we can document that in the release notes

pkg/fftypes/identity.go
```diff
@@ -16,9 +16,226 @@
...
+
+// SignerRef is the nested structure representing the identity that signed a message.
+// It might comprise a resolvable by FireFly identity DID, a blockchain signing key, or both.
+type SignerRef struct {
 	Author string `json:"author,omitempty"`
 	Key    string `json:"key,omitempty"`
 }
+
+// IdentityClaim is the data payload used in a message to broadcast an intent to publish a new identity.
+// Most claims (except root orgs, where different requirements apply) require a separate IdentityValidation
```
awrichar, line 104: Is this comment a typo (IdentityValidation vs IdentityVerification)?

pkg/fftypes/identity.go
```diff
@@ -16,9 +16,226 @@
...
+}
+
+// IdentityVerification is the data payload used in message to broadcast a verification of a child identity.
+// Must refer to the UUID and Hash of the IdentityClaim message, and must contain the same base identity data.
+type IdentityVerification struct {
+	Claim    MessageRef   `json:"claim"`
+	Identity IdentityBase `json:"identity"`
+
+	// SignerRef lets us store back the message when broadcasting, but isn't part of the payload
+	SignerRef *Identity `json:"-"`
```
awrichar, line 118: Slightly confusing that this field is called "SignerRef" but is not of type "SignerRef"...

awrichar: Should it be IdentityRef as it is in the next struct below?

internal/definitions/definition_handler_identity_claim.go
```diff
@@ -0,0 +1,218 @@
...
+	existingVerifier, err := dh.database.GetVerifierByValue(ctx, verifier.Type, identity.Namespace, verifier.Value)
+	if err != nil {
+		return ActionRetry, err // retry database errors
+	}
+	if existingVerifier != nil && !existingVerifier.Identity.Equals(identity.ID) {
+		log.L(ctx).Warnf("Unable to process identity claim %s - verifier type=%s value=%s already registered: %v", msg.Header.ID, verifier.Type, verifier.Value, existingVerifier.ID)
+		return ActionReject, nil
+	}
+
+	if parent != nil && identity.Type != fftypes.IdentityTypeNode {
```
awrichar, line 174: Struggling to follow this block - it only executes for non-root, non-node identities? Could be worth an extra comment...

internal/events/operation_update.go
```diff
@@ -54,6 +56,13 @@ func (em *eventManager) operationUpdateCtx(ctx context.Context, operationID *fft
 	// Special handling for OpTypeTokenApproval, which writes an event when it fails
 	if op.Type == fftypes.OpTypeTokenApproval && txState == fftypes.OpStatusFailed {
 		event := fftypes.NewEvent(fftypes.EventTypeApprovalOpFailed, op.Namespace, op.ID, op.Transaction)
+		var tokenAppoval fftypes.TokenApproval
+		err = txcommon.RetrieveTokenApprovalInputs(ctx, op, &tokenAppoval)
+		if err != nil {
+			log.L(em.ctx).Warnf("Could not determine token retrieval: %s", err)
```
awrichar, line 62: Typo: token approval?

internal/i18n/errors.go
awrichar: I'm on the fence with this one... it changes the function from "if there was an error, wrap it and return it" to "there was definitely an error, so either wrap this one or create a new one". But I suppose either is fine, as long as all usages expect the latter/new behavior...

internal/config/config.go
awrichar: Works for me

internal/definitions/definition_handler_identity_claim.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	for _, candidate := range candidates {
+		data, foundAll, err := dh.data.GetMessageData(ctx, candidate, true)
+		if err != nil {
+			return nil, err
+		}
+		identityMatches := false
+		var verificationID *fftypes.UUID
+		var verificationHash *fftypes.Bytes32
```
awrichar, line 111: minor note, but it's not necessary to pre-declare these 3 vars, is it?

pkg/fftypes/identity.go
```diff
@@ -16,9 +16,226 @@
...
+// IdentityUpdate is the data payload used in message to broadcast an update to an identity profile.
+// The broadcast must be on the same identity as the currently established identity claim message for the identity,
+// and it must contain the same identity data.
+// The profile is replaced in its entirety.
+type IdentityUpdate struct {
+	Identity IdentityBase    `json:"identity"`
+	Updates  IdentityProfile `json:"updates,omitempty"`
+
+	// IdentityRef lets us store back the message when broadcasting, but isn't part of the payload
+	IdentityRef *Identity `json:"-"`
```
awrichar, line 130: Slightly confused about how this is used "when broadcasting" but isn't part of the broadcast payload. I may be missing something subtle.

internal/networkmap/did.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"context"
+
+	"github.com/hyperledger/firefly/internal/log"
+	"github.com/hyperledger/firefly/pkg/database"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+type DIDDocument struct {
```
awrichar, line 27: A link to the reference spec might be useful here

awrichar: possibly here? https://www.w3.org/TR/did-core/#core-properties

internal/definitions/definition_handler_identity_claim.go
awrichar: After reading more of the code, it does confirm my understanding that root identities and nodes have self-signed claims and do not require a separately signed verification message. But they still require "verifiers"? I think I'm very close to understanding, but not quite there.

internal/syncasync/sync_async_bridge.go
```diff
@@ -322,16 +381,19 @@ func (sa *syncAsyncBridge) handleTransferOpFailedEvent(event *fftypes.EventDeliv
...
-		go sa.resolveFailedTokenTransfer(inflight, transfer.LocalID)
-	}
+
+	go sa.resolveFailedTokenTransfer(inflight, transfer.LocalID)
 
 	return nil
 }
 
 func (sa *syncAsyncBridge) handleApprovalOpFailedEvent(event *fftypes.EventDelivery) error {
+	// See if this is a failure of an inflight token transfer operation
```
awrichar, line 391: token approval operation

internal/syncasync/sync_async_bridge.go
```diff
@@ -341,11 +403,8 @@ func (sa *syncAsyncBridge) handleApprovalOpFailedEvent(event *fftypes.EventDeliv
 	if err := txcommon.RetrieveTokenApprovalInputs(sa.ctx, op, &approval); err != nil {
 		log.L(sa.ctx).Warnf("Failed to extract token approval inputs for operation '%s': %s", op.ID, err)
 	}
-	// See if this is a failure of an inflight token transfer operation
-	inflight := sa.getInFlight(event.Namespace, tokenApproveConfirm, approval.LocalID)
-	if inflight != nil {
-		go sa.resolveFailedTokenTransfer(inflight, approval.LocalID)
-	}
+
+	go sa.resolveFailedTokenTransfer(inflight, approval.LocalID)
```
awrichar, line 407: I see this was preserved, but it looks wrong..?

test/e2e/e2e_test.go
```diff
@@ -233,8 +244,12 @@ func beforeE2ETest(t *testing.T) *testState {
 		t.Logf("Waiting for 2 orgs to appear. Currently have: node1=%d node2=%d", len(orgsC1), len(orgsC2))
 		time.Sleep(3 * time.Second)
 	}
+	ts.org1key = GetIdentityBlockchainKeys(t, ts.client1, ts.org1.ID, 200)[0]
+	ts.org2key = GetIdentityBlockchainKeys(t, ts.client2, ts.org2.ID, 200)[0]
+	t.Logf("Org1: ID=%s DID=%s Key=%s", ts.org1.DID, ts.org1.ID, ts.org1key.Value)
+	t.Logf("Org2: ID=%s DID=%s Key=%s", ts.org2.DID, ts.org2.ID, ts.org2key.Value)
 
-	eventNames := "message_confirmed|token_pool_confirmed|token_transfer_confirmed|blockchain_event|token_approval_confirmed"
+	eventNames := "message_confirmed|token_pool_confirmed|token_transfer_confirmed|blockchain_event|token_approval_confirmed|identity_confirmed"
```
awrichar, line 252: Maybe not the time, but I feel like we should probably just remove this filter...

awrichar (commented):
Alright, finally read this whole thing! Looks really good - most of my comments are just around making sure I understand it correctly and flagging places we could potentially make it more clear for others.

internal/apiserver/route_get_identity_by_id.go
peterbroadhurst: Fixed this 👍 

Also added `/api/v1/verifiers` resource collection, because I realized there was no API for reverse lookups any more without that.

internal/config/config.go
peterbroadhurst: ok - that's done

internal/broadcast/definition.go
peterbroadhurst: Resolved with the `normalization` rename

internal/identity/identitymanager.go
peterbroadhurst: Resolved alongside the `normalization` rename

internal/events/aggregator.go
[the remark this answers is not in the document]
peterbroadhurst: I've created a bigger `HandlerResult` structure to return that can contain a `CustomCorrelator` in addition to an `Action`

internal/orchestrator/status.go
```diff
@@ -43,34 +44,35 @@ func (or *orchestrator) GetNodeUUID(ctx context.Context) (node *fftypes.UUID) {
...
 
-	org, err := or.database.GetOrganizationByName(ctx, status.Org.Name)
-	if err != nil {
-		return nil, err
-	}
 	if org != nil {
 		status.Org.Registered = true
 		status.Org.ID = org.ID
-		status.Org.Identity = org.Identity
+		status.Org.Identity = org.DID
```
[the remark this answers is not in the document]
peterbroadhurst: I'll go ahead and make the change. I think this is the last place we have Identity, so I agree that change makes sense.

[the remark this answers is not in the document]
peterbroadhurst: I believe the `registered` field is what all current automation around this `/status` API depends on 🤞

pkg/fftypes/identity.go
peterbroadhurst: It does get used, but it is confusing. I'll make the change to favor clarity in the structures, vs. consistency with the `Definition` interface

peterbroadhurst: It fell out in the wash in the end, as the claim processing logic actually sets the `messages.verification` ID on the `Identity` (not the verification processing logic). Thanks for spotting this cleanup.

internal/definitions/definition_handler_identity_claim.go
peterbroadhurst: > root identities 

They have self-signed claims. The `messages.verifier` entry will be empty.

> nodes have self-signed claims 

They have claims signed by the `parent` directly at this time, as Data Exchange keys have not been made first class such that they can sign a claim. Only blockchain keys can do that at this time. The `messages.verifier` entry will be empty.

internal/definitions/definition_handler_identity_claim.go
peterbroadhurst: It allows us to have consistent logging outside of the `foundAll` branch

pkg/fftypes/identity.go
peterbroadhurst: Clarified comment - I got a better understanding after I wrote this one.

peterbroadhurst: Actually, might go the same route as with verifications and rip this out. I was trying for consistency, but the pattern of there being multiple messages for this definition just didn't fit in the end.

peterbroadhurst: ... a future PR could go a step further and rip out the `SetBroadcastMessage()` function completely, as it's not adding a massive amount of value at this point.

internal/syncasync/sync_async_bridge.go
peterbroadhurst: good spot - looks like we were missing a helper function and message for failed approvals. Added

peterbroadhurst:
Thanks Andrew for the detailed review, and excellent suggestions.

I've gone through each of the comments and addressed them with code/comment changes.

Before we merge I'm planning to re-run the migration test I mentioned above one last time, but would be great to get your feedback first (as the migration test takes a while to set up and run).

pkg/fftypes/identity.go
awrichar: Yep, I came to that same conclusion after another look. There are only a few paths that reap the value of `SetBroadcastMessage`; many other places are manually adding the message reference in a different way.

db/migrations/postgres/000063_create_identities_table.up.sql
```diff
@@ -0,0 +1,143 @@
...
+  vtype          VARCHAR(256)    NOT NULL,
+  namespace      VARCHAR(64)     NOT NULL,
+  value          TEXT            NOT NULL,
+  created        BIGINT          NOT NULL
+);
+
+CREATE UNIQUE INDEX verifiers_hash ON verifiers(hash);
+CREATE UNIQUE INDEX verifiers_value ON verifiers(vtype, namespace, value);
+CREATE UNIQUE INDEX verifiers_identity ON verifiers(identity);
```
awrichar, line 36: So we require that each verifier must have a 1:1 relationship with an identity (and an identity cannot have more than one verifier). But it is possible for an identity to exist _without_ a verifier in certain cases.

db/migrations/postgres/000063_create_identities_table.up.sql
```diff
@@ -0,0 +1,143 @@
...
+
+INSERT INTO verifiers (
+    hash,
+    namespace,
+    identity,
+    vtype,
+    value,
+    created
+  ) SELECT 
+    REPLACE(o.id, '-', ''),
```
awrichar, line 100: Worth a comment here on the kludge to fake a hash?

db/migrations/postgres/000063_create_identities_table.down.sql
```diff
@@ -0,0 +1,97 @@
...
+    i.profile,
+    i.created,
+    v.value as identity
+  FROM identities as i
+  LEFT JOIN verifiers v ON v.hash = REPLACE(i.id,'-','')
+  LEFT JOIN verifiers pv ON pv.hash = REPLACE(i.parent,'-','')
+  WHERE i.did LIKE 'did:firefly:org/%' AND v.hash IS NOT NULL;
+
+-- We only reconstitute nodes that were dropped during the original up migration.
+-- These have the UUID of the verifier set to the same UUID as the node.
```
awrichar, line 61: They have the hash of the verifier set to the UUID of the node (with dashes removed)

db/migrations/postgres/000063_create_identities_table.up.sql
awrichar: This feels good to me. Suggested a few comments here in the migration file for anyone looking, but this definitely needs to make it into release notes as well.

internal/apiserver/route_get_verifier_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var getVerifierByID = &oapispec.Route{
+	Name:   "getVerifierByID",
+	Path:   "namespaces/{ns}/verifiers/{vid}",
+	Method: http.MethodGet,
+	PathParams: []*oapispec.PathParam{
+		{Name: "ns", ExampleFromConf: config.NamespacesDefault, Description: i18n.MsgTBD},
+		{Name: "vid", Example: "id", Description: i18n.MsgTBD},
```
awrichar, line 34: Just want to note that this became a "hash" rather than an "id" in the latest pass. Might suggest a tweak to reflect that on the Swagger?

internal/identity/identitymanager.go
```diff
@@ -77,192 +89,388 @@ func NewIdentityManager(ctx context.Context, di database.Plugin, ii identity.Plu
...
+		msgSignerRef := &fftypes.SignerRef{}
+		err = im.ResolveNodeOwnerSigningIdentity(ctx, msgSignerRef)
+		if err != nil {
+			return "", err
+		}
+		return msgSignerRef.Key, nil
+	}
+	// If the caller is not confident that the blockchain plugin/connector should be used to resolve,
+	// for example it might be a different blockchain (Eth vs Fabric etc.), or it has it's own
+	// verification/management of keys, it should set resolveViaBlockchainPlugin=false
```
awrichar, line 114: resolveViaBlockchainPlugin was renamed, right?

internal/i18n/en_translations.go
```diff
@@ -266,4 +265,25 @@ var (
 	MsgFFIGenerationFailed          = ffm("FF10346", "Error generating smart contract interface: %s", 400)
 	MsgFFIGenerationUnsupported     = ffm("FF10347", "Smart contract interface generation is not supported by this blockchain plugin", 400)
 	MsgBlobHashMismatch             = ffm("FF10348", "Blob hash mismatch sent=%s received=%s", 400)
+	MsgDIDResovlerUnknown           = ffm("FF10349", "DID resolver unknown for DID: %s", 400)
```
awrichar, line 268: typo: Resovler

internal/definitions/definition_handler.go
```diff
@@ -56,35 +63,44 @@ const (
 	ActionWait
 )
 
-// DefinitionBatchActions are actions to be taken at the end of a definition batch
+// DefinitionBatchState tracks the state between definition handlers that run in-line on the pin processing route in the
+// aggregator as part of a batch of pins. They might have complex API calls, and interdependencies, that need to be managed via this state.
+// The actions to be taken at the end of a definition batch.
 // See further notes on "batchActions" in the event aggregator
```
awrichar, line 69: Minor note - the aggregator type has been renamed "batchState" at this point

internal/definitions/definition_handler_identity_verification.go
```diff
@@ -0,0 +1,86 @@
...
+	if parent.DID != verifyMsg.Header.Author {
+		log.L(ctx).Warnf("Invalid verification message %s - parent '%s' does not match signer '%s'", verifyMsg.Header.ID, parent.DID, verifyMsg.Header.Author)
+		return HandlerResult{Action: ActionReject}, nil
+	}
+
+	// At this point, this is a valid verification, but we don't know if the claim has arrived.
+	// It might be being processed in the same pin batch as us - so we can't
+
+	// See if the message has already arrived, if so we need to queue a rewind to it
+	claimMsg, err := dh.database.GetMessageByID(ctx, verification.Claim.ID)
```
awrichar, line 58: Would it make sense to check the pending messages in the batch _before_ hitting the database? Unless we expect the database to be the more common case. Probably getting into micro-optimization at this point anyway.

internal/events/aggregator.go
awrichar: Looks like this one still fell through the cracks?

internal/definitions/definition_handler_identity_claim.go
awrichar: This one also might still need a change?

pkg/fftypes/identity.go
peterbroadhurst: cool - resolving this for now, but we can come back to this if it confuses us again in the future

db/migrations/postgres/000063_create_identities_table.up.sql
peterbroadhurst: > So we require that each verifier must have a 1:1 relationship with an identity

Not quite. `Verifier - many - one - Identity`

peterbroadhurst: Ha, well now I see why you made that comment though! Let me work out what's going on with that Index.

peterbroadhurst: The `UNIQUE` keyword is a typo - thanks for catching

internal/apiserver/route_get_verifier_by_id.go
peterbroadhurst: I actually thought about that and didn't do it. My thinking was that it's actually a resource identifier, just one that's deterministically generated on all nodes. We don't rely on it being a hash for processing, and due to an edge case on migration there's a small number of environments where it might not actually be the Hash.

pkg/database/plugin.go
```diff
@@ -802,27 +803,31 @@ var PinQueryFactory = &queryFields{
...
+	"name":                  &StringField{},
+	"description":           &StringField{},
+	"profile":               &JSONField{},
+	"created":               &TimeField{},
+	"updated":               &TimeField{},
+}
+
+// VerifierQueryFactory filter fields for identities
+var VerifierQueryFactory = &queryFields{
+	"id":        &UUIDField{},
```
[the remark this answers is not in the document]
peterbroadhurst: Thanks for spotting this 👍

internal/definitions/definition_handler_identity_verification.go
peterbroadhurst: Your suggestion feels right to me - will do it.

peterbroadhurst: Actually, it did make the logic a lot less clean... so going to defer this for now (the database case is more likely, as my best guess):

```go
	// See if the message has already arrived, if so we need to queue a rewind to it
	claimMsg := state.GetPendingConfirm()[*verification.Claim.ID]
	if claimMsg == nil {
		claimMsg, err := dh.database.GetMessageByID(ctx, verification.Claim.ID)
		if err != nil {
			return HandlerResult{Action: ActionRetry}, err
		}
		if claimMsg != nil && claimMsg.State != fftypes.MessageStateConfirmed {
			claimMsg = nil
		}
	}
```

peterbroadhurst:
Moving onto the migration test re-run after the `hash` change

peterbroadhurst:
Sadly the cheat of using a UUID to fill in the bytes didn't work. First problem was `UUID` cannot be used in a `REPLACE` as a string, then the second problem was after I did `REPLACE(o.id::text, '-', '')` I ended up with too few characters `c0f5e5b543a7477daf1e33f96702d995` (UUID is 16B, not 32).

```
postgres=# select hash from verifiers;
                               hash
------------------------------------------------------------------
 c0f5e5b543a7477daf1e33f96702d995
 f4ac3f2dab2b4da096ee70a13d2c21ba
 bc0ae3c7d7574fac9e0ef49017cf972b
 5f6003a1461d4c37a68ddf88dfe07f3f
(4 rows)
```

Then the REST API code on retrieval failed with `"FF10121: Database resultset read error from table 'verifiers': sql: Scan error on column index 0, name \"hash\": encoding/hex: invalid byte: U+0020 ' '"`:

So I can write them twice, but it's feeling very messy at this point 🤔

peterbroadhurst:
I added the double-UUID extension to the migration code, and tested the migration e2e.

Racked my brains, and couldn't think of another simpler solution to supporting up+down migrations for this table split. I considered allocating a eyecatcher prefixed ID in the `up` migration, and then doing a Go based proper Hash calculation. But this brakes the ability to down migration.

So let me know if you have any further inspiration @awrichar - otherwise, I believe this is all tested and good to go.

internal/i18n/errors.go
awrichar: Assuming we keep this new behavior, we might consider renaming it to `NewWrappedError` or similar. I think `WrapError` is potentially confusing since it now intentionally behaves differently from `errors.Wrap`.

internal/definitions/definition_handler_identity_verification.go
awrichar: Works for me; thanks for giving it a pass.

internal/apiserver/route_get_verifier_by_id.go
awrichar: OK - I'm mainly worried about the case where I query `/verifiers` and get a list of things, how do I know which one to put in `vid` in order to look up that one item? I have to know that `hash` maps to `vid`. Of course this is a detail that can easily be tweaked later.

awrichar:
Yea, I was suspicious that the UUID to hash hack wouldn't quite work. I think where it's landed is fine though.

I considered the fact that Postgres could probably do actual hashing - but it would be a problem for sqlite, and would make the down migration harder in any case.

Looks good to me now - will leave the button push to you!

awrichar (approved):

PR #555, opened 2022-02-28
Explicitly name all subscriptions

### LINKED ISSUE #554: FireFly v0.13.0 is not specifying a name for subscriptions
In this change:
https://github.com/hyperledger/firefly/commit/98890dc0dd07ad15d5486916902e95e0ecb20f03

We moved to use the flat `POST` `/subscriptions` route on EthConnect, rather than requiring an ABI to be deployed to EthConnect first.

However, we seem to have missed specifying the `name` when we create it - so this check fails on every startup, and we're creating duplicate subscriptions for the same event:
https://github.com/hyperledger/firefly/blob/a4353742a3b8955eed6e989061944968558e751c/internal/blockchain/ethereum/eventstream.go#L177-L185

peterbroadhurst:
Fixes #554

### DIFF
--- a/internal/blockchain/ethereum/ethereum.go
+++ b/internal/blockchain/ethereum/ethereum.go
@@ -615,7 +615,8 @@ func (e *Ethereum) AddSubscription(ctx context.Context, subscription *fftypes.Co
 		return i18n.WrapError(ctx, err, i18n.MsgContractParamInvalid)
 	}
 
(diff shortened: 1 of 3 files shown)


nguyer (approved):
Looks good. Thanks for catching this. Is it worth adding a unit test to check this to prevent similar issues in the future?

PR #577, opened 2022-03-03
Add missing itype column in SQLite migration

peterbroadhurst:
The testing for #549 focussed on PostgreSQL and one of the updates to the migration got missed in the SQLite one.

Raising and will do a test before looking for a review.


peterbroadhurst:
I've tested this migration now with SQLite from `v0.13.1`

nguyer (approved):

PR #579, opened 2022-03-03
Enhance subscription filters & event enrichment

shorsher:
- Support for Transaction and BlockchainEvent event filtering
 - Added ws event enrichment for transactions and blockchain events


db/migrations/postgres/000070_add_subscriptions_filters.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+ALTER TABLE subscriptions ADD COLUMN filters TEXT;
```
awrichar, line 2: We do need to migrate existing filters. You may be able to build the JSON yourself with something like:
```
UPDATE subscriptions SET filters='{"events":"' || filter_events || '","topics":"' || filter_topics || '"}';
```
^ totally untested, but maybe a starting point to try out.

internal/events/event_dispatcher.go
```diff
@@ -217,27 +254,51 @@ func (ed *eventDispatcher) filterEvents(candidates []*fftypes.EventDelivery) []*
...
-		if filter.tagFilter != nil && !filter.tagFilter.MatchString(tag) {
-			continue
+
+		if tx != nil {
+			txType = tx.Type.String()
 		}
-		if filter.authorFilter != nil && !filter.authorFilter.MatchString(author) {
-			continue
+
+		if be != nil {
```
nguyer, line 262: If be != nil, that's just how it be

internal/events/subscription_manager.go
```diff
@@ -275,47 +289,84 @@ func (sm *subscriptionManager) parseSubscriptionDef(ctx context.Context, subDef
...
+			nameFilter, err = regexp.Compile(filter.BlockchainEvent.Name)
+			if err != nil {
+				return nil, i18n.WrapError(ctx, err, i18n.MsgRegexpCompileFailed, "filter.BlockchainEvent.name", filter.BlockchainEvent.Name)
+			}
+		}
+
+		bf := &blockchainFilter{
+			nameFilter: nameFilter,
 		}
+		sub.blockchainFilter = bf
```
nguyer, line 352: This is good, and I would keep this, but I'd also like to see filtering by `ContractListener` here as well. That's the most specific thing we can filter by for BlockchainEvents. Just filtering on name is not bad, but for example, if you had multiple ERC20 contracts both emitting `Transfer` events, this filter could pick up both of them.

pkg/fftypes/subscription.go
```diff
@@ -26,13 +26,28 @@ import (
...
 	Tag    string `json:"tag,omitempty"`
 	Group  string `json:"group,omitempty"`
 	Author string `json:"author,omitempty"`
 }
 
+type TransactionFilter struct {
+	Type string `json:"type,omitempty"`
+}
+
+// TODO: Add listener ID
```
nguyer, line 46: Please do 🙂

nguyer (changes requested):
Looks really good. One minor request for one more thing to add re: ContractSubscription IDs. But other than that, looks good to me. Thanks!

internal/events/subscription_manager.go
```diff
@@ -275,47 +290,93 @@ func (sm *subscriptionManager) parseSubscriptionDef(ctx context.Context, subDef
...
-	}
 
-	var topicsFilter *regexp.Regexp
-	if filter.Topics != "" {
-		topicsFilter, err = regexp.Compile(filter.Topics)
-		if err != nil {
-			return nil, i18n.WrapError(ctx, err, i18n.MsgRegexpCompileFailed, "filter.topics", filter.Topics)
+		var groupFilter *regexp.Regexp
+		if filter.Message.Group != "" {
+			groupFilter, err = regexp.Compile(filter.Message.Group)
```
peterbroadhurst, line 310: I think we need to handle `filter.DeprecatedGroup` to refer to the old JSON input fields.

Either by referencing both here (with a deprecation warning in the logs on the latter), or at a point up the stack (that's common to durable subscriptions and ephemeral) we copy the value over before we enter this code.

internal/events/websockets/websocket_connection.go
```diff
@@ -88,9 +88,11 @@ func (wc *websocketConnection) processAutoStart(req *http.Request) {
 			Name:      query.Get("name"),
 			Filter: fftypes.SubscriptionFilter{
 				Events: query.Get("filter.events"),
-				Topics: query.Get("filter.topics"),
-				Group:  query.Get("filter.group"),
-				Tag:    query.Get("filter.tag"),
+				Message: fftypes.MessageFilter{
```
peterbroadhurst, line 91: Think we need to handle the old options too, to avoid breaking apps.

Also the docs have examples that use these, which should use the new syntax.

internal/txcommon/event_enrich.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	switch event.Type {
+	case fftypes.EventTypeTransactionSubmitted:
+		tx, err := t.database.GetTransactionByID(ctx, event.Reference)
+		if err != nil {
+			return nil, err
+		}
+		e.Transaction = tx
+	case fftypes.EventTypeMessageRejected:
```
peterbroadhurst, line 37: Just a note that you can comma separate these, rather than using `fallthrough`:
```go
case fftypes.EventTypeMessageConfirmed, fftypes.EventTypeMessageRejected:
```

pkg/fftypes/subscription.go
```diff
@@ -26,13 +26,28 @@ import (
 
 // SubscriptionFilter contains regular expressions to match against events. All must match for an event to be dispatched to a subscription
 type SubscriptionFilter struct {
-	Events string `json:"events,omitempty"`
+	Events          string                `json:"events,omitempty"`
+	Message         MessageFilter         `json:"message,omitempty"`
+	Transaction     TransactionFilter     `json:"transaction,omitempty"`
+	BlockchainEvent BlockchainEventFilter `json:"blockchainevent,omitempty"`
```
peterbroadhurst, line 32: See earlier comments about being able to parse in the old options at the top level:

```
	DeprecatedTopics string `json:"topics,omitempty"`
	DeprecatedTag    string `json:"tag,omitempty"`
	DeprecatedGroup  string `json:"group,omitempty"`
	DeprecatedAuthor string `json:"author,omitempty"`
```

peterbroadhurst (changes requested):
Great stuff @shorsher !
Just a couple of migration/docs comments that I think would be high impact on existing users if we didn't address, and one style suggestion.

peterbroadhurst (approved):
Thanks for moving the subscription filter and handling the deprecation/migration 👍

Issue #568, opened 2022-03-02
Operations: Add operation for Shared Storage (IPFS) uploads, and move API calls to batch worker


peterbroadhurst:

awrichar:
Took an initial look at this, and found one blocker. Currently we write the public IPFS reference as part of the batch payload, which is hashed and stored by the batch processor _before_ being passed to the broadcast manager for dispatch. This means that the IPFS upload must happen prior to all of this in order to compute a valid batch hash, which implies it cannot happen during the dispatch step.

There are roughly two routes forward (although there may be many details on how to achieve either one):
1. Break the batch processing into more stages, such that the IPFS upload can happen during batch processing but prior to hashing and storing the batch
2. Perform the IPFS upload when dispatching the batch, but adjust the batch payload so that the IPFS reference does not invalidate the hash when it is added

I'd lean slightly toward the latter, although this probably needs some discussion and may tie into the batch payload changes in #506.

awrichar:
Part of this work is now accomplished in #596

PR #582, opened 2022-03-04
Update batch logic to only store+hash the manifest

### LINKED ISSUE #506: Only store the batch manifest
A couple of important things have clicked into place:
- #403 - we added the concept of a manifest to a batch, so DX implementations can confirm what the received without needing to sign the whole massive payload of the batch
- #501 - we made it so all message transfers include a batch, including unpinned
- #501 - the batch logic only stores the batch at the end of the dispatch, as part of passing it to a transport plugin (IPFS/DX)
- When a batch arrives, we always store all of the _contents_ of the batch individually - the messages and data

When you add all these together, the only reason you need to store the full batch is because the on-chain transaction contains a `hash` that is a function of the entire batch payload.

If instead we did the following small things:
- Include the TX type/ID in the manifest
- Include the batch ID in the manifest
- Make the `hash` of a batch, a function only of the manifest

... then all we would need to store is the manifest.

peterbroadhurst:
Resolves #506 

A few notes on the implementation.

Improvements related to performance:
- The hash on a batch is now just a hash of the manifest, rather than the full payload
  - `tx` was added to the manifest to include in the hash
- The database object for a batch is now the manifest
- The manifest has been updated to include everything the batch aggregator needs to find pins
  - A count of the `topics` were needed for this
- We now have a cache for messages + all data associated with a message

Improvements related to debug:
- Added to-string helpers to definition batch actions, and log the results
- Added a `GET` `/status/pins` collection to peekinside the pins status

Migration:
- The code copes with a persisted batch of the old type stored in the DB
  - `Version` in manifest used to distinguish this, and provide future extensibility
- The code copes with processing a batch that has a payload hash, rather than a manifest hash
  - To handle late-join/re-sync to a network processing old broadcasts

Potential follow-on work:
- Update the code in #587 to use the message cache
- Add a `batch` cache - to help the aggregator logic when managing pins
- Add a `transaction` cache - to help #587 event enrichment
- Trawl for any remaining `GetMessageByID` calls

## Message/data cache implementation notes

Messages have fields that are mutable, in two categories

  1. Can change multiple times like `state` - you *cannot* rely on the cache for these
  2. Can go from being un-set, to being set, and once set are immutable.

For (2) the cache provides a set of `CacheReadOption` modifiers that makes it safe to query the cache, even if the cache we slow to update asynchronously (active/active cluster being the ultimate example here, but from code inspection this is possible in the current cache).

If you use `CRORequestBatchID` then the cache will return a miss, if there is no `BatchID` set.

If you use `CRORequirePins` then the cache will return a miss, if the number of pins does not match the number of topics in the message.


pkg/fftypes/data.go
```diff
@@ -52,6 +52,38 @@ type Data struct {
 	ValueSize int64 `json:"-"` // Used internally for message size calcuation, without full payload retrieval
 }
 
+func (br *BlobRef) BatchBlobRef(broadcast bool) *BlobRef {
+	if br == nil {
+		return nil
+	}
+	// For broadcast data the blob reference contains the "public" (shared storage) reference, which
+	// must have been allocated to this data item before sealing the batch.
```
awrichar, line 60: This is the operative decision for #568 - I'm still wondering if it's possible for the public reference to be removed from the batch hash altogether. Perhaps it could be store on the `Blob` instead of on the `BlobRef`?

awrichar: It's mainly odd that private blob transfer happens _after_ batch sealing, but public blob transfer must happen _before_ batch sealing in order to include the IPFS ref. Seems like they should happen in the same order regardless.

When a node receives an IPFS ref, it does do some checking of the fetched contents to verify they match the blob hash before recording the blob as received. The question is whether this is "good enough" to say we can send IPFS refs without including them in the batch's hash proof.

peterbroadhurst: > Seems like they should happen in the same order regardless.

I do not think this is possible, as IPFS does not have a "messaging" capability. It's just a storage system.

internal/data/data_manager.go
```diff
@@ -74,6 +98,11 @@ func NewDataManager(ctx context.Context, di database.Plugin, pi sharedstorage.Pl
 		ccache.Configure().
 			MaxSize(config.GetByteSize(config.ValidatorCacheSize)),
 	)
+	dm.messageCache = ccache.New(
+		// We use a LRU cache with a size-aware max
+		ccache.Configure().
+			MaxSize(config.GetByteSize(config.MessageCacheTTL)),
```
awrichar, line 104: Should this be `MessageCacheSize`?

internal/data/data_manager.go
```diff
@@ -52,8 +56,28 @@ type dataManager struct {
...
+	msg  *fftypes.Message
+	data []*fftypes.Data
+	size int64
+}
+
+func (mce *messageCacheEntry) Size() int64 {
+	return mce.size
+}
+
+type CacheReadOption int
```
awrichar, line 73: Might need some comments on what this is for, unless I missed them elsewhere in the code?

awrichar: I got the explanation from the PR description, but it should probably make its way into the code too.

internal/data/data_manager.go
```diff
@@ -265,83 +381,114 @@ func (dm *dataManager) validateAndStore(ctx context.Context, ns string, validato
...
+				Data: d,
 				Blob: blob,
 			})
 		}
 	}
-	return refs, dataToPublish, nil
+	return data, dataToPublish, nil
+}
+
+// HydrateBatch fetches the full messages for a persited batch, ready for transmission
```
awrichar, line 457: typo: persited

internal/batch/batch_manager.go
```diff
@@ -175,10 +177,14 @@ func (bm *batchManager) getProcessor(txType fftypes.TransactionType, msgType fft
 	return processor, nil
 }
 
-func (bm *batchManager) assembleMessageData(msg *fftypes.Message) (data []*fftypes.Data, err error) {
+func (bm *batchManager) assembleMessageData(processor *batchProcessor, msg *fftypes.Message) (retData fftypes.DataArray, err error) {
+	var cro []data.CacheReadOption
+	if processor.conf.DispatcherOptions.BatchType == fftypes.BatchTypeBroadcast {
```
awrichar, line 182: Could consider just passing the BatchType into this method, instead of the processor itself...

internal/batch/batch_processor.go
```diff
@@ -348,47 +358,62 @@ func (bp *batchProcessor) assemblyLoop() {
...
+	//   For example updates to the Blob.Public of a Data entry must be written do the DB,
+	//   and updated in Payload.Data[] array.
+	err = bp.dispatchBatch(state)
 	if err != nil {
 		return err
 	}
 
-	err = bp.markMessagesDispatched(batch)
+	// Dispatched phase: Writes back the changes to the DB, so that these messages will not be
+	//   are all tagged as part of this batch, and won't be included in any future batches.
```
awrichar, line 380: I think this comment needs a fix

internal/batch/batch_processor.go
```diff
@@ -445,81 +476,106 @@ func (bp *batchProcessor) maskContexts(ctx context.Context, batch *fftypes.Batch
...
 		return true, bp.database.RunAsGroup(bp.ctx, func(ctx context.Context) (err error) {
 			// Update all the messages in the batch with the batch ID
-			msgIDs := make([]driver.Value, len(batch.Payload.Messages))
-			for i, msg := range batch.Payload.Messages {
+			msgIDs := make([]driver.Value, len(state.Payload.Messages))
+			for i, msg := range state.Payload.Messages {
 				msgIDs[i] = msg.Header.ID
+				// We don't want to have to read the DB again if we want to query for the batch ID, or pins,
+				// so ensure the copy in our cache gets updated.
+				bp.data.UpdateMessageIfCached(ctx, msg)
```
awrichar, line 549: Want to make sure it's OK that we update the batchID in the cache before we write the batchID to the database itself (ie in case we somehow fail to update the database)... I think it's OK and we would come back around and try to write the same batchID on the second try. Just wanted to put a comment because I was staring at this for a bit.

internal/events/event_dispatcher.go
```diff
@@ -159,6 +159,10 @@ func (ed *eventDispatcher) getEvents(ctx context.Context, filter database.Filter
 }
 
 func (ed *eventDispatcher) enrichEvents(events []fftypes.LocallySequenced) ([]*fftypes.EventDelivery, error) {
+
+	// TODO: Update this to use the message cache, only for event types that have message references.
+	//       Do this after Alex's changes to enrichment are merged
```
awrichar, line 164: I think these changes are merged, but unsure if this TODO was meant to be here or in a follow-up.

internal/events/persist_batch.go
```diff
@@ -31,63 +31,94 @@ func (em *eventManager) persistBatchFromBroadcast(ctx context.Context /* db TX c
...
 
-	return em.persistBatch(ctx, batch)
+	_, valid, err = em.persistBatch(ctx, batch)
+	return valid, err
 }
 
 // persistBatch performs very simple validation on each message/data element (hashes) and either persists
 // or discards them. Errors are returned only in the case of database failures, which should be retried.
-func (em *eventManager) persistBatch(ctx context.Context /* db TX context*/, batch *fftypes.Batch) (valid bool, err error) {
+func (em *eventManager) persistBatch(ctx context.Context /* db TX context*/, batch *fftypes.Batch) (persistedBatch *fftypes.BatchPersisted, valid bool, err error) {
```
awrichar, line 40: Is the (newly added) first return value used anywhere?

awrichar: nvm, see it used in dx_callbacks

internal/events/persist_batch.go
```diff
@@ -31,63 +31,94 @@ func (em *eventManager) persistBatchFromBroadcast(ctx context.Context /* db TX c
...
 	for i, msg := range batch.Payload.Messages {
 		if valid, err = em.persistBatchMessage(ctx, batch, i, msg, optimization); !valid || err != nil {
 			return valid, err
 		}
+		dataInBatch := true
+		msgData := make(fftypes.DataArray, len(msg.Data))
+		for di, dataRef := range msg.Data {
+			msgData[di] = dataByID[*dataRef.ID]
+			if msgData[di] == nil || !msgData[di].Hash.Equals(dataRef.Hash) {
+				log.L(ctx).Debugf("Message '%s' in batch '%s' - data not in-line in batch id='%s' hash='%s'", msg.Header.ID, batch.ID, dataRef.ID, dataRef.Hash)
```
awrichar, line 113: Should this be higher than debug? Is this an expected situation?

awrichar (approved):
Looks good - marking approved, although I did leave a few comments inline that are worth a look before you decide to merge.

internal/apiserver/route_get_status_pins.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/hyperledger/firefly/internal/i18n"
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/database"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var getStatusPins = &oapispec.Route{
+	Name:            "getStatusPins",
+	Path:            "status/pins",
```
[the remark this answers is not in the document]
peterbroadhurst: We could put them at the root, if we wanted to explain them in more detail as a first class object.
My thinking here was they are an internal read-only view of the state for problem diagnosis. Understand it's not perfect, and happy to discuss more.

internal/batch/batch_processor.go
peterbroadhurst: Yep, the only way to avoid this would be to push all of the cache updating (in all cases) to post-commit actions.
I convinced myself that wasn't required, but happy to discuss more.

internal/events/event_dispatcher.go
peterbroadhurst: The TODO has been addressed in #593

internal/events/persist_batch.go
peterbroadhurst: The architecture prior to this code change allows it.
e.g. you could send some data to a party, then send a message referring to that data, without sending that data again.

internal/privatemessaging/privatemessaging.go
```diff
@@ -245,11 +250,14 @@ func (pm *privateMessaging) sendData(ctx context.Context, tw *fftypes.TransportW
 			batch.Namespace,
 			batch.Payload.TX.ID,
 			fftypes.OpTypeDataExchangeBatchSend)
+		op.Input = fftypes.JSONObject{
+			"batch": tw.Batch.ID,
+		}
```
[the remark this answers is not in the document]
peterbroadhurst: Sorry, think this was a merge error

internal/apiserver/route_get_status_pins.go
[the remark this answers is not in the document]
peterbroadhurst: Shall be merging through the chain with this here, but with an open-ness to move it in a future commit

PR #584, opened 2022-03-04
Blockchain metrics for Ethereum

nguyer:
This PR introduces blockchain metrics for:

- FireFly performing a blockchain transaction with a smart contract method
- FireFly querying the blockchain with a smart contract method
- Events emitted by a smart contract that FireFly is listening for

Each of these has two labels `location` and `methodName` for transactions/queries or `signature` (for events). The `location` is a blockchain specific appropriate string for describing where the event came from. The value of this label is 
 in the format of `"<key1>=<value1>,<key2>=<value2>"`. For Ethereum, it is just `address=<contract_address>`.

Fabric metrics will also be added in a subsequent PR.

Resolves https://github.com/hyperledger/firefly/issues/559


pkg/fftypes/blockchainevent.go
```diff
@@ -28,4 +28,6 @@ type BlockchainEvent struct {
 	Info       JSONObject     `json:"info,omitempty"`
 	Timestamp  *FFTime        `json:"timestamp,omitempty"`
 	TX         TransactionRef `json:"tx"`
+	Location   string         `json:"-"`
```
peterbroadhurst, line 31: Is it possible to avoid these?
e.g. only include them on `blockchain.Event`, rather than `fftypes.BlockchainEvent` which is an external API object.

If we keep them here, they definitely need comments to explain they aren't available.

nguyer: Possibly. Let me try that.

internal/events/token_pool_created.go
```diff
@@ -53,6 +53,7 @@ func (em *eventManager) confirmPool(ctx context.Context, pool *fftypes.TokenPool
 	if err := em.persistBlockchainEvent(ctx, chainEvent); err != nil {
 		return err
 	}
+	em.emitBlockchainEventMetric(*ev)
```
peterbroadhurst, line 56: Just wondering why the mem copy here, rather than a pass-by-reference?
I think the blockchain.Event structure is reasonably large.

peterbroadhurst (approved):
Changes look great thanks @nguyer 
Marked approval with one minor efficiency comment to consider (passing `&event.Event` rather than `event.Event`)

PR #586, opened 2022-03-05
Added version command to Core Binary

### LINKED ISSUE #132: Add version Command to Core Binary
Just like in [ethconnect](https://github.com/hyperledger-labs/firefly-ethconnect/blob/dd53ec2afc184da9b72eb92e83b1446d784c69f7/main.go#L26-L32) where we use `ldflags` to add build metadata (commit sha, tags, etc.) to the binary and log it at startup, we should then take it a step further and add a `version` command (similar to what `helm`, `kubectl`, `go` have) that then will output the same information.

TheSwarnim:
Signed-off-by: Swarnim Pratap Singh <swarnimpratap132@gmail.com>
Fixes #132
It provides a command for listing the firefly build version and creating an internal package for the version so that it was accessible to the versionCmd in the cmd package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ $(eval $(call makemock, internal/apiserver,        IServer,            apiserver
 $(eval $(call makemock, internal/metrics,          Manager,            metricsmocks))
 
 firefly-nocgo: ${GOFILES}
-		CGO_ENABLED=0 $(VGO) build -o ${BINARY_NAME}-nocgo -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION)" -tags=prod -tags=prod -v
+		CGO_ENABLED=0 $(VGO) build -o ${BINARY_NAME}-nocgo -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION) -X 'github.com/hyperledger/firefly/internal/version.Date=$(DATE)' -X 'github.com/hyperledger/firefly/version.Commit=${GITREF}" -tags=prod -tags=prod -v
 firefly: ${GOFILES}
 		$(VGO) build -o ${BINARY_NAME} -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION)" -tags=prod -tags=prod -v
 go-mod-tidy: .ALWAYS
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,82 @@
+// Copyright © 2021 Kaleido, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


TheSwarnim:
Hi @nguyer, I've added the relevant files. I'm not able to test the command because it is showing this error. 
image

Makefile
```diff
@@ -70,7 +70,7 @@ $(eval $(call makemock, internal/apiserver,        IServer,            apiserver
 $(eval $(call makemock, internal/metrics,          Manager,            metricsmocks))
 
 firefly-nocgo: ${GOFILES}
-		CGO_ENABLED=0 $(VGO) build -o ${BINARY_NAME}-nocgo -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION)" -tags=prod -tags=prod -v
+		CGO_ENABLED=0 $(VGO) build -o ${BINARY_NAME}-nocgo -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION) -X 'github.com/hyperledger/firefly/internal/version.Date=$(DATE)' -X 'github.com/hyperledger/firefly/version.Commit=${GITREF}" -tags=prod -tags=prod -v
```
nguyer, line 73:
```suggestion
		CGO_ENABLED=0 $(VGO) build -o ${BINARY_NAME}-nocgo -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION) -X 'github.com/hyperledger/firefly/internal/version.Date=$(DATE)' -X 'github.com/hyperledger/firefly/version.Commit=${GITREF}'" -tags=prod -tags=prod -v
```

Looks like you're missing a `'` near the end of this line. That could the cause of the build error you're seeing.

cmd/version.go
```diff
@@ -0,0 +1,82 @@
+// Copyright © 2021 Kaleido, Inc.
```
nguyer, line 1: It's annoying, but the linter will want this to be this year (since this file was created/edited this year)

nguyer (changes requested):
Thanks for contributing this! Looks like there are a few changes that need to be made to get a green build. I left a couple of suggestions on issues that I spotted.

Makefile
hfuss: We'll also want to add a flag for setting the `github.com/hyperledger/firefly/internal/version.Version` to `$(BUILD_VERSION)` as well.

Makefile
```diff
@@ -70,7 +70,7 @@ $(eval $(call makemock, internal/apiserver,        IServer,            apiserver
...
 firefly: ${GOFILES}
 		$(VGO) build -o ${BINARY_NAME} -ldflags "-X main.buildDate=`date -u +\"%Y-%m-%dT%H:%M:%SZ\"` -X main.buildVersion=$(BUILD_VERSION)" -tags=prod -tags=prod -v
```
hfuss, line 75: Also please add the new LD flags to this build target too

Makefile
hfuss: `DATE` will need to be set at the top of the Makefile too, something like:
```make
DATE := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
```

peterbroadhurst (changes requested):
Please can you see https://github.com/hyperledger/firefly-cli/pull/164

I think we might want to pick up those changes.

hfuss:
Closing in favor of #763. I cherry picked your commit @TheSwarnim so that you'll still get some contribution credit. Thank you for getting the ball rolling with this one!

PR #590, opened 2022-03-08
Switch from https to ffdx

gabriel-indik:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,6 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.10
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
 	github.com/microcosm-cc/bluemonday v1.0.16
-	github.com/miracl/conflate v1.2.1 // indirect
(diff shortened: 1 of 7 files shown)


gabriel-indik:
Important: https://github.com/hyperledger/firefly-cli/pull/156 should be merged before this change.

peterbroadhurst (changes requested):
The fact the E2E is failing without https://github.com/hyperledger/firefly-cli/pull/156 merging might mean we need a tweak to the migration code in FireFly:

https://github.com/hyperledger/firefly/blob/00fda51127887df1ca783567f334d8daa0244ebd/internal/dataexchange/dxfactory/factory.go#L32-L42

gabriel-indik:
Closing based on the observations in https://github.com/hyperledger/firefly-cli/pull/156#issuecomment-1063163217

peterbroadhurst:
I do believe we need to fix this default @gabriel-indik, so I've done some work to fix the migration to:

1. Cover the case of `https.url` being set, with `ffdx.url` _not_ being set
2. Avoid initializing the prefix on the default paths, so that things like `ff config` do not show the outdated prefix

PR #592, opened 2022-03-10
Batching message writer routines, to use multi-value insert across API calls on message/data insert

peterbroadhurst:
Note this is in a PR chain with #582


internal/assets/token_transfer.go
```diff
@@ -189,14 +190,14 @@ func (s *transferSender) resolveAndSend(ctx context.Context, method sendMethod)
 	return s.sendInternal(ctx, method)
 }
 
-func (s *transferSender) resolve(ctx context.Context) error {
+func (s *transferSender) resolve(ctx context.Context) (err error) {
 	// Resolve the attached message
 	if s.transfer.Message != nil {
-		sender, err := s.buildTransferMessage(ctx, s.namespace, s.transfer.Message)
+		s.msgSender, err = s.buildTransferMessage(ctx, s.namespace, s.transfer.Message)
```
awrichar, line 196: Seems like `msgSender` is only used within the context of this method and not reused if you call it twice. Should it just be a local instead of being on the struct?

internal/privatemessaging/message.go
```diff
@@ -19,18 +19,21 @@ package privatemessaging
...
-	"github.com/hyperledger/firefly/pkg/database"
 	"github.com/hyperledger/firefly/pkg/fftypes"
 )
 
 func (pm *privateMessaging) NewMessage(ns string, in *fftypes.MessageInOut) sysmessaging.MessageSender {
 	message := &messageSender{
 		mgr:       pm,
 		namespace: ns,
-		msg:       in,
+		group:     in.Group,
```
awrichar, line 33: Where is `group` used?

awrichar (commented):
LGTM but will need a merge from main after pre-requisites go in. Left a few notes inline for consideration.

internal/privatemessaging/message.go
peterbroadhurst: Thanks, this was due to an earlier branch where I separated out the `InData` - but then I reverted to using a full `MessageInOut`

internal/assets/token_transfer.go
peterbroadhurst: It is used to do the `Staged` write at the end

awrichar (approved):

PR #593, opened 2022-03-10
Event lookup cache

peterbroadhurst:


internal/batch/batch_manager.go
```diff
@@ -343,10 +347,14 @@ func (bm *batchManager) getProcessors() []*batchProcessor {
 	bm.dispatcherMux.Lock()
 	defer bm.dispatcherMux.Unlock()
 
+	exists := make(map[*batchProcessor]bool)
 	var processors []*batchProcessor
 	for _, d := range bm.dispatchers {
 		for _, p := range d.processors {
-			processors = append(processors, p)
+			if !exists[p] {
+				processors = append(processors, p)
```
awrichar, line 355: Seems fine... but why does it contain duplicates in the first place? Guess we can have a processor shared by more than one dispatcher?

internal/contracts/manager.go
```diff
@@ -70,7 +71,7 @@ type contractManager struct {
 	operations        operations.Manager
 }
 
-func NewContractManager(ctx context.Context, di database.Plugin, bm broadcast.Manager, im identity.Manager, bi blockchain.Plugin, om operations.Manager) (Manager, error) {
+func NewContractManager(ctx context.Context, di database.Plugin, dm data.Manager, bm broadcast.Manager, im identity.Manager, bi blockchain.Plugin, om operations.Manager, txHelper txcommon.Helper) (Manager, error) {
```
awrichar, line 74: We're not using `data.Manager` here, are we?

internal/txcommon/txcommon.go
```diff
@@ -19,27 +19,60 @@ package txcommon
...
-	return &transactionHelper{
+func NewTransactionHelper(di database.Plugin, dm data.Manager) Helper {
+	t := &transactionHelper{
 		database: di,
+		data:     dm,
 	}
+	t.transactionCache = ccache.New(
+		// We use a LRU cache with a size-aware max
+		ccache.Configure().
+			MaxSize(config.GetByteSize(config.MessageCacheTTL)),
```
awrichar, line 55: I think you did it again; this should be `Size` and not `TTL` right?

awrichar (approved):
Marking approved with notes to fix a few small items

internal/batch/batch_manager.go
peterbroadhurst:
```go
for _, d := range bm.dispatchers {
```
This is iterating over a map, where a dispatcher can register itself multiple times.

I've updated the code to maintain a list as well as a map, to avoid this confusion

awrichar: Makes sense, thanks for the tweak

PR #596, opened 2022-03-11
Move blob publish to broadcast batch dispatch, ensure public refs on inbound

### LINKED ISSUE #595: Blob retrieval issue in test with message cache hit
Seen in https://github.com/hyperledger/firefly/runs/5506592801?check_suite_focus=true

[logs 2.txt](https://github.com/hyperledger/firefly/files/8232028/logs.2.txt)

```
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.953Z] DEBUG Aggregating pin 0000000030 batch=e5b9fe2c-a2c6-447c-a38a-55251f5d7ad9 msg=99b49cbb-84c3-42ad-8794-4b079be0a95d pinIndex=0 msgBaseIndex=0 hash=9e065a7cbddfc57be742bc32956674c3c389521ac2bbb1dce0500d5131fede75 masked=false dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.953Z] DEBUG Returning msg 99b49cbb-84c3-42ad-8794-4b079be0a95d from cache dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.954Z] DEBUG SQL-> query: SELECT masked, hash, batch_id, idx, signer, dispatched, created, seq FROM pins WHERE (hash = $1 AND dispatched = $2 AND seq < $3) ORDER BY seq DESC LIMIT 1 dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.954Z] DEBUG SQL<- query dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.954Z] DEBUG Attempt dispatch msg=99b49cbb-84c3-42ad-8794-4b079be0a95d broadcastContexts=[9e065a7cbddfc57be742bc32956674c3c389521ac2bbb1dce0500d5131fede75] privatePins= dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.954Z] DEBUG SQL-> query: SELECT hash, payload_ref, peer, created, size, seq FROM blobs WHERE hash = $1 LIMIT 1 dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.954Z] DEBUG SQL<- query dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.955Z] DEBUG Blob '2a0dd46679f619725ea903b844af4d43c966ad945f5ee2c7f7dbd9555ca6e786' not found dbtx=wFMXZVUv pid=1 role=aggregator
[36;1mfirefly_core_1_1  |[0m [2022-03-11T05:13:28.955Z] DEBUG Blob '2a0dd46679f619725ea903b844af4d43c966ad945f5ee2c7f7dbd9555ca6e786' not available dbtx=wFMXZVUv pid=1 role=aggregator
```

peterbroadhurst:
Resolves #595 

- No longer set `CRORequirePublicBlobRefs` when assembling the messages prior to dispatching a batch (timing issue with this seen in #595)
- Move upload of blobs from API thread, to the batch dispatch thread
- Set `CRORequirePublicBlobRefs` when aggregating pins for broadcast messages

### DIFF
--- a/internal/broadcast/manager.go
+++ b/internal/broadcast/manager.go
@@ -122,6 +122,12 @@ func (bm *broadcastManager) Name() string {
 }
 
 func (bm *broadcastManager) dispatchBatch(ctx context.Context, state *batch.DispatchState) error {
+
+	// Ensure all the blobs are published
+	if err := bm.publishBlobs(ctx, state.Payload.Data); err != nil {
+		return err
+	}
+
 	// The completed SharedStorage upload
 	op := fftypes.NewOperation(
 		bm.sharedstorage,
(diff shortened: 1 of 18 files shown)


internal/broadcast/manager.go
```diff
@@ -144,21 +150,32 @@ func (bm *broadcastManager) dispatchBatch(ctx context.Context, state *batch.Disp
...
+
+			// Stream from the local data exchange ...
+			reader, err := bm.exchange.DownloadBLOB(ctx, blob.PayloadRef)
+			if err != nil {
+				return i18n.WrapError(ctx, err, i18n.MsgDownloadBlobFailed, blob.PayloadRef)
+			}
+			defer reader.Close()
+
+			// ... to the shared storage
+			d.Blob.Public, err = bm.sharedstorage.PublishData(ctx, reader)
```
awrichar, line 173: This was the part where I got stuck before. How does this public ref 1) get stored in the database, and 2) not cause the data hash to change? Hasn't the batch been sealed and stored before we reach here?

awrichar: After looking further, it looks like the answer to (2) is that the hash calculation changed to not include the public ref as part of the hash. That's good.

Still can't figure out (1), ie where this data item is updated in the database with the public reference received here. It's definitely updated when it's received again in the event aggregator, but I'd expect it's updated prior to that as well... I'm likely just missing something.

awrichar (approved):
Approved, but I left one big question above for the sake of my own understanding.

internal/broadcast/manager.go
peterbroadhurst: Great point. It's _not_ currently at the point the batch is written to the comms layer. As you observe, it is when it comes back in the other end at the end of this PR - but I actually optimized that away in a later change.

I will add it back in, as it's only a performance hit for blobs. But if it's ok will do that as a separate PR.

peterbroadhurst: Added in https://github.com/hyperledger/firefly/pull/599/commits/17b9f907be4877ef14e7384e4bc4a1aa1a34a6e9 in #599

PR #598, opened 2022-03-11
Use multiple value insert for pins, and correct multi-row insert enablement for PSQL

peterbroadhurst:


peterbroadhurst:
going to collapse this into #599

PR #599, opened 2022-03-13
Performance enhancements from flame graph analysis

peterbroadhurst:
Set of incremental test+fix changes, based on analysis of flame graphs against runs.

- https://github.com/hyperledger/firefly/pull/599/commits/fce2ea693d709afae52477c195d4a84f36bb8f05
  - Use multiple value insert for pins.
- https://github.com/hyperledger/firefly/pull/599/commits/041005cc257c9d2023181037e3f1a7ed35797388
  - Increase the message writer timeout, accepting 50ms latency vs. benefits of more writes per commit
- https://github.com/hyperledger/firefly/pull/599/commits/4e0a86d8e584b958ff25d9ac25d19d47899b81cd
  - Delay individual events for up to 250ms when read-ahead is enabled, to reduce polling overhead
- https://github.com/hyperledger/firefly/pull/599/commits/5c4d55b990ca448ad393627d77146fe4faebb28f
    - Better debug logging for cache misses
- https://github.com/hyperledger/firefly/pull/599/commits/364e8ae6c37b9f423485c2c9059213d6de395600
  - Only query id+sequence when reading pages in batch manager. The full message is almost always in the cache when we're assembling a batch, so there's no need to query the full message+datarefs from the database.
- https://github.com/hyperledger/firefly/pull/599/commits/33c244c8313fe4425ce0a02400368fef298860be
  - Prevent new msg cache overwritting batch cache of message. We have to write the cache first before the DB update, as we have no idea how quickly the batch assembly will happen - and we can overwrite its updates if we do it after
- https://github.com/hyperledger/firefly/pull/599/commits/a0232b72fada999a2b0a90b9e0386b71402b2838
  - Better debug for cases when we have blocking due to missed pins
- https://github.com/hyperledger/firefly/pull/599/commits/8d6be5560db75b12cd1e11fc569d831a68a65079
  - Update message sequence from query of just the IDs+sequences
- https://github.com/hyperledger/firefly/pull/599/commits/8f63fc9185604773c6d973078f2b4c16e64d47eb
  - Add a minimum poll time to the batch manager, so it doesn't thrash rewinding as commits come in
- https://github.com/hyperledger/firefly/pull/599/commits/21789a0a207f7772033cdd43fcf5bc515cb395f9
  - The log formatter we have (which is great for other reasons) does a regex on every message, if you don't have a `prefix` key - so set one to the node name
- https://github.com/hyperledger/firefly/pull/599/commits/ff877bd614854cf7624b8c602983429d48b30d29
  - Cache first nodes for each parent org to avoid DB queries on every message
- https://github.com/hyperledger/firefly/pull/599/commits/d1dea98dfdf0aa5f4ea7cfa5be7737d3ea389888
  - Significant changes to how batch persistence works, to optimize the DB queries. Means that under normal circumstances, the sender only writes once (in batch assembly), and the receiver only writes once (in batch persist).
- https://github.com/hyperledger/firefly/pull/599/commits/edfaede213a7b384a4fac5ed8683285de5689c12
  - One DB operation to update all the pins within a batch, using a list of Batch+ID combinations
- https://github.com/hyperledger/firefly/pull/599/commits/7f1fb77c071393e7d04e60b9c0c591bd62c6d9a7
  - Enhancement to include the nonce in the `pins` array on a message (reducing the max topics for each message to 10), to aid debugging in the case of stuck contexts
- https://github.com/hyperledger/firefly/pull/599/commits/0d5b479d1d79d643536c4a918f3acb0d45726fdb
  - Commit offset updates to the DB asyncronously from the polling loop
  - Aggregator batch size to match input batch size
  - Fix descending sorting by default on created time to match confirmed descending sorting (which takes precedenced)
- https://github.com/hyperledger/firefly/pull/599/commits/80ed63861da18c4c9ca87d6433859220b0d6c9e5
  - Write back to the DB when the broadcast dispatch adds public references to blobs

Instead we now just query the ID+Sequence, and then read the rest from the cache

### DIFF
--- a/ff-perf.log
+++ b/ff-perf.log
@@ -1,1 +1,0 @@
-nohup: ff-perf: No such file or directory
--- a/ff-perf.log
+++ b/ff-perf.log
@@ -1,0 +1,1 @@
+nohup: ff-perf: No such file or directory
--- a/pkg/database/plugin.go
+++ b/pkg/database/plugin.go
@@ -96,6 +96,9 @@ type iMessageCollection interface {
 	// GetMessages - List messages, reverse sorted (newest first) by Confirmed then Created, with pagination, and simple must filters
 	GetMessages(ctx context.Context, filter Filter) (message []*fftypes.Message, res *FilterResult, err error)
 
+	// GetMessageIDs - Retrieves messages, but only querying the messages ID (no other fields)
+	GetMessageIDs(ctx context.Context, filter Filter) (ids []*fftypes.IDAndSequence, err error)
+
 	// GetMessagesForData - List messages where there is a data reference to the specified ID
 	GetMessagesForData(ctx context.Context, dataID *fftypes.UUID, filter Filter) (message []*fftypes.Message, res *FilterResult, err error)
 }
--- a/pkg/database/plugin.go
+++ b/pkg/database/plugin.go
@@ -185,6 +185,9 @@ type iOffsetCollection interface {
 }
 
 type iPinCollection interface {
+	// InsertPins - Inserts a list of pins - fails if they already exist, so caller can fall back to UpsertPin individually
+	InsertPins(ctx context.Context, pins []*fftypes.Pin) (err error)
+
 	// UpsertPin - Will insert a pin at the end of the sequence, unless the batch+hash+index sequence already exists
 	UpsertPin(ctx context.Context, parked *fftypes.Pin) (err error)
 
--- a/pkg/database/plugin.go
+++ b/pkg/database/plugin.go
@@ -77,7 +77,7 @@ type iMessageCollection interface {
 	//                 must match the hash of the record that is being inserted.
 	UpsertMessage(ctx context.Context, message *fftypes.Message, optimization UpsertOptimization) (err error)
 
-	// InsertMessages performs a batch insert of messages assured to be new records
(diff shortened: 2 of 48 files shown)


peterbroadhurst:
I do not believe the WS Heartbeat code coverage drop is related to my changes:
image

pkg/database/plugin.go
```diff
@@ -185,6 +188,9 @@ type iOffsetCollection interface {
 }
 
 type iPinCollection interface {
+	// InsertPins - Inserts a list of pins - fails if they already exist, so caller can fall back to UpsertPin individually
```
awrichar, line 191: > fails if they already exist, so caller can fall back to UpsertPin individually

This is a useful bit of information that (I think) is also relevant on some of the other multi-insert calls added recently.

ff-perf.log
```diff
@@ -0,0 +1 @@
+nohup: ff-perf: No such file or directory
```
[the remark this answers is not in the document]
peterbroadhurst: 🤦

pkg/database/plugin.go
peterbroadhurst: It wasn't before, but it is now after this PR 👍

PR #602, opened 2022-03-15
Remove route /messages/{msgid}/operations

awrichar:
Operations are strongly tied to transactions, but they really are not
tied to messages directly. This helper for retrieving a "third level"
object association probably creates more confusion than it's worth.

It seems cleaner to retrieve the message transaction, then retrieve
the transaction operations as separate queries.


peterbroadhurst (approved):
👍

PR #606, opened 2022-03-16
Add "info" field to token pools, consume "symbol" from connectors

awrichar:


internal/events/token_pool_created.go
```diff
@@ -28,17 +29,26 @@ import (
...
+func addPoolDetailsFromPlugin(ffPool *fftypes.TokenPool, pluginPool *tokens.TokenPool) error {
 	ffPool.Type = pluginPool.Type
 	ffPool.ProtocolID = pluginPool.ProtocolID
 	ffPool.Connector = pluginPool.Connector
 	ffPool.Standard = pluginPool.Standard
+	if pluginPool.Symbol != "" {
+		if ffPool.Symbol == "" {
+			ffPool.Symbol = pluginPool.Symbol
+		} else if ffPool.Symbol != pluginPool.Symbol {
+			return fmt.Errorf("received symbol '%s' does not match requested '%s'", pluginPool.Symbol, ffPool.Symbol)
```
nguyer, line 41: Since this is returned (and probably bubbled up to the user), should this be in the translations file? Also, maybe some wording that indicates that what was actually on the blockchain didn't match the API request. Rather than "received" vs "requested"?

nguyer (approved):

internal/events/token_pool_created.go
awrichar: All the processing in the `events` module is asynchronous on the event loop, so won't be easily visible to a user except in logs. I can tweak the wording though.

PR #611, opened 2022-03-18
Add E2E test support for ERC20/ERC721

awrichar:
Depends on https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/33


awrichar:
I attempted to add ERC20/ERC721 tokens to the E2E matrix for GitHub PRs - will see if it works and runs here. Open to whether this is needed or not (obviously we only want to run a subset of all combinations on every PR).

.github/workflows/go.yml
```diff
@@ -32,12 +32,17 @@ jobs:
...
         blockchain-provider: [geth, fabric]
+        token-provider: [none, erc1155, erc20_erc721]
         database-type: [sqlite3]
         exclude:
           - blockchain-provider: geth
             test-suite: TestFabricE2ESuite
           - blockchain-provider: fabric
             test-suite: TestEthereumE2ESuite
+          - blockchain-provider: fabric
+            token-provider: erc20_erc721
```
nguyer, line 43: Also need to exclude fabric and erc1155

nguyer (changes requested):
One minor thing on the GitHub action, but other than that, looks good

nguyer:
And there's a failing E2E test 😕

awrichar:
It's weird that Fabric + ERC1155 even executes (and passes?)... does it just ignore the token provider? I know it's definitely not actually using it.

And yea the failures in ERC20 are expected until the dependent PR noted is merged and tagged in the manifest.

awrichar:
Opened https://github.com/hyperledger/firefly-cli/issues/162 to track the fact that `-t` is silently ignored by Fabric stacks.

Issue #583, opened 2022-03-04
Add additional routes/query parameters for common queries on orgs/nodes/identities


awrichar:
In the wake of #549, identities (including orgs and nodes) have become much more flexible and powerful - but fetching identities, verifiers, and DIDs can require multiple lookups.

Some suggested enhancements to reduce API calls in a few spots:
- [x] `/network/nodes/{nid}` and `/network/organizations/{oid}` could be id-or-name lookups
- [ ] The above could possibly be extended to `/namespaces/{ns}/identities/{iid}` as well, but identities of different types could technically have overlapping names
- [ ] Lots of the "network" and "identities" routes could include a `fetchverifiers` bool in order to bring in an array of verifiers for each identity
- [x] New route to look up a DID from a DID-ID such as `did:firefly:org/org_1`

PR #617, opened 2022-03-18
Add route to lookup identity by DID

awrichar:
See #583

This is more of a starting point - the functionality works, but I'm pretty certain the API spelling is not what we want.

I was struggling with the best way to provide this functionality - it seems like it must be a query param, because `/` is a valid char for query params but not for path params. Assuming we do use a query param, it would also be nice to flag it as "required" from the Swagger perspective.

Side note: you can already query the `/namespaces/{ns}/identities` endpoint with a query param for `did`, but you have to know the namespace you're looking for in that case.


internal/apiserver/route_get_net_did_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/hyperledger/firefly/pkg/fftypes"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestGetIdentityByDID(t *testing.T) {
+	o, r := newTestAPIServer()
+	nmn := &networkmapmocks.Manager{}
+	o.On("NetworkMap").Return(nmn)
+	req := httptest.NewRequest("GET", "/api/v1/network/did?did=did:firefly:org/org_1", nil)
```
awrichar, line 33: See this looks weird 🙃

peterbroadhurst: Was reading https://w3c-ccg.github.io/did-resolution/#example-0

Wondering about:
`/api/v1/network/identities/did:firefly:org/org_1`

I think that could work. Given @awrichar is out on vacation, I might make this proposal in code for @nguyer to review.

peterbroadhurst: Ok - that's done @nguyer, including sorting the swagger generator so it can handle the custom Gorilla mux syntax.

peterbroadhurst:
Made a couple more enhancements:
- The DID resolution API uses the same cached lookup API that other code paths use in the codebase.
- Creation of a new identity can also use a name-or-UUID approach

internal/apiserver/route_get_net_did_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright © 2021 Kaleido, Inc.
```
nguyer, line 1: I'm surprised the linter didn't complain about the year here

nguyer (approved):

PR #633, opened 2022-03-27
Update batch manager dispatch to track inflight and fix private blobs

peterbroadhurst:
### Batch manager updates

I believe I've found an issue in the way the batch manager and the batch processors interact
Because the messages stay in ready state while they are being dispatched, the `readPage` loop will still see all messages that are dispatched

Currently it attempts to re-dispatch them when it's told to rewinds, and relies on the processor to de-dup those messages
It all works from a consistency perspective, but with the pattern of workload from the long-run it's a little inefficient in terms of how many rewinds and re-processing we do. It seems this can reach the point the batch manager can get jammed up trying to do those re-dispatches to the processor, while the processor is busy flushing what it's got

This PR simplifies the logic significantly, by preventing duplicate dispatches in the batch manager.

The manager keeps a map of all in-flight dispatched sequences to processors, and the processors call back to the manager when they have flushed a batch (so the messages will no longer turn up in `readPage` queries).

### Fix to private blobs

The private message batch dispatcher, was only sending the first blob in a batch. Meaning other messages with blobs in the batch would never be confirmed, because the blobs never arrived.


internal/batch/batch_manager.go
```diff
@@ -210,17 +213,56 @@ func (bm *batchManager) assembleMessageData(id *fftypes.UUID) (msg *fftypes.Mess
 	return msg, retData, nil
 }
 
-func (bm *batchManager) readPage() ([]*fftypes.IDAndSequence, error) {
-
-	// Pop out a rewind offset if there is one and it's behind the cursor
+// popRewind is called just before reading a page, to po out a rewind offset if there is one and it's behind the cursor
```
nguyer, line 216:
```suggestion
// popRewind is called just before reading a page, to pop out a rewind offset if there is one and it's behind the cursor
```

nguyer (approved):
I'm not super familiar with this part of the code, but this looks good to me. Also one minor spelling thing in a comment, but not a big deal.

PR #637, opened 2022-03-28
Update txType in private message schema

gabriel-indik:


gabriel-indik:
For some endpoints the Swagger is not dynamically generated, but instead it relies on a fixed schema. In the case of private messages, it looks like the data structure has been changed but the schema has not been updated accordingly. This PR updates the schema. Once the dynamic generation covers all endpoints, these updates won't be required.

peterbroadhurst (approved):
👍

PR #641, opened 2022-03-28
Add extra convenience functions for identities

### LINKED ISSUE #625: API enhancements for fetching signing keys
It would be great to be able to make a single API request to lookup identities in the FireFly network, and get their signing keys in the same response.

A use case for this is: I want to look up another member in the network, get their Ethereum address and transfer tokens to them. Right now, looking up the member, and then looking up their address is a two step process.

We should add a new query parameter `?fetchVerifiers` to add the list of verifiers to the identity object that is returned on the relevant endpoints.

Additionally, we should add a new endpoint for looking up a DID by a DID URL.

peterbroadhurst:
Resolves #625 

Adds a number of convenience functions:

- `GET` `/api/v1/network/identities` - global search (needed by @eberger727)
- `GET` `/api/v1/network/identities?fetchverifiers` - same with verifiers embedded
- `GET` `/api/v1/network/identities/did:firefly:org/1?fetchverifiers` - get by DID with verifiers
- `GET` `/api/v1/network/diddocs/did:firefly:org/1` - get DID Document by DID
- `GET` `/api/v1/namespaces/{ns}/identities?fetchverifiers` - get with namespace with verifiers
- `GET` `/api/v1/namespaces/{ns}/identities/{id}?fetchverifiers` - get by ID in namespace with verifiers
- `GET` `/api/v1/status` - adds an `org.verifiers` array

All of the convenience `verifiers` arrays above just have the `type` + `value` (not the `hash` or `identity` fields which are irrelevant when embedded).


internal/apiserver/route_get_net_identities_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright © 2021 Kaleido, Inc.
```
nguyer, line 1: 2022?

nguyer (approved):

peterbroadhurst:
Resolves #641

PR #642, opened 2022-03-29
Update our cache for event enrichment on message confirm

peterbroadhurst:
Noticed in the UI that confirmed messages on the Timeline view are showing as `Pending` - which is because they are using the cache rather than an explicit query against the DB. This PR updates the cache when messages are batched, or confirmed.


peterbroadhurst:
Also included a fix for this intermittent test failure:

```
--- FAIL: TestDownloadBatchE2EOk (0.00s)
    download_manager_test.go:98: PASS:	Name()
    download_manager_test.go:98: PASS:	DownloadData(string,string)
    download_manager_test.go:99: PASS:	InsertOperation(string,string,string)
    download_manager_test.go:99: FAIL:	ResolveOperation(string,string,fftypes.OpStatus,string,fftypes.JSONObject)
        		at: [download_manager_test.go:82]
    download_manager_test.go:99: FAIL: 1 out of 2 expectation(s) were met.
        	The code you are testing needs to make 1 more call(s).
        	at: [download_manager_test.go:99]
```

PR #643, opened 2022-03-29
Fix `transactions_id` index

peterbroadhurst:
Discovered we're doing full table scans for transactions when looking by ID, which is a performance critical query for the blockchain event processing routine (to update operations when we receive receipt notifications from EthConnect/FabConnect).

The reason was a simple typo in a migration.

```
postgres=# explain ANALYZE select * from transactions where id = '159b0bab-1a3f-406f-8862-bf8e05248789';
                                                  QUERY PLAN
---------------------------------------------------------------------------------------------------------------
 Seq Scan on transactions  (cost=0.00..4253.98 rows=1 width=116) (actual time=170.148..201.237 rows=1 loops=1)
   Filter: (id = '159b0bab-1a3f-406f-8862-bf8e05248789'::uuid)
   Rows Removed by Filter: 130390
 Planning Time: 0.126 ms
 Execution Time: 201.268 ms
(5 rows)
```


db/migrations/postgres/000005_create_transactions_table.up.sql
```diff
@@ -13,7 +13,6 @@ CREATE TABLE transactions (
   info        TEXT
 );
 
-CREATE UNIQUE INDEX transactions_id ON data(id);
```
awrichar, line 16: 😬

awrichar (approved):

PR #646, opened 2022-03-30
Add background workers to flush updates to operations

peterbroadhurst:
Updating an operation is quite an expensive set of DB calls when taken in isolation, and currently happens on the critical path of a number of event processing threads:

- Blockchain events - this is critical as it affects batch pin and custom contract event dispatch latency
- Token events - similar to blockchain events, with some extra complexity for event emission of operation events
- DX Transfer results - affects the notification of arrival of incoming events

So this PR proposes a dedicated set of workers (similar to we now have for message writing), that asynchronously flush operation updates to the DB in efficient batch transactions.

Some additional notes:
- Unlike message workers, we use a hash-bucket allocation of operation UUIDs to workers, so if there are retries etc. going on for a single operation we don't risk having two workers processing updates in parallel so the wrong one "wins" (a retry failure overtakes eventual success).
- After a recent change that means we're passing the ID into DX when we perform a send, this PR removes the logic that performed multiple retries to find an operation ID and was specific to DX (where originally DX generated the ID and passed it back to FF core to do the insert). Now DX is not "special", so if we find we need such logic in the future, we can put it into this logic by re-scheduling the update back to the worker as a retry (rather than discarding it with a warning as we do currently).


internal/orchestrator/bound_callbacks.go
```diff
@@ -30,22 +31,35 @@ type boundCallbacks struct {
...
+	return bc.om.SubmitOperationUpdate(&operations.OperationUpdate{
+		ID:             operationID,
+		State:          txState,
+		BlockchainTXID: blockchainTXID,
+		ErrorMessage:   errorMessage,
+		Output:         opOutput,
+	})
 }
 
 func (bc *boundCallbacks) TokenOpUpdate(plugin tokens.Plugin, operationID *fftypes.UUID, txState fftypes.OpStatus, blockchainTXID, errorMessage string, opOutput fftypes.JSONObject) error {
```
awrichar, line 47: Looks like the `plugin` parameter is now ignored/not needed?

internal/operations/operation_updater.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+	}
+	if tx != nil {
+		if err := ou.txHelper.AddBlockchainTX(ctx, tx, update.BlockchainTXID); err != nil {
+			return err
+		}
+	}
+
+	// Special handling for OpTypeTokenTransfer, which writes an event when it fails
+	if op.Type == fftypes.OpTypeTokenTransfer && update.State == fftypes.OpStatusFailed {
```
awrichar, line 248: Quite possibly these "special" cases should be farmed out to registered handlers as a new type of callback. Doesn't need to be done now, but I think we're a step closer to making it possible since the Operations Manager has a map of op types to handlers.

internal/orchestrator/bound_callbacks.go
peterbroadhurst: I've added it in for consistency with the tokens path 👍

internal/operations/operation_updater.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/hyperledger/firefly/internal/retry"
+	"github.com/hyperledger/firefly/internal/txcommon"
+	"github.com/hyperledger/firefly/pkg/database"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+// OperationUpdate is dispatched asynchronously to perform an update.
+type OperationUpdate struct {
+	ID             *fftypes.UUID
+	State          fftypes.OpStatus
```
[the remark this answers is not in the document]
peterbroadhurst: yeah... that was historical. Let me see if I can tidy it up everywhere.... there are quite a few places

internal/operations/operation_updater.go
peterbroadhurst: Agreed 👍 ... and deferring for now

internal/operations/manager.go
```diff
@@ -147,6 +155,35 @@ func (om *operationsManager) RetryOperation(ctx context.Context, ns string, opID
...
+		ErrorMessage:   update.Error,
+		Output:         update.Info,
+	}
+
+	// Pass manifest verification code to the background worker, for once it has loaded the operation
+	if opUpdate.VerifyManifest {
+		if update.Manifest != "" {
+			opUpdate.Manifest = update.Manifest
+		} else if update.Hash != "" {
+			opUpdate.Manifest = update.Hash
```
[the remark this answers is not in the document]
peterbroadhurst: This is good feedback on readability - going to go a step further and separate the vars.

PR #647, opened 2022-03-30
Use a single commit to generate all batch ops, before initiating the HTTP calls

peterbroadhurst:
- Use a single database transaction to write all the `pending` `operations` needed to dispatch a private messaging batch
- Then initiate all the HTTP calls to DX separately - noting these should be *pending* until the asynchronous callback comes

### DIFF
--- a/internal/privatemessaging/privatemessaging.go
+++ b/internal/privatemessaging/privatemessaging.go
@@ -18,6 +18,7 @@ package privatemessaging
 
 import (
 	"context"
+	"sync"
 
 	"github.com/hyperledger/firefly/internal/batch"
 	"github.com/hyperledger/firefly/internal/batchpin"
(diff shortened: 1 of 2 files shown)


internal/privatemessaging/privatemessaging.go
```diff
@@ -190,39 +197,71 @@ func (pm *privateMessaging) dispatchBatchCommon(ctx context.Context, state *batc
...
+					pm.exchange,
+					d.Namespace,
+					txid,
+					fftypes.OpTypeDataExchangeSendBlob)
+				addTransferBlobInputs(op, node.ID, blob.Hash)
+				if err = pm.operations.AddOrReuseOperation(ctx, op); err != nil {
+					return err
+				}
+
+				// Check the operation isn't already successful from a previous run
```
awrichar, line 230: Is there a check for previous success..?

awrichar (approved):

internal/privatemessaging/privatemessaging.go
peterbroadhurst: No :)

PR #650, opened 2022-03-30
Assign nonces more efficiently, with minimal DB ops

peterbroadhurst:
This PR addresses two problems:

1) Currently when we are flushing a batch, we perform a DB read+update individually. This is very inefficient when you have a large number of messages for the same topic+group combination in the same batch. Instead we can keep a track in memory of the nonce increments, and flush them one time at the end of the dispatch.

2) I found while investigating (1) that we were not including the `author` of the message in the calculation of which Nonce we assigned from the DB. This is a bug. e.g. if you sent a message on `topicA` in a group with both `bob` and `sally` using author `bob`, then sent another message _from the same node_ as `sally` on the same `topicA` - then `sally` would get the wrong nonce.

As per the comments, some consideration has been made to ensure we take allocation of nonces seriously. Two key scenarios:

- Double assigning a nonce to a message: If we crashed after allocating a nonce in sending a batch, then after restart included the same messages back into a _new_ batch then we need to **re-use the same nonce** rather than allocating a new one.
- Failing to assign a nonce due to DB retry: If we span round the `sealBatch` logic once, then got a DB error and retried, the nonces we notionally assigned to the messages would not have been flushed to disk (we flush the nonces to the DB right at the end of this logic, and all DB plugins today have atomic TXs). This means when we retry we need to do the allocation again.

The above scenarios is why we check that the `msg.Pins` array hasn't been assigned, and only assign it after we've exited the retry loop.

### DIFF
--- a/db/migrations/postgres/000078_add_nonce_author.down.sql
+++ b/db/migrations/postgres/000078_add_nonce_author.down.sql
@@ -5,7 +5,6 @@ DROP INDEX nonces_hash;
 ALTER TABLE nonces RENAME COLUMN "hash" TO "context";
 ALTER TABLE nonces ADD COLUMN group_hash CHAR(64);
 ALTER TABLE nonces ADD COLUMN topic VARCHAR(64);
-ALTER TABLE nonces ADD COLUMN
 
 CREATE INDEX nonces_context ON nonces(context);
 CREATE INDEX nonces_group ON nonces(group_hash);
--- a/db/migrations/postgres/000078_add_nonce_author.down.sql
+++ b/db/migrations/postgres/000078_add_nonce_author.down.sql
@@ -1,0 +1,13 @@
+BEGIN;
+
+DROP INDEX nonces_hash;
+
+ALTER TABLE nonces RENAME COLUMN "hash" TO "context";
+ALTER TABLE nonces ADD COLUMN group_hash CHAR(64);
+ALTER TABLE nonces ADD COLUMN topic VARCHAR(64);
+ALTER TABLE nonces ADD COLUMN
+
+CREATE INDEX nonces_context ON nonces(context);
+CREATE INDEX nonces_group ON nonces(group_hash);
+
+COMMIT;
--- a/internal/batch/batch_processor.go
+++ b/internal/batch/batch_processor.go
@@ -94,10 +94,12 @@ type batchProcessor struct {
 }
 
 type DispatchState struct {
-	Persisted fftypes.BatchPersisted
-	Messages  []*fftypes.Message
-	Data      fftypes.DataArray
-	Pins      []*fftypes.Bytes32
+	Persisted      fftypes.BatchPersisted
+	Messages       []*fftypes.Message
+	Data           fftypes.DataArray
+	Pins           []*fftypes.Bytes32
+	noncesAssigned map[fftypes.Bytes32]int64
+	msgPins        map[fftypes.UUID]fftypes.FFStringArray
 }
 
 const batchSizeEstimateBase = int64(512)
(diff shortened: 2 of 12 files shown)


db/migrations/postgres/000078_add_nonce_author.down.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+DROP INDEX nonces_hash;
+
+ALTER TABLE nonces RENAME COLUMN "hash" TO "context";
+ALTER TABLE nonces ADD COLUMN group_hash CHAR(64);
+ALTER TABLE nonces ADD COLUMN topic VARCHAR(64);
+ALTER TABLE nonces ADD COLUMN
```
awrichar, line 8: Extra line here that needs to be removed?

internal/batch/batch_processor.go
```diff
@@ -411,76 +445,106 @@ func (bp *batchProcessor) maskContext(ctx context.Context, msg *fftypes.Message,
...
-	return contextsOrPins, nil
+
+	return nil
 }
 
 func (bp *batchProcessor) sealBatch(state *DispatchState) (err error) {
 	err = bp.retry.Do(bp.ctx, "batch persist", func(attempt int) (retry bool, err error) {
 		return true, bp.database.RunAsGroup(bp.ctx, func(ctx context.Context) (err error) {
+			state.noncesAssigned = make(map[fftypes.Bytes32]int64)
+			state.msgPins = make(map[fftypes.UUID]fftypes.FFStringArray)
```
awrichar, line 543: Are these specifically (re)initialized at this moment (rather than in `initFlushState` with the rest of the struct) in order to clear state upon retries? Might be worth a comment.

peterbroadhurst: Yes - so we re-read Nonces from the DB. I'll add a comment

internal/batch/batch_processor.go
```diff
@@ -411,76 +445,108 @@ func (bp *batchProcessor) maskContext(ctx context.Context, msg *fftypes.Message,
...
+		return "", nil, err
+	}
+
+	// Now we have the nonce, add that at the end of the hash to make it unqiue to this message
 	nonceBytes := make([]byte, 8)
-	binary.BigEndian.PutUint64(nonceBytes, uint64(gc.Nonce))
+	binary.BigEndian.PutUint64(nonceBytes, uint64(nonce))
 	hashBuilder.Write(nonceBytes)
 
 	pin := fftypes.HashResult(hashBuilder)
```
awrichar, line 466: Isn't this recomputing the same value from line 455?

internal/batch/batch_processor.go
```diff
@@ -411,76 +445,108 @@ func (bp *batchProcessor) maskContext(ctx context.Context, msg *fftypes.Message,
...
+			// Clear state from any previous retry. We need to do fresh queries against the DB for nonces.
+			state.noncesAssigned = make(map[fftypes.Bytes32]int64)
+			state.msgPins = make(map[fftypes.UUID]fftypes.FFStringArray)
+
 			if bp.conf.txType == fftypes.TransactionTypeBatchPin {
 				// Generate a new Transaction, which will be used to record status of the associated transaction as it happens
-				if state.Pins, err = bp.maskContexts(ctx, state.Messages); err != nil {
+				if state.Pins, err = bp.maskContexts(ctx, state); err != nil {
 					return err
 				}
```
awrichar, line 551: Might consider calling `bp.flushNonceState()` separately here, instead of relying on `bp.maskContexts()` to chain into it.

awrichar (approved):

internal/batch/batch_processor.go
peterbroadhurst: Line 455 is the state of the hash after we've written:
- The topic
- The group
- The author

Then afterwards is generating a big endian 8byte hex value for the nonce (which we've just determined from the above) and adding it to the end of the hash.

Example here:
https://github.com/hyperledger/firefly/blob/25a2b21893c5d32c117977db4ea1bb52aa3294ad/internal/batch/batch_manager_test.go#L189

Issue #631, opened 2022-03-26
Tokens dashboard histogram API fails with `FF10115: Database query failed: no such column: timestamp`


peterbroadhurst:
<img width="1369" alt="image" src="https://user-images.githubusercontent.com/6660217/160222085-98af0e8b-652b-4221-b513-5ffbba4da855.png">

```
http://127.0.0.1:5000/api/v1/namespaces/default/charts/histogram/tokentransfers?startTime=1648176782&endTime=1648263182&buckets=15
```

PR #651, opened 2022-03-30
[charts-config] chart buckets are now capped at a configurable max

eberger727:
Previously, histogram buckets were created with a `CASE` query. This lead to performance issues when a large amount of data was queried from the database. 

This new method sets a max amount of rows to be queried in each bucket, with a default of 100 rows. 

The data structure and expected response is unchanged.

### DIFF
--- a/internal/database/sqlcommon/chart_sql.go
+++ b/internal/database/sqlcommon/chart_sql.go
@@ -19,53 +19,14 @@ package sqlcommon
 import (
 	"context"
 	"database/sql"
-	"fmt"
 	"strconv"
 
 	sq "github.com/Masterminds/squirrel"
+	"github.com/hyperledger/firefly/internal/config"
 	"github.com/hyperledger/firefly/internal/i18n"
 	"github.com/hyperledger/firefly/pkg/database"
 	"github.com/hyperledger/firefly/pkg/fftypes"
 )
-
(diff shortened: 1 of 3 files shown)


eberger727:
#631

internal/database/sqlcommon/chart_sql.go
```diff
@@ -86,174 +47,116 @@ func (s *SQLCommon) getTableNameFromCollection(ctx context.Context, collection d
...
+			Limit(uint64(config.GetInt(config.DatabaseMaxChartRows))))
 	}
-	rows.Close()
 
-	return dataTypes, nil
+	return queries
 }
 
-func (s *SQLCommon) getBucketTotal(typeBuckets []*fftypes.ChartHistogramType) (string, error) {
+func (s *SQLCommon) histogramResult(ctx context.Context, tableName string, rows *sql.Rows, onlyTimestamp bool) (map[string]int, string, error) {
```
nguyer, line 67: Is there a reason this returns `map[string]int, string, error` and not `map[string]int, int, error`? It seems like it's always empty or a number

nguyer (approved):
Looks good, with one minor question

internal/database/sqlcommon/chart_sql.go
eberger727: The API expects strings and not integers in the response, so this function returns a string for the `total` value so that it doesn't have to be converted in `GetChartHistogram()`

PR #655, opened 2022-03-31
[chart-test-fix] chart testing fix

dechdev:
Signed-off-by: David Echelberger <eberger727@gmail.com>


nguyer (changes requested):
Rather than listing out each possible order of the array (only 2 here), I would recommend using something like https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch to do a comparison ignoring the order.

PR #656, opened 2022-03-31
Move DX events interface to async with callback

peterbroadhurst:
Update FireFly Core to exploit the parallelism enabled by https://github.com/hyperledger/firefly-dataexchange-https/pull/56

- Change spelling of Ack on callback interface to DX from `commit` to `ack`
  - This means there is a migration dependency on the DX-HTTPS `v0.11.0` release (or other DX microservice implementations to upgrade to support the new API)
- Refactor Data Exchange callback interface for asynchronous confirmation via callback
- Blanket `BLOB` -> `Blob` rename
- Add multi-insert support to Blobs in database layer
- Add `blobreceiver` workers/config for parallel batched insertion of blobs to the DB
- Dispatch `MessageReceived` events to insert batches as go routines
- Add retry to `SubmitOperationUpdate` and remove error return
- Updated manifest to pick up EthConnect `v3.1.6`


pkg/dataexchange/plugin.go
```diff
@@ -42,11 +42,11 @@ import (
 // - Must drive events on the target node that contain the input data
 // - No requirement to retain the data beyond the confirmation of receipt of the event at the target
 //
-// BLOBS
+// BlobS
```
awrichar, line 45: `BlobS`

internal/events/blob_receiver.go
```diff
@@ -0,0 +1,282 @@
...
+			if err != nil {
+				log.L(ctx).Debugf("Blob receiver worker exiting: %s", err)
+				return
+			}
+			batch = nil
+		}
+	}
+}
+
+func (br *blobReceiver) handleBlobNoficiationsRetry(ctx context.Context, notifications []*blobNotification) error {
```
awrichar, line 164: `Noficiations`

awrichar (approved):
Looks good to me!

PR #657, opened 2022-04-01
Avoid blocking event processing threads when queuing rewinds

peterbroadhurst:
In PR chain with #656


awrichar (approved):
Looks like only one additional commit on top of #656, and that looks good too.

PR #658, opened 2022-04-01
Refactor OpenAPI Spec generation

nguyer:
This PR includes some significant changes to the way we automatically generate the OpenAPI Spec in FireFly.

## Input masking
- The old InputMask is totally gone. We now use tags on individual fields to indicate how to build the correct JSON Schema for a field.
- `ffexclude:"true"` can be used to always exclude a field from JSON Schema
- `ffexclude:"routeName,otherRouteName"` can be used to exclude a field from JSON Schema on routes named `routeName` or `otherRouteName`. It is case-sensitive and comma separated
- `ffexcludeinput:"true"` can be used to to always exclude a field from the **input** JSON Schema
- `ffexcludeinput:"routeName,otherRouteName"` can be used to exclude a field from the **input** JSON Schema on routes named `routeName` or `otherRouteName`. It is case-sensitive and comma separated

## Field level descriptions

Field level descriptions can be added in `internal/i18n/en_translations_descriptions.go`. Each struct field that wants to use one of these needs to have an ffstruct tag on it, indicating the name of the struct. That will be combined with the JSON field name (note, it is not the GO field name, but the JSON serialized name), separated by a "." This is the key used to lookup the translation below. If it is not found, the description is left blank in the OpenAPI spec.

Example: 

```go
// message.go
type Message struct {
	Header    MessageHeader `ffstruct:"Message" json:"header"`
```

```go
// en_translations_descriptions.go
MessageHeader    = ffm("Message.header", "The message header")
```

## Other notes, or points worthy of discussion / further exploration:

- I have removed all "handcrafted" JSON schemas. I did not go so far as to remove the input and output schema fields from the Route struct, because the FFI autogeneration still uses them, as it basically just passes through the JSON Schema from the FFI. I may have broken this at the moment though - need to check that.
- It looks like the CSS is "weird" on the Swagger UI. The `description` field doesn't appear to be formatted consistently depending on what level the field is nested
- @awrichar requested the ability to mark fields as `required` in the JSON Schema. I think this is a great idea, but still need to work out how to add it, because the `required` field is a list of field names that go onto the _parent_ JSON Schema, which we don't currently have a pointer to at the time of field level customization.

## Updated kin-openapi library
Right now this PR will require you to clone the `master` branch of https://github.com/getkin/kin-openapi, as there was a change just merged today that it needs, and it hasn't had a new release yet. Just clone it into a directory that sits next to your `firefly` directory and it should be good to go.


internal/apiserver/route_admin_put_config_record.go
```diff
@@ -60,7 +37,6 @@ var putConfigRecord = &oapispec.Route{
 	JSONInputValue:  func() interface{} { return fftypes.JSONAnyPtr("{}") },
```
awrichar, line 37: So this will now not have a well-defined JSON schema in OpenAPI?

internal/database/sqlcommon/group_sql.go
```diff
@@ -32,7 +32,6 @@ var (
 		"message_id",
 		"namespace",
 		"name",
-		"ledger",
```
awrichar, line 35: Should this be removed from the SQL table definition too?

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	DatatypeID        = ffm("Datatype.id", "The UUID of the datatype")
+	DatatypeMessage   = ffm("Datatype.message", "The UUID of the broadcast message that was used to publish this datatype to the network")
+	DatatypeValidator = ffm("Datatype.validator", "The validator that should be used to verify this datatype")
+	DatatypeNamespace = ffm("Datatype.namespace", "The namespace of the datatype. Data resources can only be created referencing datatypes in the same namespace")
+	DatatypeName      = ffm("Datatype.name", "The name of the datatype")
+	DatatypeVersion   = ffm("Datatype.version", "The version of the datatype. Multiple versions can exist with the same name. Use of semantic versioning is encourages, such as v1.0.1")
+	DatatypeHash      = ffm("Datatype.hash", "The hash of the value, such as the JSON schema. Allows all parties to be confident they have the exact same rules for verifying data created against a datatype")
+	DatatypeCreated   = ffm("Datatype.created", "The time the datatype was created")
+	DatatypeValue     = ffm("Datatype.value", "The definition of the datatype, in the syntax supported by the validator. Such as a JSON Schema definition")
```
awrichar, line 119: I would parenthesize `(such as a JSON Schema definition)`

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	OperationCreated     = ffm("Operation.created", "The time the operation was created")
+	OperationUpdated     = ffm("Operation.updated", "The last update time of the operation")
+	OperationRetry       = ffm("Operation.retry", "If this operation was initiated as a retry to a previous operation, this field points to the UUID of the operation being retried")
+
+	// BlockchainEvent field descriptions
+	BlockchainEventID         = ffm("BlockchainEvent.id", "The UUID assigned to the event by FireFly")
+	BlockchainEventSource     = ffm("BlockchainEvent.source", "The blockchain plugin or token service that detected the event")
+	BlockchainEventNamespace  = ffm("BlockchainEvent.namespace", "The namespace of the listener that detected this blockchain event")
+	BlockchainEventName       = ffm("BlockchainEvent.name", "The name of the event in the blockchain smart contract")
+	BlockchainEventListener   = ffm("BlockchainEvent.listener", "The UUID of the listener that detected this event, or nil for the built-in BatchPin events in the system namespace")
```
awrichar, line 176: It will be nil for all built-in events (ie token events as well, not just BatchPin)

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	BlockchainEventProtocolID = ffm("BlockchainEvent.protocolId", "A alphanumerically sortable string that represents this event uniquely on the blockchain. Convention for plugins is zero-padded values BLOCKNUMBER/TXN_INDEX/EVENT_INDEX")
+	BlockchainEventOutput     = ffm("BlockchainEvent.output", "The data output by the event, parsed to JSON according to the interface of the smart contract")
+	BlockchainEventInfo       = ffm("BlockchainEvent.info", "Detailed blockchain specific information about the event, as generated by the blockchain connector")
+	BlockchainEventTimestamp  = ffm("BlockchainEvent.timestamp", "The time allocated to this event by the blockchain. This is the block timestamp for most blockchain connectors")
+	BlockchainEventTX         = ffm("BlockchainEvent.tx", "If this blockchain event is coorelated to FireFly transaction such as a FireFly submitted token transfer, this field is set to the UUID of the FireFly transaction")
+
+	// ChartHistogramBucket field descriptions
+	ChartHistogramBucketCount     = ffm("ChartHistogramBucket.count", "Total count of entires in this time bucket within the histogram")
```
awrichar, line 184: `entries`

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	ChartHistogramBucketTimestamp = ffm("ChartHistogramBucket.timestamp", "Starting timestamp for the bucket")
+	ChartHistogramBucketTypes     = ffm("ChartHistogramBucket.types", "Array of separate counts for individual types of record within the bucket")
+
+	// ChartHistogramType field descriptions
+	ChartHistogramTypeCount = ffm("ChartHistogramType.count", "Count of entires of a given type within a bucket")
```
awrichar, line 189: `entries`

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+
+	// SubscriptionBlockchainEventFilter field descriptions
+	SubscriptionBlockchainEventFilterName     = ffm("SubscriptionBlockchainEventFilter.name", "Regular expression to apply to the blockchain event 'name' field, which is the name of the event in the underlying blockchain smart contract")
+	SubscriptionBlockchainEventFilterListener = ffm("SubscriptionBlockchainEventFilter.listener", "Regular expression to apply to the blockchain event 'listener' field, which is the UUID of the event listener. So you can restrict your subscription to certain blockchain listeners. Alternatively to avoid your application need to know listener UUIDs you can set the 'topic' field of blockchain event listeners, and use a topic filter on your subscriptions")
+
+	// SubscriptionCoreOptions field descriptions
+	SubscriptionCoreOptionsFirstEvent = ffm("SubscriptionCoreOptions.firstEvent", "Whether your appplication would like to receive events from the 'earliest' event emitted by your FireFly node (from the beginning of time), the 'latest' event (from now), or a specific event sequence. Default is 'latest'")
```
awrichar, line 445: Are the semantics for earliest/latest/newest/oldest _different_ here than for contract listeners?

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	SubscriptionCoreOptionsReadAhead  = ffm("SubscriptionCoreOptions.readAhead", "The number of events to stream ahead to your application, while waiting for confirmation of consumption of those events. At least once delivery semantics are used in FireFly, so if your application crashes/reconnects this is the maximum number of events you would expect to be redelivered after it restarts")
+	SubscriptionCoreOptionsWithData   = ffm("SubscriptionCoreOptions.withData", "Whether message events delivered over the subscription, should be packaged with the full data of those messages in-line as part of the event JSON payload. Or if the application should make separate REST calls to download that data")
```
awrichar, line 447: Could note that this is only supported for some transports (I know it's _not_ supported for websockets)

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	TokenApprovalNamespace       = ffm("TokenApproval.namespace", "The namespace for the approval, which must match the namespace of the token pool")
+	TokenApprovalProtocolID      = ffm("TokenApproval.protocolId", "The protocol identifier of the blockchain event for the token approval. See the 'protocolId' field on blockchain events for more information")
+	TokenApprovalCreated         = ffm("TokenApproval.created", "The creation time of the token approval")
+	TokenApprovalTX              = ffm("TokenApproval.tx", "If submitted via FireFly through the local node, this will reference the UUID of the FireFly transaction")
+	TokenApprovalBlockchainEvent = ffm("TokenApproval.blockchainEvent", "The UUID of the blockchain event")
+	TokenApprovalConfig          = ffm("TokenApproval.config", "Input only field, with token connector specific configuration of the approval.  See your chosen token connector documentation for details")
+
+	// TokenApprovalInputInput field descriptions
+	TokenApprovalInputInputPool = ffm("TokenApprovalInput.pool", "The name or UUID of a token pool")
```
awrichar, line 466: Required if more than one pool exists.

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	SubscriptionCoreOptionsWithData   = ffm("SubscriptionCoreOptions.withData", "Whether message events delivered over the subscription, should be packaged with the full data of those messages in-line as part of the event JSON payload. Or if the application should make separate REST calls to download that data")
+
+	// TokenApproval field descriptions
+	TokenApprovalLocalID         = ffm("TokenApproval.localId", "The UUID of this token approval, in the local FireFly node")
+	TokenApprovalPool            = ffm("TokenApproval.pool", "The UUID the token pool this approval applies to")
+	TokenApprovalTokenIndex      = ffm("TokenApproval.tokenIndex", "The index of the token within the pool that this approval applies to")
+	TokenApprovalConnector       = ffm("TokenApproval.connector", "The token connector. Required on input when there are more than one token connectors configured")
```
awrichar, line 453: Might be worth calling out "the name of the token connector, as specified in the FireFly core configuration file"

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	// TokenBalance field descriptions
+	TokenConnectorName = ffm("TokenConnector.name", "The name of the token connector, as configured in the FireFly core configuration file")
+
+	// TokenPool field descriptions
+	TokenPoolID         = ffm("TokenPool.id", "The UUID of the token pool")
+	TokenPoolType       = ffm("TokenPool.type", "The type of token the pool contains, such as fungible/non-fungible")
+	TokenPoolNamespace  = ffm("TokenPool.namespace", "The namespace for the token pool")
+	TokenPoolName       = ffm("TokenPool.name", "The name of the token pool. Note the name is not validated against the description of the token on the blockchain")
+	TokenPoolStandard   = ffm("TokenPool.standard", "The ERC standard the token pool conforms to, as reported by the token connector")
+	TokenPoolProtocolID = ffm("TokenPool.protocolId", "The protocol identifier of the blockchain event that established the existence of the pool on-chain. See the 'protocolId' field on blockchain events for more information")
```
awrichar, line 487: This is actually _not_ the same as the blockchain protocol ID. Format is totally up to the connector. Suggestion: `An identifier for the pool, as provided by the token connector.`

For the record, erc1155 uses something like "F1" and erc20 uses something like "address=0x123&schema=erc20" (not sure if examples are worth calling out here though).

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	TokenTransferNamespace       = ffm("TokenTransfer.namespace", "The namespace for the transfer, which must match the namespace of the token pool")
+	TokenTransferKey             = ffm("TokenTransfer.key", "The blockchain signing key for the transfer. On input defaults to the first signing key of the organization that operates the node")
+	TokenTransferFrom            = ffm("TokenTransfer.from", "The source account for the transfer. On input defaults to the value of 'key'")
+	TokenTransferTo              = ffm("TokenTransfer.to", "The target account for the transfer. Required on input")
+	TokenTransferAmount          = ffm("TokenTransfer.amount", "The amount for the transfer. For non-fungible tokens will always be 1. For fungible tokens, the number of decimals for the token pool should be considered when inputting the amount. For example, with 18 decimals a fractional balance of 10.234 will be specified as 10,234,000,000,000,000,000")
+	TokenTransferProtocolID      = ffm("TokenTransfer.protocolId", "The protocol identifier of the blockchain event for the token transfer. See the 'protocolId' field on blockchain events for more information")
```
awrichar, line 510: This one does happen to somewhat align with the formatting used for blockchain events, although I don't know that we should enforce it. Suggestion: `A unique identifier for this transfer, as provided by the token connector`

For the record, erc1155 uses `BLOCK/TX/EVENT/N` where the final segment is optional and is used only when exploding batch events into individual transfer events.

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	TokenTransferPool            = ffm("TokenTransfer.pool", "The UUID the token pool this transfer applies to")
+	TokenTransferTokenIndex      = ffm("TokenTransfer.tokenIndex", "The index of the token within the pool that this transfer applies to")
+	TokenTransferURI             = ffm("TokenTransfer.uri", "The URI of the token this transfer applies to")
+	TokenTransferConnector       = ffm("TokenTransfer.connector", "The token connector. Required on input when there are more than one token connectors configured")
+	TokenTransferNamespace       = ffm("TokenTransfer.namespace", "The namespace for the transfer, which must match the namespace of the token pool")
+	TokenTransferKey             = ffm("TokenTransfer.key", "The blockchain signing key for the transfer. On input defaults to the first signing key of the organization that operates the node")
+	TokenTransferFrom            = ffm("TokenTransfer.from", "The source account for the transfer. On input defaults to the value of 'key'")
+	TokenTransferTo              = ffm("TokenTransfer.to", "The target account for the transfer. Required on input")
```
awrichar, line 508: Not technically required - it's actually `On input defaults to the value of 'key'`

For minting in particular, this is useful so you'll mint tokens to yourself unless otherwise specified. It does mean that for transfers you must specify either "from" or "to", as they can't be the same. Not sure these additional caveats are necessary to call out though.

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	TokenApprovalOperator        = ffm("TokenApproval.operator", "The blockchain identity that is granted the approval")
+	TokenApprovalApproved        = ffm("TokenApproval.approved", "Whether this record grants permission for an operator to perform actions on the token balance (true), or revokes permission (false)")
+	TokenApprovalInfo            = ffm("TokenApproval.info", "Token connector specific information about the approval operation, such as whether it applied to a limited balance of a fungible token. See your chosen token connector documentation for details")
+	TokenApprovalNamespace       = ffm("TokenApproval.namespace", "The namespace for the approval, which must match the namespace of the token pool")
+	TokenApprovalProtocolID      = ffm("TokenApproval.protocolId", "The protocol identifier of the blockchain event for the token approval. See the 'protocolId' field on blockchain events for more information")
+	TokenApprovalCreated         = ffm("TokenApproval.created", "The creation time of the token approval")
+	TokenApprovalTX              = ffm("TokenApproval.tx", "If submitted via FireFly through the local node, this will reference the UUID of the FireFly transaction")
```
awrichar, line 461: This isn't specific to the local node... I think just `If submitted via FireFly, this will reference the UUID of the FireFly transaction`

Caveat that the token connector must support attaching data for this to be true... although there are some planned enhancements to that in the near term.

internal/i18n/en_translations_descriptions.go
```diff
@@ -0,0 +1,544 @@
...
+	TransactionStatusDetailsType      = ffm("TransactionStatusDetails.type", "The type of the transaction status detail record")
+	TransactionStatusDetailsSubType   = ffm("TransactionStatusDetails.subtype", "A sub-type, such as an operation type, or an event type")
+	TransactionStatusDetailsStatus    = ffm("TransactionStatusDetails.status", "The status of the detail record. Cases where an event is required for completion, but has not arrived yet are marked with a 'pending' record")
+	TransactionStatusDetailsTimestamp = ffm("TransactionStatusDetails.timestamp", "The time relevant to when the record was updated, such as the time an event was created, or the last update time of an operation")
+	TransactionStatusDetailsID        = ffm("TransactionStatusDetails.id", "The UUID of the entry referenced by this detail. The type of this record can be inferred from the entry type")
+	TransactionStatusDetailsError     = ffm("TransactionStatusDetails.error", "If an error occurred related to the detail entry, it is included here")
+	TransactionStatusDetailsInfo      = ffm("TransactionStatusDetails.info", "Input data for operations, and output data for events")
```
awrichar, line 532: I believe we only show outputs for all detail items (operations included)

pkg/fftypes/message.go
```diff
@@ -68,31 +68,31 @@ var (
...
 	SignerRef
-	Created   *FFTime       `json:"created,omitempty"`
-	Namespace string        `json:"namespace,omitempty"`
-	Group     *Bytes32      `json:"group,omitempty"`
-	Topics    FFStringArray `json:"topics,omitempty"`
-	Tag       string        `json:"tag,omitempty"`
-	DataHash  *Bytes32      `json:"datahash,omitempty"`
+	Created   *FFTime       `ffstruct:"MessageHeader" json:"created,omitempty"`
+	Namespace string        `ffstruct:"MessageHeader" json:"namespace,omitempty"`
+	Group     *Bytes32      `ffstruct:"MessageHeader" json:"group,omitempty" ffexclude:"postNewMessageBroadcast" ffrequired:"postNewMessagePrivate"`
```
awrichar, line 78: Is `ffrequired` implemented?

pkg/fftypes/tokentransfer.go
```diff
@@ -25,27 +25,27 @@ var (
...
+	Key             string            `ffstruct:"TokenTransfer" json:"key,omitempty"`
+	From            string            `ffstruct:"TokenTransfer" json:"from,omitempty" ffexcludeinput:"postTokenMint"`
+	To              string            `ffstruct:"TokenTransfer" json:"to,omitempty" ffexcludeinput:"postTokenBurn"`
+	Amount          FFBigInt          `ffstruct:"TokenTransfer" json:"amount"`
+	ProtocolID      string            `ffstruct:"TokenTransfer" json:"protocolId,omitempty" ffexcludeinput:"true"`
+	Message         *UUID             `ffstruct:"TokenTransfer" json:"message,omitempty"`
+	MessageHash     *Bytes32          `ffstruct:"TokenTransfer" json:"messageHash,omitempty" ffexcludeinput:"true"`
+	Created         *FFTime           `ffstruct:"TokenTransfer" json:"created,omitempty" ffexcludeinput:"true"`
+	TX              TransactionRef    `ffstruct:"TokenTransfer" json:"tx" ffexcludeinput:"true"`
+	BlockchainEvent *UUID             `ffstruct:"TokenTransfer" json:"blockchainEvent,omitempty"`
```
awrichar, line 44: This should be `ffexcludeinput`

internal/apiserver/route_admin_put_config_record.go
nguyer: We should probably get rid of this one

internal/i18n/en_translations_descriptions.go
nguyer: Yeah it should actually be `newest` or `oldest`. FireFly will pass through anything else so technically `latest` will also work.

pkg/fftypes/message.go
nguyer: No :(

internal/apiserver/route_admin_put_config_record.go
nguyer: The input schema for this one is any key/value mapping, so I think this is correct

awrichar:
Looks like 2 comments were left unresolved:
https://github.com/hyperledger/firefly/pull/658#discussion_r842969159
https://github.com/hyperledger/firefly/pull/658#discussion_r842977260

However I'll mark Approved and leave you to take one more pass on those if you're able.

awrichar (approved):

Issue #661, opened 2022-04-04
Synchronous token approval timeout


shorsher:
Currently, synchronous token approvals will time out when submitting an approval for a pool that already has an existing token approval. This is because the `localID` assigned to the sync request has the possibility to be overwritten with the ID of the existing token approval because token approvals are mutable.

Scenario:
1. A token pool `123` exists with an existing approval `X`
2. A new approval `Y` is submitted for pool `123` and assigned a local ID.
3. When FireFly receives the approval event for `Y`, it determines that there is an existing approval `X` with the same `connector`, `pool`, and `protocolID` attributes.
4. FireFly updates approval `X` with the attributes from `Y`.
5. Approval `Y`'s local ID is discarded, so the FireFly sync/async bridge is stuck waiting.

PR #662, opened 2022-04-04
Proper handling of token approval events

shorsher:
- Ensure existing approvals are not being overwritten
- Remove localID from async/sync POST response
- Commented out approvals from synchronous tokens e2e test due to https://github.com/hyperledger/firefly/issues/661

### DIFF
--- a/internal/database/sqlcommon/tokenapproval_sql.go
+++ b/internal/database/sqlcommon/tokenapproval_sql.go
@@ -187,6 +187,14 @@ func (s *SQLCommon) GetTokenApprovalByProtocolID(ctx context.Context, connector,
 	})
 }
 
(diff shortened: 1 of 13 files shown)


shorsher:
Coverage should be fixed, codecov bot isn't updating for some reason.

internal/database/sqlcommon/tokenapproval_sql.go
```diff
@@ -187,6 +187,14 @@ func (s *SQLCommon) GetTokenApprovalByProtocolID(ctx context.Context, connector,
 	})
 }
 
+func (s *SQLCommon) GetTokenApprovalByProtocolIDAndPoolID(ctx context.Context, connector, protocolID string, poolID *fftypes.UUID) (*fftypes.TokenApproval, error) {
```
awrichar, line 190: Naming thing, but I might suggest deleting `GetTokenApprovalByProtocolID`, calling this method here just `GetTokenApproval`, and calling the by-id method `GetTokenApprovalByID`. I think that's fairly consistent with naming elsewhere and avoids some really long names...

Issue #663, opened 2022-04-05
Timing condition can leave a transfer unconfirmed


peterbroadhurst:
The `TokensTransferred` message attempts to deal with either the message or the transfer arriving first.

[ReplaceMessage](https://github.com/hyperledger/firefly/blob/77e9c879a35ec89363f49c64d09dd453bc8c28f6/internal/events/tokens_transferred.go#L122-L162)

However, as can be seen from these log snippets of two concurrent DB transactions happening on the event aggregator and the token event loop concurrently, this logic is not safe:

### Token event loop

```
[2022-04-04T21:54:48.533Z] DEBUG node_0: Added message e7f04b6f-9ce2-4d22-9836-f465c0e4a0e6 sequence=5236245 to in-flight batch assembly 03749925-36a0-4177-91ba-787e9e99d505 d=pinned_broadcast p=default|did:firefly:org/org_0| role=batchmgr
[2022-04-04T21:54:52.416Z] DEBUG node_1: SQL-> begin dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.418Z] DEBUG node_1: SQL<- begin dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.418Z] DEBUG node_1: SQL-> query: SELECT type, local_id, dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.428Z] DEBUG node_1: SQL<- query dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.428Z] DEBUG node_1: Token transfer '000000287895/000001/000001' not found dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.428Z] DEBUG node_1: SQL-> query: SELECT id, namespace, dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.442Z] DEBUG node_1: SQL<- query dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.442Z] DEBUG node_1: SQL-> query: SELECT id, namespace, dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.445Z] DEBUG node_1: SQL<- query dbtx=MPBYNUow role=event-manager
[2022-04-04T21:54:52.445Z] DEBUG node_1: SQL-> query: SELECT id, ttype, dbtx=MPBYNUow role=event-manager
[...]

PR #665, opened 2022-04-05
Add rewinder to handle enrichment off event poller critical path

peterbroadhurst:
Issue #663 demonstrated that the event loops of the various plugins cannot safely ask the DB for information the event aggregator loop is updating, concurrently to the event poller writing it.

However, we don't want to slow the event poller down by making it do complex enrichment tasks to work out how to rewind. These tasks are as follows:
- For blobs: Query all `data` IDs associated with a `blob` hash, then find all `message` IDs that are associated with that data
- For token transfers: Query the `batch` ID associated with the message, which is associated with the on-chain transfer

So this PR introduces the architecture summarized in the diagram below. This ensures we do **not** start the enrichment DB activities (in loop 2) until the event poller has completed the cycle that might have been active while the plugin event that generated the rewind was active

pr_665_diagram_1


internal/database/sqlcommon/message_sql.go
```diff
@@ -498,6 +499,37 @@ func (s *SQLCommon) GetMessageIDs(ctx context.Context, filter database.Filter) (
 	return ids, nil
 }
 
+func (s *SQLCommon) GetMessagesBatchIDsForDataIDs(ctx context.Context, dataIDs []driver.Value) (batchIDs []*fftypes.UUID, err error) {
+	query := sq.Select("m.batch_id").From("messages_data AS md").LeftJoin("messages AS m ON m.id = md.message_id").Where(sq.Eq{"md.data_id": dataIDs})
```
awrichar, line 503: JOIN 👀 👀 👀

internal/events/aggregator_rewind.go
```diff
@@ -0,0 +1,284 @@
...
+func (rw *rewinder) start() {
+	go rw.rewindLoop1()
+	go rw.rewindLoop2()
+}
+
+// rewindLoop1 is responsible from taking requests to the aggregator for rewinds, from other
+// event loops that execute completely independently.
+// It takes them in as fast as possible, and puts them in a staging area ready for the aggregator
+// itself to mark it's finished processing anything in-flight
+func (rw *rewinder) rewindLoop1() {
```
awrichar, line 89: Would suggest more descriptive names than `rewindLoop1` and `rewindLoop2`... perhaps `receiveLoop` and `processLoop`?

internal/events/aggregator_rewind.go
```diff
@@ -0,0 +1,284 @@
...
+		readyRewinds:     make(map[fftypes.UUID]bool),
+	}
+}
+
+func (rw *rewinder) start() {
+	go rw.rewindLoop1()
+	go rw.rewindLoop2()
+}
+
+// rewindLoop1 is responsible from taking requests to the aggregator for rewinds, from other
```
awrichar, line 85: _for_ taking requests

internal/events/aggregator_rewind.go
```diff
@@ -0,0 +1,284 @@
...
+			// Pop when we've timed out
+			popCtx = timerCtx
+		} else {
+			// Only pop for new work, or shutdown
+			popCtx = rw.ctx
+		}
+		select {
+		case <-popCtx.Done():
+			if timerCtx == nil && timerCancel == nil {
+				log.L(rw.ctx).Debugf("Rewind loop 2 stopping")
```
awrichar, line 129: Should there just be separate `case` entries for `<-rw.ctx.Done()` and `<-timerCtx.Done()`? Seems like they're handled separately.

internal/events/aggregator_rewind.go
```diff
@@ -0,0 +1,284 @@
...
+				timerCtx, timerCancel = context.WithTimeout(rw.ctx, rw.minRewindTimeout)
+			}
+		}
+	}
+}
+
+func (rw *rewinder) processStagedRewinds() bool {
+	batchIDs := make(map[fftypes.UUID]bool)
+	var msgRewinds []*fftypes.UUID
+	var newBlobHashes []driver.Value
```
awrichar, line 153: any reason not to type this to `*fftypes.Bytes32`?

pkg/database/plugin.go
```diff
@@ -104,6 +105,12 @@ type iMessageCollection interface {
 
 	// GetMessagesForData - List messages where there is a data reference to the specified ID
 	GetMessagesForData(ctx context.Context, dataID *fftypes.UUID, filter Filter) (message []*fftypes.Message, res *FilterResult, err error)
+
+	// GetMessageBatchIDs - an optimized query to retrieve any non-null batch IDs for a list of message IDs
+	GetMessageBatchIDs(ctx context.Context, msgIDs []driver.Value) (batchIDs []*fftypes.UUID, err error)
```
awrichar, line 110: Can we not use `[]*fftypes.UUID` for the IDs?

internal/events/aggregator_test.go
```diff
@@ -36,12 +35,14 @@ import (
...
 	"github.com/hyperledger/firefly/pkg/database"
 	"github.com/hyperledger/firefly/pkg/fftypes"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
 func newTestAggregatorCommon(metrics bool) (*aggregator, func()) {
 	config.Reset()
+	logrus.SetLevel(logrus.DebugLevel)
```
awrichar, line 45: Is this needed here?

awrichar (approved):
Left some naming and typing suggestions, but functionally looks great. Nice to have all the rewind functionality consolidated now.

internal/database/sqlcommon/message_sql.go
```diff
@@ -498,6 +499,37 @@ func (s *SQLCommon) GetMessageIDs(ctx context.Context, filter database.Filter) (
...
+	return s.queryBatchIDs(ctx, query)
+}
+
+func (s *SQLCommon) GetMessageBatchIDs(ctx context.Context, msgIDs []driver.Value) (batchIDs []*fftypes.UUID, err error) {
```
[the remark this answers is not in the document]
peterbroadhurst: I'd put the `Message` first just to clarify why it's in the `message_sql` set of functions, but totally agree with the feedback. One tweak in what I changed it to:

- `GetBatchIDsForMessages` (as you suggested)
- `GetBatchIDsForDataAttachments` - added attachments to make it clear it's the messages

internal/events/aggregator_rewind.go
peterbroadhurst: Thanks for the suggestions 👍  ... had scratched my head and left these names in out lack of inspiration

internal/events/aggregator_rewind.go
peterbroadhurst: The problem is there isn't always a timer, so you get into complex logic where the select is dynamic. So this approach of having two options of which context to listen on, was the simplest thing I could find to do.

pkg/database/plugin.go
peterbroadhurst: It means an extra loop and array allocation - I don't mind doing it here as it's a component boundary interface.

internal/events/aggregator_rewind.go
peterbroadhurst: aha, it's valid in `sq.Eq`, but not in our builder `In("blob.hash", newHashes)` - so while I can remove the ones below on the `Database` package inputs, I can't remove this one without significant extra work.

pkg/database/plugin.go
peterbroadhurst: (sorry comments here and above are a bit interleaved)

PR #666, opened 2022-04-06
Remove misleading payload_ref DB field and fix Group JSON tag

peterbroadhurst:
In diagnosing a recent issue, I noticed that the `payloadRef` of a batch is a Database stored field, but is never set when you query in the database.

This is because we were writing it when we sealed the batch for delivery, but then overwriting it when we confirmed the batch.

It also had `VARCHAR(64)` in PSQL and `VARCHAR(256)` in SQLite.

The code actually only _needed_ it to pass the `payloadRef` from the batch upload operation, to the batch pin operation. Instead of relying on that, I've added it explicitly to the inputs of the Batch Pin operation - which overall feels much more consistent.

Then I've removed the DB field (for newly created envs - it nullable so no need for a migration to remove it).

I had to update the operations interface to return the outputs though, as it wasn't currently.

Unrelated fix - the `json:` tag on `Group` was wrong on `BatchPersisted` so we were returning `Group` rather than `group` on the JSON.


internal/privatemessaging/privatemessaging.go
```diff
@@ -298,11 +298,7 @@ func (pm *privateMessaging) sendData(ctx context.Context, tw *fftypes.TransportW
 				batch.Namespace,
 				batch.Payload.TX.ID,
 				fftypes.OpTypeDataExchangeSendBatch)
-			var groupHash *fftypes.Bytes32
-			if tw.Group != nil {
```
peterbroadhurst, line 302: The `tw.Group` object is only set on un-pinned transfers, where we can't guarantee the group has already been confirmed in an earlier pinned transfer. This code meant we weren't persisting the group for retry for pinned batches.

image

awrichar (approved):

PR #668, opened 2022-04-07
New operations admin APIs and revamped change-event listener on admin WebSocket

peterbroadhurst:
The policy engine described in https://github.com/hyperledger/firefly-ethconnect/issues/149#issuecomment-920866244 is starting to come together, with some evolved architecture.

The intention is for this to be a separate microservice, so it can be pluggable with different implementations easily, have a separate scaling model, and be closer to EthConnect in where it sits in the architecture than to the core.

However, the source of truth for the completion of operations is still FireFly. So this is more like an augmented _component_ of FireFly, than it is a plugin/connector that performs a particular task on behalf of FireFly. So as such I believe it should connect _in_ to FireFly to perform actions.

Currently (before this PR) we have some of things it would need - with gaps:
- Operations API - query only, and limited to one namespace.
- Operation change events - oddly side-car attached to the main application events, without sophisticated filtering

So this PR proposes we create a proper separation of concerns between the components that have super-access to update FireFly state, because they are part of it (the `admin` API already exists for this), and the external API that applications use.

- Adding `GET` `/admin/api/v1/operations` routes - including a `PUT` to update operation status (per same rules that apply to internal plugins like Data Exchange today)
- **Moving** the database change events firehose WebSocket to `/admin/ws` with extra options, and a simplified no-confirm model. It's best effort up to a buffer size, and then operation of the core is prioritized over avoiding any event loss, and always ephemeral. No guarantees.


nguyer:
I like the idea of the separation of APIs (no apostrophe). Perhaps maybe the name `admin` is a bit out of place now? The original intent of the `admin` API was for administrators of the system to have a way to perform "superuser" type actions. Perhaps it should be renamed to `internal`? It looks like it is evolving into an API for things _inside_ FireFly to use now. I'm not sure that has to be in this PR, but food for thought.

pkg/events/plugin.go
```diff
@@ -51,18 +51,6 @@ type Plugin interface {
 	DeliveryRequest(connID string, sub *fftypes.Subscription, event *fftypes.EventDelivery, data fftypes.DataArray) error
 }
 
-// ChangeEventListener is an optional interface for delivering database change events, only supported for ephemeral connections
-type ChangeEventListener interface {
-	// ChangeEvent delivers a change event - no ack for these
-	ChangeEvent(connID string, ce *fftypes.ChangeEvent)
-}
```
nguyer, line 58: So `ChangeEventListeners` are not a distinct concept anymore? `ChangeEvents` are just a thing that you get on the admin websocket now?

nguyer (approved):
Approved, but I still have the question about whether it should still be called the `admin` API or not. But we can defer that decision if you want.

pkg/events/plugin.go
peterbroadhurst: Mixing together application level events, and these change events, on the same WebSocket with different payloads has been removed. It felt really confusing, and the danger of opening a firehose you can't cope with is now pretty self-evident.

So I guess I think of it as more distinct now, in that you have a whole separate WebSocket to listen to.

So it's got nothing to do with the `events` plugin at all - it's a system-level thing built into the core.

internal/apiserver/route_admin_get_op_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/fftypes"
+)
+
+var adminGetOpByID = &oapispec.Route{
+	Name:   "adminGetOpByID",
+	Path:   "operations/{opid}",
+	Method: http.MethodGet,
+	PathParams: []*oapispec.PathParam{
+		{Name: "opid", Description: coremsgs.APIMessageTBD},
```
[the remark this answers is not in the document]
peterbroadhurst: 😳 🙇

PR #675, opened 2022-04-08
Update token connector snapshots in manifest

awrichar:
Ensures correct parsing of events in the wake of #673


awrichar:
Superseded by #677

PR #678, opened 2022-04-08
Configuration docs, and packaging config/i18n for re-use by microservices

peterbroadhurst:
A bunch of changes to improve our config docs (thanks @nguyer for your work there), and some structural changes to the `config` and `i18n` packages to allow them to be used by other microservices building in Go within the FireFly project.

- `pkg/config` - moved out the package for configuration, including the facility for Markdown generation
  - Split out the configuration that is base (just `log` and `lang`) from the Core configuration
  - `internal/coreconfig` contains the core configuration
  - Note that `coreconfig.Reset()` initializes all the defaults in a lock (needed to avoid concurrent map errors)
- `pkg/i18n` - moved out the translation framework
  - All the framework for translation moved
  - Tweak to how the default `en` translations are initialized, to allow sub-modules to keep adding keys
  - Split out `ErrorMessageKey` from `MessageKey` - with the distinction that errors have to have a registered prefix, that means microservices will not re-use message codes from other microservices
  - `FF10` errors are the FireFly Core errors
  - `FF00` errors are raised from the common utilities in `pkg` - note this mean renames of some codes
  - `FF201` errors are for the new Transaction Manager microservice being split out of EthConnect (the reason for doing this work now)


pkg/config/config.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+	log.SetLevel(GetString(LogLevel))
+	log.L(ctx).Debugf("Log level: %s", logrus.GetLevel())
+}
+
+func GenerateConfigMarkdown(ctx context.Context, keys []string) ([]byte, error) {
+	b := bytes.NewBuffer([]byte{})
+
+	rootKeyHeaderLevel := 2
+	// currentHeaderHierarchy := make([]string, 0)
```
nguyer, line 507: This can get removed. I'll take care of that in my follow up PR

pkg/i18n/en_base_config_descriptions.go
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package i18n
+
+//revive:disable
+var (
+	ConfigLang          = FFM("lang", "Default language for translation (API calls may support language override using headers)")
```
nguyer, line 21: Should this be `config.lang`?

nguyer (approved):
A couple of things to follow up on (which I will take care of) and one or two quick questions, but otherwise looks good.

internal/apiserver/route_admin_delete_config_record.go
```diff
@@ -28,11 +28,11 @@ var deleteConfigRecord = &oapispec.Route{
 	Path:   "config/records/{key}",
 	Method: http.MethodDelete,
 	PathParams: []*oapispec.PathParam{
-		{Name: "key", Example: "database", Description: i18n.MsgTBD},
+		{Name: "key", Example: "database", Description: coremsgs.APIMessageTBD},
```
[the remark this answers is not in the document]
nguyer: I see you've written the same thing elsewhere in the code comments 🙂

pkg/i18n/en_base_config_descriptions.go
peterbroadhurst: Yes... not immediately sure why the doc generator didn't complain - but I've fixed it.

PR #680, opened 2022-04-08
Add /api/v1/status/websockets

### LINKED ISSUE #679: Add a status endpoint to view currently connected websocket clients

awrichar:
Fixes #679


internal/events/websockets/websockets.go
```diff
@@ -152,3 +152,28 @@ func (ws *WebSockets) WaitClosed() {
 		ws.waitClose()
 	}
 }
+
+func (ws *WebSockets) GetStatus() *fftypes.WebSocketStatus {
+	status := &fftypes.WebSocketStatus{
+		Enabled:     true,
+		Connections: make([]*fftypes.WSConnectionStatus, 0),
+	}
+	for _, c := range ws.connections {
```
nguyer, line 161: Do we need to lock here to avoid a read while another goroutine is writing?

nguyer (commented):
Looks good. One question though on a potential concurrency issue.

nguyer:
And there are also merge conflicts to sort before merging

internal/events/websockets/websockets.go
awrichar: Yea, good call. Added that.

internal/apiserver/route_get_status_websockets.go
```diff
@@ -0,0 +1,41 @@
...
+var getStatusWebSockets = &oapispec.Route{
+	Name:            "getStatusWebSockets",
+	Path:            "status/websockets",
+	Method:          http.MethodGet,
+	PathParams:      nil,
+	QueryParams:     nil,
+	FilterFactory:   nil,
+	Description:     coremsgs.APIMessageTBD,
+	JSONInputValue:  nil,
+	JSONOutputValue: func() interface{} { return []*fftypes.WebSocketStatus{} },
```
peterbroadhurst, line 36: According to the test this is `&fftypes.WebSocketStatus{}` now

internal/events/websockets/websockets.go
```diff
@@ -138,3 +138,35 @@ func (ws *WebSockets) WaitClosed() {
...
 }
+
+func (ws *WebSockets) GetStatus() *fftypes.WebSocketStatus {
+	status := &fftypes.WebSocketStatus{
+		Enabled:     true,
+		Connections: make([]*fftypes.WSConnectionStatus, 0),
+	}
+
+	ws.connMux.Lock()
+	connections := ws.connections
```
peterbroadhurst, line 149: I don't think on its own this gives a safe copy.
Think this has to be:

```
wc.mux.Lock()
connections = make([]connection, len(ws.connections))
for i, c := range ws.connections {
  connections[i] = c
}
wc.mux.Unlock()
```

PR #686, opened 2022-04-10
Add helper at /apis/{apiName}/listen

awrichar:
This existed in the past and was removed due to perceived redundancy. However, it _would_ allow creating contract listeners with a single step if you know only the API name and not the underlying FFI ID (currently creating a listener in that case is a 2-step process, to go from API to FFI and then create the listener).

Note that this exists in parallel to the other RPC-style endpoints at `/apis/{apiName}/query` and `/apis/{apiName}/invoke`. It's worth discussing if a similar API should be added next to `/contracts/interfaces/{interfaceId}/query` and `/contracts/interfaces/{interfaceId}/invoke` (or possibly discussing the opposite, ie if _those_ should be pruned as redundant).

### DIFF
--- a/docs/swagger/swagger.yaml
+++ b/docs/swagger/swagger.yaml
@@ -778,6 +778,158 @@ paths:
           description: Success
         default:
           description: ""
+  /namespaces/{ns}/apis/{apiName}/listen/{eventPath}:
+    post:
+      description: 'TODO: Description'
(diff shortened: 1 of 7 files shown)


awrichar:
Replaced by https://github.com/hyperledger/firefly/pull/696

PR #687, opened 2022-04-10
Include interface metadata routes in FFI-generated Swagger

awrichar:
Allows fetching info about the associated FFI, including the details of all
methods and events.


internal/oapiffi/ffi2swagger.go
```diff
@@ -42,7 +42,22 @@ func NewFFISwaggerGen() FFISwaggerGen {
...
+			JSONOutputValue: func() interface{} { return &fftypes.FFI{} },
+			JSONOutputCodes: []int{http.StatusOK},
+		},
+	}
+	for _, method := range ffi.Methods {
+		routes = og.addMethodMetadata(routes, method)
+	}
+	for _, event := range ffi.Events {
+		routes = og.addEventMetadata(routes, event)
+	}
```
awrichar, line 60: After coming back to consider these along with the additional routes added in #696, I think these extra routes under `/interface/methods` and `/interface/events` are not necessary. The `/interface` helper would actually be the only one I vote to keep from this PR now.

awrichar: Pushed another commit to remove these routes completely.

PR #695, opened 2022-04-11
Add completed configuration descriptions and types

nguyer:
The main thing this PR adds is completed descriptions for all configuration options. It also adds a new `ffc()` for FireFly Configuration message keys for translations. The configuration key requires that a `fieldType` be set as another parameter in addition to the configuration description. The `fieldType` is a `string` that describes the Go type that FireFly Core will parse the field as.


internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigGlobalTTL  = ffc("config.global.cache.ttl", "The time to live (TTL) for the cache", timeDurationType)
+
+	ConfigGlobaltWsHeartbeatInterval     = ffc("config.global.ws.heartbeatInterval", "The amount of time to wait between heartbeat signals on the WebSocket connection", timeDurationType)
+	ConfigGlobalWsInitialConnectAttempts = ffc("config.global.ws.initialConnectAttempts", "The number of attempts FireFly will make to connect to the WebSocket when starting up, before failing", intType)
+	ConfigGlobalWsPath                   = ffc("config.global.ws.path", "The WebSocket sever URL to which FireFly should connect", "WebSocket URL "+stringType)
+	ConfigGlobalWsReadBufferSize         = ffc("config.global.ws.readBufferSize", "The size in bytes of the read buffer for the WebSocket connection", byteSizeType)
+	ConfigGlobalWsWriteBufferSize        = ffc("config.global.ws.writeBufferSize", "The size in bytes of the write buffer for the WebSocket connection", byteSizeType)
+
+	ConfigGlobalTLSCaFile           = ffc("config.global.tls.caFile", "The path to the CA file for the admin API", stringType)
```
peterbroadhurst, line 58: Thank `admin` isn't applicable here as it's a `global.`

peterbroadhurst: Same applies to next few lines

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigGlobalTLSCertFile         = ffc("config.global.tls.certFile", "The path to the certificate file for the admin API", stringType)
+	ConfigGlobalTLSClientAuth       = ffc("config.global.tls.clientAuth", "Enables or disables client auth for the admin API", stringType)
+	ConfigGlobalTLSEnabled          = ffc("config.global.tls.enabled", "Enables or disables TLS on the admin API", booleanType)
+	ConfigGlobalTLSKeyFile          = ffc("config.global.tls.keyFile", "The path to the private key file for the admin API", stringType)
+	ConfigGlobalTLSHandshakeTimeout = ffc("config.global.tlsHandshakeTimeout", "The maximum amount of time to wait for a successful TLS handshake", timeDurationType)
+
+	ConfigGlobalBodyTemplate          = ffc("config.global.bodyTemplate", "The body go template string to use when making HTTP requests", "Go template "+stringType)
```
peterbroadhurst, line 65: Would be nice to make go template a type and link off to the go docs - non go people might need some help with the templating `{{ .Stuff }}` syntax

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigGlobalCustomClient          = ffc("config.global.customClient", "Allows injection of a custom HTTP client. **Used only for testing.**", stringType)
```
peterbroadhurst, line 66: Not a blocker for merge, but we might want to have a way to retract this, and the `panic` one (will note that one separately) that are only there for build time.

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigGlobalExpectContinueTimeout = ffc("config.global.expectContinueTimeout", "See [ExpectContinueTimeout in the Go docs](https://pkg.go.dev/net/http#Transport)", timeDurationType)
+	ConfigGlobalHeaders               = ffc("config.global.headers", "Adds custom headers to HTTP requests", stringType)
```
peterbroadhurst, line 68: I though this one was a map, not a string

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigAdminPreInit      = ffc("config.admin.preinit", "Enables the pre-init mode. This mode will let the FireFly Core process start, but not initialize any plugins, besides the database to read any configuration overrides. This allows the admin HTTP API to be used to define custom configuration before starting the rest of FireFly Core.", booleanType)
+	ConfigAdminPublicURL    = ffc("config.admin.publicURL", "The fully qualified public URL for the admin API. This is used for building URLs in HTTP responses and in OpenAPI Spec generation.", "URL "+stringType)
+	ConfigAdminReadTimeout  = ffc("config.admin.readTimeout", "The maximum time to wait when reading from an HTTP connection", timeDurationType)
+	ConfigAdminWriteTimeout = ffc("config.admin.writeTimeout", "The maximum time to wait when writing to an HTTP connection", timeDurationType)
+
+	ConfigAPIDefaultFilterLimit = ffc("config.api.defaultFilterLimit", "The maximum number of rows to return if no limit is specified on an API request", intType)
+	ConfigAPIMaxFilterLimit     = ffc("config.api.maxFilterLimit", "The maximum number of rows to return if no limit is specified on an API request", intType)
```
peterbroadhurst, line 85: This one is the largest value of `limit` that a user of the API can supply

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigBlockchainEthereumAddressResolverRetainOriginal = ffc("config.blockchain.ethereum.addressResolver.retainOriginal", "When true the original pre-resolved string is retained after the lookup, and passed down to Ethconnect as the from address", booleanType)
+	ConfigBlockchainEthereumAddressResolverURL            = ffc("config.blockchain.ethereum.addressResolver.url", "The URL of the address resolver", stringType)
+	ConfigBlockchainEthereumAddressResolverURLTemplate    = ffc("config.blockchain.ethereum.addressResolver.urlTemplate", "The URL Go template string to use when calling the address resolver", "Go template "+stringType)
+
+	ConfigBlockchainEthereumAddressResolverProxyURL = ffc("config.blockchain.ethereum.addressResolver.proxy.url", "The URL of the address resolver proxy", "URL "+stringType)
+
+	ConfigBlockchainEthereumEthconnectBatchSize    = ffc("config.blockchain.ethereum.ethconnect.batchSize", "The maximum number of transactions to send in a single request to Ethconnect", intType)
```
peterbroadhurst, line 117: I believe this relates to how we configure the event stream, rather than transaction submission. So this is:

```
The number of events EthConnect should batch together for delivery to FireFly core. Only applies when automatically creating a new event stream.
```

internal/coremsgs/en_config_descriptions.go
nguyer: I like that. I'll think of a way to do this cleanly

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigBlockchainFabricFabconnectChaincode    = ffc("config.blockchain.fabric.fabconnect.chaincode", "The name of the Fabric chaincode that FireFly will use for BatchPin transactions", stringType)
+	ConfigBlockchainFabricFabconnectChannel      = ffc("config.blockchain.fabric.fabconnect.channel", "The Fabric channel that FireFly will use for BatchPin transactions", stringType)
+	ConfigBlockchainFabricFabconnectPrefixLong   = ffc("config.blockchain.fabric.fabconnect.prefixLong", "The prefix that will be used for Fabconnect specific HTTP headers when FireFly makes requests to Fabconnect", stringType)
+	ConfigBlockchainFabricFabconnectPrefixShort  = ffc("config.blockchain.fabric.fabconnect.prefixShort", "The prefix that will be used for Fabconnect specific query parameters when FireFly makes requests to Fabconnect", stringType)
+	ConfigBlockchainFabricFabconnectSigner       = ffc("config.blockchain.fabric.fabconnect.signer", "The Fabric signing key to use when submitting transactions to Fabconnect", stringType)
```
peterbroadhurst, line 133: I don't know that this one is correct, might need to check with @jimthematrix or the code.
The signer of transactions should be set consistently with everything else in FireFly - using the `key` field on the input, or falling back to the `org.key` configuration key.

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+
+	ConfigCorsCredentials = ffc("config.cors.credentials", "CORS setting to control whether a browser allows credentials to be sent to this API", booleanType)
+	ConfigCorsDebug       = ffc("config.cors.debug", "Whether debug is enabled for the CORS implementation", booleanType)
+	ConfigCorsEnabled     = ffc("config.cors.enabled", "Whether cors is enabled", booleanType)
+	ConfigCorsHeaders     = ffc("config.cors.headers", "CORS setting to control the allowed headers", stringType)
+	ConfigCorsMaxAge      = ffc("config.cors.maxAge", "The maximum age a browser should rely on CORS checks", timeDurationType)
+	ConfigCorsMethods     = ffc("config.cors.methods", " CORS setting to control the allowed methods", stringType)
+	ConfigCorsOrigins     = ffc("config.cors.origins", "CORS setting to control the allowed origins", stringType)
+
+	ConfigDatabaseMaxChartRows = ffc("config.database.maxChartRows", "The maximum rows to fetch for each histogram bucket", intType)
```
peterbroadhurst, line 152: I assert this config key is wrong (not related to your change).
`config.database.XYZ` should be reserved for plugins of name `XYZ`.
This should be changed to `config.histograms.maxChartRows` or similar

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigDatabaseType         = ffc("config.database.type", "The type of the database interface plugin to use", intType)
+
+	ConfigDatabasePostgresMaxConnIdleTime = ffc("config.database.postgres.maxConnIdleTime", "The maximum amount of time a database connection can be idle", timeDurationType)
```
peterbroadhurst, line 155: Might be worth a `global.` for future database plugins

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigDataexchangeFfdxProxyURL = ffc("config.dataexchange.ffdx.proxy.url", "The URL of the Data Exchange HTTP proxy", "URL "+stringType)
+
+	ConfigDebugPort = ffc("config.debug.port", "An HTTP port on which to enable the go debugger", intType)
+
+	ConfigDownloadWorkerCount       = ffc("config.download.worker.count", "The number of download workers", intType)
+	ConfigDownloadWorkerQueueLength = ffc("config.download.worker.queueLength", "The length of the work queue in the channel to the workers - defaults to 2x the worker count", intType)
+
+	ConfigEventAggregatorBatchSize            = ffc("config.event.aggregator.batchSize", "The maximum number of records to read from the DB before performing an aggregation run", byteSizeType)
+	ConfigEventAggregatorBatchTimeout         = ffc("config.event.aggregator.batchTimeout", "How long to wait for new events to arrive before performing aggregation on a page of events", timeDurationType)
+	ConfigEventAggregatorFirstEvent           = ffc("config.event.aggregator.firstEvent", "The first event the aggregator should process, if no previous offest is stored in the DB. Valid options are `oldest` or `newest`", stringType)
```
peterbroadhurst, line 182: Think this is defunct config now - the key can be removed (not queried anywhere).

internal/coremsgs/en_config_descriptions.go
nguyer: Makes sense. Want me to change it in this PR?

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+
+	ConfigOpupdateWorkerBatchMaxInserts = ffc("config.opupdate.worker.batchMaxInserts", "The maximum number of database inserts to include when writing a single batch of messages + data", intType)
+	ConfigOpupdateWorkerBatchTimeout    = ffc("config.opupdate.worker.batchTimeout", "How long to wait for more messages to arrive before flushing the batch", timeDurationType)
+	ConfigOpupdateWorkerCount           = ffc("config.opupdate.worker.count", "The number of operation update works", intType)
+	ConfigOpupdateWorkerQueueLength     = ffc("config.opupdate.worker.queueLength", "The size of the queue for the Operation Update worker", intType)
+
+	ConfigOrchestratorStartupAttempts = ffc("config.orchestrator.startupAttempts", "The number of times to attempt to connect to core infrastructure on startup", stringType)
+
+	ConfigOrgDescription = ffc("config.org.description", "A description of the organization to which this FireFly node belongs", stringType)
+	ConfigOrgIdentity    = ffc("config.org.identity", "`DEPRECATED` Please use `org.key` instead", stringType)
```
peterbroadhurst, line 244: Different treatment of `DEPRECATED` vs. **Deprecated** below

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
-	ConfigTokensProxyURL = ffm("config.tokens[].proxy.url", "TBD")
-
-	ConfigUIEnabled = ffm("config.ui.enabled", "Enables the web user interface")
-	ConfigUIPath    = ffm("config.ui.path", "The file system path which contains the static HTML, CSS, and JavaScript files for the user interface")
+	ConfigPublicstorageType = ffc("config.publicstorage.type", "**Deprecated** See `config.sharedstorage.type` instead.", stringType)
+
+	ConfigPublicstorageIpfsAPIURL          = ffc("config.publicstorage.ipfs.api.url", "The URL for the IPFS API", "URL "+stringType)
+	ConfigPublicstorageIpfsAPIProxyURL     = ffc("config.publicstorage.ipfs.api.proxy.url", "The URL for the IPFS proxy", "URL "+stringType)
+	ConfigPublicstorageIpfsGatewayURL      = ffc("config.publicstorage.ipfs.gateway.url", "The URL for the IPFS Gateway", "URL "+stringType)
+	ConfigPublicstorageIpfsGatewayProxyURL = ffc("config.publicstorage.ipfs.gateway.proxy.url", "The URL for the IPFS Gateway proxy", "URL "+stringType)
```
peterbroadhurst, line 260: This might applied to other similar `.proxy` entires, but particularly for IPFS. This sounds like it's mandatory, but actually it's more like `Optional HTTP proxy server to use when connecting to the IPFS Gateway`

internal/coremsgs/en_config_descriptions.go
nguyer: I actually thought of and tried this. Without introducing some sort of wildcard key searching it doesn't work with the way globals currently work, because `postgres` is in the middle of the key.

So `config.database.myKey` would match `config.global.myKey`
But there's no way to match `config.database.postgres.myKey` to `config.database.*.myKey` or anything like that right now. I'm not opposed to adding this if we really want it to eliminate a few keys here.

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigTokensName      = ffc("config.tokens[].name", "The name of the Tokens Connector. This will be used in the FireFly API path to refer to this specific Token Connector", stringType)
+	ConfigTokensPlugin    = ffc("config.tokens[].plugin", "The name of the Tokens Connector plugin to use", stringType)
+	ConfigTokensURL       = ffc("config.tokens[].url", "The URL of the Token Connector", "URL "+stringType)
+
+	ConfigTokensProxyURL = ffc("config.tokens[].proxy.url", "The URL of the Token Connector proxy", "URL "+stringType)
+
+	ConfigUIEnabled = ffc("config.ui.enabled", "Enables the web user interface", booleanType)
+	ConfigUIPath    = ffc("config.ui.path", "The file system path which contains the static HTML, CSS, and JavaScript files for the user interface", stringType)
+
+	ConfigAPIOASPanicOnMissingDescription = ffc("config.api.oas.panicOnMissingDescription", "Used when building FireFly to verify all structures and APIs have documentation. Primarily used for testing.", booleanType)
```
peterbroadhurst, line 280: As above for build/test only endpoints, that should be redacted in the docs

peterbroadhurst (changes requested):
Some ideas, and some explicit requests for wording changes.

internal/coremsgs/en_config_descriptions.go
nguyer: This one is still queried here https://github.com/hyperledger/firefly/blob/main/internal/events/aggregator.go#L81

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,265 @@ import (
...
+	ConfigDownloadWorkerQueueLength = ffc("config.download.worker.queueLength", "The length of the work queue in the channel to the workers - defaults to 2x the worker count", intType)
+
+	ConfigEventAggregatorBatchSize            = ffc("config.event.aggregator.batchSize", "The maximum number of records to read from the DB before performing an aggregation run", byteSizeType)
+	ConfigEventAggregatorBatchTimeout         = ffc("config.event.aggregator.batchTimeout", "How long to wait for new events to arrive before performing aggregation on a page of events", timeDurationType)
+	ConfigEventAggregatorFirstEvent           = ffc("config.event.aggregator.firstEvent", "The first event the aggregator should process, if no previous offest is stored in the DB. Valid options are `oldest` or `newest`", stringType)
+	ConfigEventAggregatorOpCorrelationRetries = ffc("config.event.aggregator.opCorrelationRetries", "How many times to correlate an event for an operation (such as tx submission) back to an operation. Needed because the operation update might come back before we are finished persisting the ID of the request", intType)
```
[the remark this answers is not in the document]
nguyer: This one is still queried here https://github.com/hyperledger/firefly/blob/main/internal/events/event_manager.go#L136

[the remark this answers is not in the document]
peterbroadhurst: sorry @nguyer - I meant `ConfigEventAggregatorFirstEvent`

internal/coremsgs/en_config_descriptions.go
```diff
@@ -20,301 +20,268 @@ import (
 	"github.com/hyperledger/firefly/pkg/i18n"
 )
 
-var ffm = i18n.FFM
+var ffc = i18n.FFC
+var timeDurationType = "[`time.Duration`](https://pkg.go.dev/time#Duration)"
```
peterbroadhurst, line 24: Would be nice for these to be exposed, so I could re-use them in docs for a Microservice using this config utility

nguyer: That makes sense. I can probably move them somewhere that would do that

internal/coremsgs/en_config_descriptions.go
peterbroadhurst: We can defer this 👍

internal/coremsgs/en_config_descriptions.go
peterbroadhurst: Thanks for moving that 👍

internal/coremsgs/en_config_descriptions.go
peterbroadhurst: Sorry, my bad -agreed looking at the code. I don't thin it makes sense to keep it

internal/coremsgs/en_config_descriptions.go
peterbroadhurst: Thanks @nguyer 👍

PR #696, opened 2022-04-11
Add helpers at /apis/{apiName}/listeners/{eventPath}

awrichar:
This includes the commit from #693.

Add helper URLs:

```
GET /apis/{apiName}/listeners/{eventPath} - list all listeners matching this event on this API
POST /apis/{apiName}/listeners/{eventPath} - create a new listener for this event on this API
```

Also makes `topic` a required field for contract listeners, and enforces uniqueness of listeners across these 2 dimensions:
* namespace + name
* topic + contract location + event signature

Violating either uniqueness constraint when attempting to create a listener will result in HTTP 409, so applications can rely on this behavior to perform "create this listener if it does not already exist". Note that `name` is still optional, and a unique one will be assigned at creation time if none is specified.


internal/blockchain/ethereum/ethereum.go
```diff
@@ -616,9 +616,17 @@ func (e *Ethereum) QueryContract(ctx context.Context, location *fftypes.JSONAny,
...
 
-func (e *Ethereum) ValidateContractLocation(ctx context.Context, location *fftypes.JSONAny) (err error) {
-	_, err = parseContractLocation(ctx, location)
-	return
+func (e *Ethereum) NormalizeContractLocation(ctx context.Context, location *fftypes.JSONAny) (result *fftypes.JSONAny, err error) {
+	parsed, err := parseContractLocation(ctx, location)
+	if err != nil {
+		return nil, err
+	}
+	parsed.Address = strings.ToLower(parsed.Address)
```
peterbroadhurst, line 624: We have a `validateEthAddress` normalization function for this

internal/blockchain/fabric/fabric.go
```diff
@@ -334,6 +336,10 @@ func (f *Fabric) handleBatchPinEvent(ctx context.Context, msgJSON fftypes.JSONOb
 	})
 }
 
+func (f *Fabric) buildEventLocationString(msgJSON fftypes.JSONObject) string {
+	return fmt.Sprintf("chaincode=%s", msgJSON.GetString("chaincodeId"))
```
peterbroadhurst, line 340: Have I understood correctly that we're using a `x=y` querystring like syntax for Fabric, and a serialized JSON object for Ethereum? Wondering why that choice

For Fabric, how is the `channel` represented?

peterbroadhurst (commented):
Few questions from me here @awrichar

internal/blockchain/fabric/fabric.go
awrichar: This particular string is used for Prometheus metrics filtering. It would ideally include the channel as well, but there's no easy way to get it in the context where this is called (and seems like not a major issue as it only affects metrics). Ethereum has a similar "address=0x0000" format for this same purpose.

The serialization of the location string for storage is JSON and is consistent across Ethereum/Fabric. But yes, we do have two different serializations of the same info for different purposes.

internal/oapiffi/ffi2swagger.go
```diff
@@ -24,13 +24,25 @@ import (
...
 	"github.com/hyperledger/firefly/pkg/fftypes"
 )
 
 type FFISwaggerGen interface {
 	Generate(ctx context.Context, baseURL string, api *fftypes.ContractAPI, ffi *fftypes.FFI) *openapi3.T
 }
 
+type ContractListenerInput struct {
+	Name    string                           `ffstruct:"ContractListener" json:"name,omitempty"`
+	Topic   string                           `ffstruct:"ContractListener" json:"topic,omitempty"`
```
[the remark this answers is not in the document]
awrichar: There's a separate type below that includes Location. This generator has knowledge of whether the underlying API is bound to a location or requires specifying one on each call, and generates the input schemas accordingly.

peterbroadhurst:
Also, in the header comment of the PR (and soon in API descriptions fyi @nguyer ) would be good to note the `409` conflict behavior that allows apps the ability to do a create-if-not-exist check on startup with this great new API.

db/migrations/postgres/000080_add_contractlistener_signature.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE contractlisteners DROP COLUMN signature;
```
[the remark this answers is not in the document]
awrichar: Probably. I made a mental note to come back on indexes and then promptly forgot about it.

internal/blockchain/fabric/fabric.go
peterbroadhurst: Thanks for the clarification. Understood 👍

internal/blockchain/ethereum/ethereum.go
```diff
@@ -690,6 +698,26 @@ func (e *Ethereum) FFIMethodToABI(ctx context.Context, method *fftypes.FFIMethod
 	return abiElement, nil
 }
 
+func ABIMethodToSignature(abi *ABIElementMarshaling) string {
+	result := abi.Name + "("
+	if len(abi.Inputs) > 0 {
```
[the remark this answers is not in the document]
awrichar: Added handling for tuples

peterbroadhurst (approved):
👍

PR #709, opened 2022-04-13
Fix token approvals and clarify fields "protocolId", "locator", and "subject"

### LINKED ISSUE #661: Synchronous token approval timeout

awrichar:
Depends on https://github.com/hyperledger/firefly-tokens-erc1155/pull/67 and https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/42

This started as a quest to clean up token approvals, but ended up encompassing a few related items:

* The term "protocolId" has come to mean (fairly specifically) "an ID meaningful in the context of an underlying blockchain protocol".
  * The existing "protocolId" of token approvals is renamed to "subject", and token approvals get a new "protocolId" field that better aligns with the above.
  * Token transfers keep their "protocolId" unchanged, as it already aligns.
  * The existing "protocolId" of token pools is renamed to "locator".
  * Blockchain events will be de-duplicated based on their "protocolId" - only one event per namespace+listener+protocolId will be recorded, to avoid recording the same blockchain event multiple times.
* Token approvals will record _all_ historical approvals (instead of sometimes overwriting old ones). As a convenience for query purposes, the most recent approval for each subject will be notated with a new field "active=true".
* When matching token approval (and transfer) events to operations, event manager will now consider the operation ID _and_ the token connector and pool embedded in the operation inputs before considering it a match. This ensures that side-effects in other pools will not be matched with the wrong operation and return unexpected results for synchronous actions (when using confirm=true). Resolves #661.

This does introduce changes in the interface with token connectors, so will require connectors to be upgraded alongside it.


internal/events/tokens_approved.go
```diff
@@ -54,61 +56,71 @@ func (em *eventManager) loadApprovalOperation(ctx context.Context, tx *fftypes.U
...
-			}
-
-			if existingLocal != nil {
-				approval.LocalID = fftypes.NewUUID()
-			}
-		} else {
-			log.L(ctx).Infof("Updating existing approval with local ID: %s", existing.LocalID)
-			approval.LocalID = existing.LocalID
+		if existingLocal != nil {
+			approval.LocalID = fftypes.NewUUID()
```
peterbroadhurst, line 93: A comment here would be worthwhile. My understanding is:

```
// The approval was initiated by FireFly and one UUID was allocated as a `localId` for the approval when it was
// initiated. This allows the `confirm=true` code to know when the first approval record gets inserted into the
// database that was correlated to that approval. However, approvals can result in many approval records
// depending on the on-chain mapping of the scope of an approval. So the connector might generate multiple
// notifications (with different `protocolId`), but all will reach this code with the same `localId`.
// So we need to generate a new one (for all apart from the first)
```

peterbroadhurst (approved):
Looks great. Think this is a significant step forwards in consistency.

I also reviewed:
- https://github.com/hyperledger/firefly-tokens-erc1155/pull/67
- https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/42

I think those need merging, releases cutting, and a manifest update here before this merges.

There's one place I had to stare at the code for a while to work it out, so I've proposed a comment based on my understanding.

Marking approval, and leaving you to work through the mechanics of the merge.

internal/events/tokens_approved.go
awrichar: Yep, I'll add some notes

awrichar: Added a lot of notes and also cleaned up the method further to (hopefully) make it clearer to follow.

Issue #692, opened 2022-04-11
New Orgs / Nodes Rejecting Previous Node Identity Claims


hfuss:
## Description

When creating multi-org networks via the Helm chart, we would stand up the first node for an org, then stand up a new node / org, then another and so on (a bit different from how the CLI creates networks). In the new nodes, we would observe them rejecting the first org's node identity claim messages due to:

```
[2022-04-11T17:19:21.074Z]  WARN poc4-par2-firefly: Unable to process identity claim 564be456-7834-4218-84b9-1a82c9011270: FF10214: Identity '8e8b6544-df58-43fe-935e-f0078b760e7d' not found in identity chain for did:firefly:node/poc4-par1-firefly 'd81e6067-bb73-4613-83c3-49b4a0a08051' dbtx=85C5Fddx role=aggregator
```

The ethconnects were in catchup mode, so the other nodes were presumed to be processing the org and node identity claim messages in the same batchpin. So it appears to be a bug in the logic where the aggregator does not account for needing to process org identity claims before node identity claims within the same batch. 

## Other Info

FireFly Version: `1.0.0-rc3`
Ethconnect Version: `v3.1.5`
Blockchain Provider: `Quorum v21.10.0 (Geth v1.9.25)`

Relevant Ethconnect Configuration:

```yaml
rest:
  rest-gateway:
    leveldb:
      path: "/qdata/ethconnect/receiptsdb"
      queryLimit: 100
    alwaysManageNonce: false
    maxTXWaitTime: 120
    openapi:
      storagePath: "/qdata/ethconnect/openapi"
      eventsDB: "/qdata/ethconnect/eventsdb"
      catchupModeBlockGap: 10000
      catchupModePageSize: 10000
```

## Logs

First Org / Node : [poc4-par1.log](https://github.com/hyperledger/firefly/files/8466801/poc4-par1.log)

Other Orgs / Nodes:
* [poc4-par2.log](https://github.com/hyperledger/firefly/files/8466802/poc4-par2.log)
* [poc4-par3.log](https://github.com/hyperledger/firefly/files/8466803/poc4-par3.log)

peterbroadhurst:
```
... The child identity ended up coming in first to the aggregator, and was rejected
[...]

PR #715, opened 2022-04-13
Perform all identity claims, under the root identity topic ordering context

peterbroadhurst:
Investigating https://github.com/hyperledger/firefly/issues/692 I realized that the only safe thing to do in terms of an ordering context, given that identities are arbitrarily nested in their hierarchy, is to broadcast them all under a single ordering context up to the root identity.  Then if a node is replaying a chain and potentially processing batches of blockchain transactions, it will be assured to get the same outcome in sequencing of identities for the whole hierarchy.

### DIFF
--- a/internal/broadcast/definition.go
+++ b/internal/broadcast/definition.go
@@ -50,6 +50,11 @@ func (bm *broadcastManager) BroadcastIdentityClaim(ctx context.Context, ns strin
 		return nil, err
 	}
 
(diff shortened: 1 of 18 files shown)


peterbroadhurst:
After thinking about this a bunch more, I'm not yet sure this is the right fix.

Because after the identity is established, it's possible you would then broadcast/private-send a message from that identity straight away. There's no way for the send to be on the same `topic`.

So this leads me to think we need to add a new Rewind, by Author DID.

PR #707, opened 2022-04-13
E2E account creation

nguyer:
This PR includes only testing changes, which are necessary for the upcoming `v0.0.47` release of the FireFly CLI.

It moves the responsibility for blockchain account creation to the CLI, and uses the new `stackState.json` file generated by the CLI to get a list of available accounts that can be used in tests.

This PR also enables the identity E2E tests for Fabric and Besu based FireFly stacks.

PR #716, opened 2022-04-13
DID rewind

### LINKED ISSUE #692: New Orgs / Nodes Rejecting Previous Node Identity Claims

peterbroadhurst:
Fixes #692 
In PR chain with #707

This PR adds a new reason for a rewind - to go back and confirm messages that were previously parked because the author was not known.

We cannot assure that the identity verification for an identity, will be processed by the aggregator before messages sent by that identity, or child identities registered under that identity. Because they are all sent on different topics.


internal/definitions/definition_handler_identity_claim_test.go
```diff
@@ -177,13 +177,14 @@ func TestHandleDefinitionIdentityClaimCustomWithExistingParentVerificationOk(t *
 	action, err := dh.HandleDefinitionBroadcast(ctx, bs, claimMsg, fftypes.DataArray{claimData}, fftypes.NewUUID())
 	assert.Equal(t, HandlerResult{Action: ActionConfirm}, action)
 	assert.NoError(t, err)
+	assert.Equal(t, bs.confirmedDIDClaims, []string{custom1.DID})
```
nguyer, line 180: I'm not sure I really trust these `bs` claims...

internal/events/aggregator_rewind.go
```diff
@@ -283,3 +293,27 @@ func (rw *rewinder) getRewindsForBlobs(ctx context.Context, newHashes []driver.V
...
+
+func (rw *rewinder) getRewindsForDIDs(ctx context.Context, dids []driver.Value, batchIDs map[fftypes.UUID]bool) error {
+
+	// We need to find all pending messages, that are authored by this DID
+	fb := database.MessageQueryFactory.NewFilter(ctx)
+	filter := fb.And(
+		fb.Eq("state", fftypes.MessageStatePending),
+		fb.In("author", dids),
+	)
+	records, err := rw.database.GetMessageIDs(ctx, filter)
```
nguyer, line 305: Is there a risk here that this becomes a large query and needs to be paginated?

manifest.json
```diff
@@ -44,6 +44,6 @@
     "release": "v0.7.0_50810a4"
   },
   "cli": {
-    "tag": "v0.0.46"
+    "tag": "6fb3585a8e279ff9860f3910625cb5b0ee56c749"
```
nguyer, line 47: Should be changed to an actual release when `v0.0.48` is available

nguyer (approved):

internal/events/aggregator_rewind.go
peterbroadhurst: I did consider that, and determined the answer should be no (for now):
- We are only querying the id+sequence, so the in-memory size is very small per record
- We are only querying things from a given author that has just become confirmed, so we should be dealing with a small number of records

PR #722, opened 2022-04-15
Honor "confirm" parameter for blockchain invoke

awrichar:
~~In a PR chain with #720~~ (rebased now)

Summary of the changes:
* The `confirm` parameter for `/invoke` will now be honored (previously it was ignored)
* The `fftypes.ContractCallResponse` type is removed, and `/invoke` requests now just return `fftypes.Operation`
* Two new event types are added - `EventTypeBlockchainInvokeOpSucceeded` and `EventTypeBlockchainInvokeOpFailed` - which reference an operation ID (open to thoughts on whether `topic` and `correlator` should be filled in)


internal/contracts/operations.go
```diff
@@ -73,6 +73,21 @@ func (cm *contractManager) RunOperation(ctx context.Context, op *fftypes.Prepare
 }
 
 func (cm *contractManager) OnOperationUpdate(ctx context.Context, op *fftypes.Operation, update *operations.OperationUpdate) error {
+	// Special handling for OpTypeBlockchainInvoke, which writes an event when it succeeds or fails
+	if op.Type == fftypes.OpTypeBlockchainInvoke {
+		if update.Status == fftypes.OpStatusSucceeded {
+			event := fftypes.NewEvent(fftypes.EventTypeBlockchainInvokeOpSucceeded, op.Namespace, op.ID, op.Transaction, "")
```
awrichar, line 79: Note that I am not setting a `topic` or `correlator` on this event. Open to thoughts if there's anything sensible to be used for either. All we have is the operation (and therefore the inputs to the "invoke" request).

nguyer (approved):

peterbroadhurst:
I do see a couple of coverage gaps:
- https://app.codecov.io/gh/hyperledger/firefly/compare/722/tree/internal/contracts/operations.go
- https://app.codecov.io/gh/hyperledger/firefly/compare/722/tree/internal/contracts/manager.go

internal/contracts/operations.go
peterbroadhurst: I don't have and great solutions.
For blockchain events, we use the specified topic, then fall back to the listener ID. Special case for the built-in listener.

https://github.com/hyperledger/firefly/blob/0cdaa9747040fb6ee07683464b1c21463dbf29c4/internal/events/blockchain_event.go#L72-L87

I don't see any obvious parallels.
My best suggestion in terms of usefulness on the `topic` is the first 64 characters of the query-stringified `location`.
Whether or not that's good enough to justify the cost of extracting it from the input, I don't know.

peterbroadhurst: ... and I realize it would need sanitization to fit the string requirements of the topic I think

awrichar: Yea - extracting, truncating, and sanitizing that location seems like a lot to me. Going to leave it for now and we can revisit if needed.

awrichar:
There should not be any coverage gaps; I think it was just weirdness with being behind main. Synced the branch to hopefully clear that up.

PR #724, opened 2022-04-18
Parse input params for /query endpoints in ffi2swagger

### LINKED ISSUE #718: Simple Storage Contract error via contract swagger API.
Screen Shot 2022-04-15 at 12 37 58 PM
The query post method should have an entry for an empty body. As reported by @nguyer 
It does not as the screen shot shows.
This probably a bug.

awrichar:
Fixes #718

### DIFF
--- a/internal/oapiffi/ffi2swagger_test.go
+++ b/internal/oapiffi/ffi2swagger_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/ghodss/yaml"
 	"github.com/hyperledger/firefly/pkg/fftypes"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


internal/oapiffi/ffi2swagger_test.go
```diff
@@ -64,6 +65,11 @@ func testFFI() *fftypes.FFI {
 					},
 				},
 			},
+			{
+				Name:     "method2",
+				Pathname: "method2",
+				/* no params */
```
nguyer, line 71: I _think_ even when there are no params, the FFI parser wants to at least see `params: []`

nguyer (approved):

internal/oapiffi/ffi2swagger_test.go
awrichar: I think this unit test confirms that `nil` and `&fftypes.FFIParams{}` are effectively the same thing here. Either way the JSON schema will get generated with an `input` object that contains no properties.

PR #725, opened 2022-04-18
Add "tx.blockchainId" to BlockchainEvent type

awrichar:
Every blockchain event will have a blockchain TX ID (such as an Ethereum
transaction hash), whether or not it has a FireFly transaction. This is
useful to record as another indexed field on the table.

It will also allow manually correlating blockchain events to FireFly transactions
in the case that the blockchain doesn't pass through enough data to do the
correlation based on FireFly IDs such as (custom smart contracts, ERC20 transfers
without a "data" argument, etc).


internal/coremsgs/en_struct_descriptions.go
```diff
@@ -161,6 +161,7 @@ var (
 	TransactionType          = ffm("Transaction.type", "The type of the FireFly transaction")
 	TransactionNamespace     = ffm("Transaction.namespace", "The namespace of the FireFly transaction")
 	TransactionCreated       = ffm("Transaction.created", "The time the transaction was created on this node. Note the transaction is individually created with the same UUID on each participant in the FireFly transaction")
+	TransactionBlockchainID  = ffm("Transaction.blockchainId", "The blockchain transaction ID, in the format specific to the blockchain involved in the transaction. Not all FireFly transactions include a blockchain")
```
nguyer, line 164: It looks like this message is a copy of the one below it. Was it intentional to not include the last half of the string, and not end the sentence with a period?

nguyer (approved):

internal/coremsgs/en_struct_descriptions.go
awrichar: Yes. Seems like every message in this file omits the final period (not sure if there's a reason, but that's how it looks). I omitted the last bit since it seemed to apply specifically to the plural field (ie listing multiple blockchain IDs).

peterbroadhurst (approved):

PR #729, opened 2022-04-19
Fix account creation for nightly integration runs

nguyer:
We recently merged a change in how blockchain accounts are created for the E2E tests. This worked fine for a normal run. The nightly integration test also stops all the services, starts them up again, and re-runs all the tests. This meant that there weren't enough "extra" test blockchain accounts allocated for the second test run after the restart. This PR fixes that.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -11,19 +11,19 @@ STACKS_DIR=~/.firefly/stacks
 
 create_accounts() {
   if [ "$TEST_SUITE" == "TestEthereumE2ESuite" ]; then
-      # Create 5 new accounts for use in testing
+      # Create 4 new accounts for use in testing
       for i in {1..4}
       do
           $CLI accounts create $STACK_NAME
       done
   elif [ "$TEST_SUITE" == "TestFabricE2ESuite" ]; then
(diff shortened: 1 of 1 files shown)


test/e2e/run.sh
```diff
@@ -9,6 +9,24 @@ CLI="ff -v --ansi never"
...
 
+create_accounts() {
+  if [ "$TEST_SUITE" == "TestEthereumE2ESuite" ]; then
+      # Create 5 new accounts for use in testing
+      for i in {1..4}
+      do
+          $CLI accounts create $STACK_NAME
+      done
+  elif [ "$TEST_SUITE" == "TestFabricE2ESuite" ]; then
+      # Create 5 new accounts for use in testing
```
peterbroadhurst, line 20: Comment doesn't match code

peterbroadhurst (approved):
Thanks @nguyer - one note about the comments (didn't get introduced in this PR, just moved)

nguyer:
Thanks. Just running a couple more tests locally here too. I think fabric key names need to move to random IDs. Creating `user_1`, `user_2`... each time will cause collisions which will throw off the check the tests do to see if the key is in use already. Testing one more small tweak here then will push an update.

PR #733, opened 2022-04-19
Use blank type for JSONAny in OpenAPI

awrichar:
Type is unknown, as it may be any valid JSON.

See https://github.com/hyperledger/firefly-sdk-nodejs/issues/18

### DIFF
--- a/internal/oapispec/openapi3.go
+++ b/internal/oapispec/openapi3.go
@@ -155,20 +155,20 @@ func ffTagHandler(ctx context.Context, route *Route, name string, tag reflect.St
 
 func addCustomType(t reflect.Type, schema *openapi3.Schema) {
 	typeString := "string"
-	if schema.Type == "" {
-		switch t.Name() {
-		case "UUID":
-			schema.Type = typeString
-			schema.Format = "uuid"
-		case "FFTime":
-			schema.Type = typeString
-			schema.Format = "date-time"
-		case "Bytes32":
-			schema.Type = typeString
-			schema.Format = "byte"
-		case "FFBigInt":
-			schema.Type = typeString
-		}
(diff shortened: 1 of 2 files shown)


internal/oapispec/openapi3.go
```diff
@@ -155,20 +155,20 @@ func ffTagHandler(ctx context.Context, route *Route, name string, tag reflect.St
...
+	case "FFTime":
+		schema.Type = typeString
+		schema.Format = "date-time"
+	case "Bytes32":
+		schema.Type = typeString
+		schema.Format = "byte"
+	case "FFBigInt":
+		schema.Type = typeString
+	case "JSONAny":
+		schema.Type = ""
```
peterbroadhurst, line 171: Should this not be: https://swagger.io/docs/specification/data-models/data-types/#any

peterbroadhurst (changes requested):
If possible to use the any type

internal/oapispec/openapi3.go
awrichar: Seems appropriate - I'll look into it

awrichar: I think this is actually what I'm doing. That link says to omit the type if you want to allow "any type". As a reference, for something like a contract `location`, this is what is written (notice the omission of "type" similar to the examples in the docs):

```
location:
  description: If this API is tied to an individual instance of
    a smart contract, this field can include a blockchain specific
    contract identifier. For example an Ethereum contract address,
    or a Fabric chaincode name and channel
```

peterbroadhurst: Thanks @awrichar - makes sense 👍

peterbroadhurst (approved):

PR #737, opened 2022-04-20
Configurable JSON Formatter and Caller Reporting for Logs

### LINKED ISSUE #713: Structured JSON Logging
FireFly is using https://github.com/sirupsen/logrus which is great. However, when integrating with log aggregation systems which allow searching across multiple indexed fields of documents, i.e. OpenSearch, the current experience requires doing inefficient searches of message IDs, etc. within the log message itself. Instead, we can structure the logs as JSON so certain useful / often searched pieces of metadata within logs can be included as separate fields and therefore more easily searchable. Such pieces of metadata can be very high in cardinality which serves as a great compliment to our Prometheus metrics which cannot be as high in cardinality without exploding timeseries costs.

To close this issue, we should achieve two things:

1. Offer new [`log`](https://hyperledger.github.io/firefly/reference/config.html#log) options for enabling JSON logs, which just sets the logrus formatter to the `JSONFormatter` behind the scenes: https://github.com/sirupsen/logrus#environments
2. Refactor log statements throughout the code base to use `WithFields` which will help improve log clarity in the unstructured case, while simultaneously enabling searchable fields when using structured logs: https://github.com/sirupsen/logrus#fields

hfuss:
Closes #713, useful for parsing logs w/ `jq` or searching through logs using log aggregators like Filebeat + OpenSearch

Using a slightly modified config from a default `dev` stack:
```yaml
log:
  level: debug
  reportCaller: true
  json:
    enabled: true
```

and running the image locally:

```shell
make docker
docker run -v ${HOME}/.firefly/stacks/dev/init/config/:/etc/firefly -it hyperledger/firefly -f /etc/firefly/firefly_core_0.yml
```

results in:

```json
{"@timestamp":"2022-04-20T04:53:12.477Z","file":"/firefly/pkg/config/config.go:528","func":"github.com/hyperledger/firefly/pkg/config.SetupLogging","level":"debug","message":"Log level: debug","prefix":"node_0"}
{"@timestamp":"2022-04-20T04:53:12.477Z","file":"/firefly/cmd/firefly.go:105","func":"github.com/hyperledger/firefly/cmd.run","level":"info","message":"Project Firefly","prefix":"node_0"}
{"@timestamp":"2022-04-20T04:53:12.477Z","file":"/firefly/cmd/firefly.go:106","func":"github.com/hyperledger/firefly/cmd.run","level":"info","message":"© Copyright 2021 Kaleido, Inc.","prefix":"node_0"}
{"@timestamp":"2022-04-20T04:53:12.479Z","file":"/firefly/cmd/firefly.go:159","func":"github.com/hyperledger/firefly/cmd.startFirefly","level":"debug","message":"Debug HTTP endpoint listening on localhost:6060","prefix":"node_0"}

```


pkg/config/config.go
```diff
@@ -65,6 +65,20 @@ var (
 	LogMaxAge = ffc("log.maxAge")
 	// LogCompress sets whether to compress backups
 	LogCompress = ffc("log.compress")
+	// LogReportCaller enables the report caller for including the calling file and line number
+	LogReportCaller = ffc("log.reportCaller")
```
hfuss, line 69: Noting that the text formatter we use doesn't currently respect this:

```yaml
log:
  level: debug
  reportCaller: true
  noColor: true
```

```
[2022-04-20T05:04:35.099Z] DEBUG node_0: Log level: debug
[2022-04-20T05:04:35.099Z]  INFO node_0: Project Firefly
[2022-04-20T05:04:35.099Z]  INFO node_0: © Copyright 2021 Kaleido, Inc.
[2022-04-20T05:04:35.100Z] DEBUG node_0: Debug HTTP endpoint listening on localhost:6060
```

so could use input on whether this option should be fall under `json` or not.

docs/reference/config.md
```diff
@@ -517,9 +517,26 @@ nav_order: 3
 |maxAge|The maximum time to retain old log files based on the timestamp encoded in their filename.|[`time.Duration`](https://pkg.go.dev/time#Duration)|`<nil>`
 |maxBackups|Maximum number of old log files to retain|`int`|`<nil>`
 |noColor|Force color to be disabled, event when TTY output is detected|`boolean`|`<nil>`
+|reportCaller|Enables the report caller for including the calling file and line number, as well as the calling function. Only works for JSON formatted logs.|`boolean`|`<nil>`
```
peterbroadhurst, line 520: I wonder if `includeCodeInfo` would be clearer here, or similar?

pkg/config/config.go
peterbroadhurst: See note above on spelling

peterbroadhurst: Does the other logger formatter support it?
e.g. if I set `reportCaller` (or `includeCodeInfo` etc.) do my _non_ JSON logs include it?

peterbroadhurst (commented):
Looks great 👍  - couple of questions on there

docs/reference/config.md
hfuss: Fine with `includeCodeInfo` does sound a bit clearer than `reportCaller` 👍🏻

pkg/config/config.go
hfuss: Not currently no, its unclear if the `prefixed` text formatter we use doesn't support logrus' report caller or if maybe because we're setting `ForceFormatting`. I'll play with it a bit

peterbroadhurst: If it only works for `json`... then best to just pop it under the `json` section I think.

hfuss: In theory it shouldn't only work for `json`, so gonna try to understand why that's not working before giving up and moving it under there. Could definitely see local stacks benefitting from have `includeCodeInfo` enabled with human-readable text logs.

docs/reference/config.md
```diff
@@ -513,13 +513,30 @@ nav_order: 3
 |filename|Filename is the file to write logs to.  Backup log files will be retained in the same directory|`string`|`<nil>`
 |filesize|MaxSize is the maximum size the log file before it gets rotated|[`BytesSize`](https://pkg.go.dev/github.com/docker/go-units#BytesSize)|`<nil>`
 |forceColor|Force color to be enabled, even when a non-TTY output is detected|`boolean`|`<nil>`
+|includeCodeInfo|Enables the report caller for including the calling file and line number, and the calling function. If using text logs, it uses the logrus text format rather than the default prefix format.|`boolean`|`<nil>`
```
hfuss, line 516: Note this causes FireFly to use the logrus text formatter rather than the prefixed logger if JSON logs are not enabled.

peterbroadhurst (approved):
Thanks @hfuss for reconciling the functionality here.

PR #740, opened 2022-04-20
Add first step of FireFly Transaction Manager support - separate URL for submit

peterbroadhurst:
This is the first step, to allow a URL to be specified, and use it only for `SendTransaction`.


peterbroadhurst:
Moving branch

PR #745, opened 2022-04-20
Ensuring Senders Do Not Download Their Own Batches in Response to Pin Events

### LINKED ISSUE #694: Own Identity Claims Stuck in Pending
## Description

When registering orgs and new nodes, we have observed identity claims being stuck in `pending` on the node that registering its org even though registration succeeded and we see message confirmed events in the logs and on other nodes:

Screen Shot 2022-04-11 at 2 38 15 PM
Screen Shot 2022-04-11 at 2 38 29 PM

This had been hopefully fixed in https://github.com/hyperledger/firefly/pull/670 however it does not appear to be since this was observed with a recent release candidate and we can see similar logs still in the registering node:

```
[2022-04-11T16:53:30.270Z] DEBUG poc4-par1-firefly: SQL-> insert INSERT INTO messages dbtx=wKFeQgiZ role=event-manager
[2022-04-11T16:53:30.272Z] DEBUG poc4-par1-firefly: SQL<- inserted sequences=[3 4] dbtx=aiTOEbd5 role=aggregator
[2022-04-11T16:53:30.272Z] DEBUG poc4-par1-firefly: SQL-> commit dbtx=aiTOEbd5 role=aggregator
[2022-04-11T16:53:30.275Z] DEBUG poc4-par1-firefly: SQL<- commit dbtx=aiTOEbd5 role=aggregator
[2022-04-11T16:53:30.276Z]  INFO poc4-par1-firefly: Emitted identity_confirmed event 7a18c9cb-8e64-44b4-84bb-53fb9085e8e5 for ff_system:8e8b6544-df58-43fe-935e-f0078b760e7d (correlator=,topic=ff_definition) dbtx=aiTOEbd5 role=aggregator
[2022-04-11T16:53:30.276Z]  INFO poc4-par1-firefly: Emitted message_confirmed event fc00e5e9-f71e-4c84-b5bf-91b8db3dea9f for ff_system:31d09e35-1b6e-4df5-bb86-483b22fdcfb4 (correlator=,topic=e5b3a0483b8312f9f68725e69cfadb0494fb9f38290be911d546e2e7abf40884) dbtx=aiTOEbd5 role=aggregator
[2022-04-11T16:53:30.276Z] DEBUG poc4-par1-firefly: SQL insert failed (conflictEmptyRequested=true): FF10375: Failed to retrieve sequence for insert row 1 (could mean duplicate insert) sql=[ INSERT INTO messages (id,cid,mtype,author,key,created,namespace,topics,tag,group_hash,datahash,hash,pins,state,confirmed,tx_type,batch_id) VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15,$16,$17)  ON CONFLICT DO NOTHING RETURNING seq ]: FF10375: Failed to retrieve sequence for insert row 1 (could mean duplicate insert) dbtx=wKFeQgiZ role=event-manager
[2022-04-11T16:53:30.276Z] DEBUG poc4-par1-firefly: Batch message insert optimization failed for batch '175adf2a-df06-4c5b-95af-12085bdac7c0': FF10116: Database insert failed: FF10375: Failed to retrieve sequence for insert row 1 (could mean duplicate insert) dbtx=wKFeQgiZ role=event-manager
[2022-04-11T16:53:30.276Z] DEBUG poc4-par1-firefly: SQL-> update: UPDATE messages SET dbtx=wKFeQgiZ role=event-manager
```

## Other Info

FireFly Version: `1.0.0-rc3`
Ethconnect Version: `v3.1.5`
Blockchain Provider: `Quorum v21.10.0 (Geth v1.9.25)`

## Logs

[poc4-par1-registration.log](https://github.com/hyperledger/firefly/files/8467463/poc4-par1-registration.log)

hfuss:
Fixes #694, we were seeing the sender's aggregator process pins before a batch was ever persisted. So although message confirmed events were still being sent by the dispatchers, the database updates to mark the message pending (what is done as part of persisting a batch) were occurring after the messages were already confirmed.

When persisting the batch, we guarantee that the batch is not cached until its successfully persisted:
```
	valid, err = em.validateAndPersistBatchContent(ctx, batch)
	if err != nil || !valid {
		return nil, valid, err
	}
	em.aggregator.cacheBatch(em.aggregator.getBatchCacheKey(persistedBatch.ID, persistedBatch.Hash), persistedBatch, manifest)
	return persistedBatch, true, err
```

so if the aggregator repolls any pins who's batch is not in the cache this should avoid the issue. Other option is instead of repolling, letting the aggregators rewind handle processing the pins.

See https://github.com/hyperledger/firefly/pull/745/files#discussion_r854706368 for a more up-to-date explanation of the cause and fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,8 @@ Other repositories you might be interested in containing those microservice comp
   - Public EVM compatible chains: Learn more in the [documentation](https://hyperledger.github.io/firefly)
 - Corda connector starter: https://github.com/hyperledger/firefly-cordaconnect
   - CorDapp specific customization is required
-- Firefly Performance CLI: https://github.com/hyperledger/firefly-perf-cli
+- FireFly Performance CLI: https://github.com/hyperledger/firefly-perf-cli
+- Helm Charts for Deploying to Kubernetes: https://github.com/hyperledger/firefly-helm-charts
 
 ## FireFly Core code hierarchy
 
--- a/internal/events/batch_pin_complete.go
+++ b/internal/events/batch_pin_complete.go
@@ -70,9 +70,12 @@ func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockch
 				return err
 			}
 
-			_, err := em.database.GetBatchByID(ctx, batchPin.BatchID)
+			batch, err := em.database.GetBatchByID(ctx, batchPin.BatchID)
+			if err != nil {
+				return err
+			}
 			// Kick off a download for broadcast batches if the batch isn't already persisted
-			if !private && err != nil {
+			if !private && batch == nil {
 				if err := em.sharedDownload.InitiateDownloadBatch(ctx, batchPin.Namespace, batchPin.TransactionID, batchPin.BatchPayloadRef); err != nil {
 					return err
 				}
--- a/internal/events/batch_pin_complete.go
+++ b/internal/events/batch_pin_complete.go
@@ -70,7 +70,10 @@ func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockch
 				return err
 			}
 
-			batch, err := em.database.GetBatchByID(ctx, batchPin.BatchID)
+			batch, _, err := em.aggregator.GetBatchForPin(ctx, &fftypes.Pin{
+				Batch:     batchPin.BatchID,
+				BatchHash: batchPin.BatchHash,
+			})
 			if err != nil {
 				return err
 			}
--- a/internal/events/batch_pin_complete.go
+++ b/internal/events/batch_pin_complete.go
@@ -35,6 +35,7 @@ func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockch
 		log.L(em.ctx).Errorf("Invalid BatchPin transaction - ID is nil")
 		return nil // move on
 	}
+
 	if err := fftypes.ValidateFFNameField(em.ctx, batchPin.Namespace, "namespace"); err != nil {
 		log.L(em.ctx).Errorf("Invalid transaction ID='%s' - invalid namespace '%s': %a", batchPin.TransactionID, batchPin.Namespace, err)
 		return nil // move on
@@ -68,8 +69,10 @@ func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockch
 			if err := em.persistContexts(ctx, batchPin, signingKey, private); err != nil {
 				return err
 			}
-			// Kick off a download for broadcast batches
-			if !private {
+
+			_, err := em.database.GetBatchByID(ctx, batchPin.BatchID)
+			// Kick off a download for broadcast batches if the batch isn't already persisted
+			if !private && err != nil {
 				if err := em.sharedDownload.InitiateDownloadBatch(ctx, batchPin.Namespace, batchPin.TransactionID, batchPin.BatchPayloadRef); err != nil {
 					return err
 				}
--- a/internal/events/event_poller.go
+++ b/internal/events/event_poller.go
@@ -217,7 +217,9 @@ func (ep *eventPoller) eventLoop() {
 		}
 
 		// Once we run out of events, wait to be woken
-		if !repoll {
+		if repoll {
+			time.Sleep(10 * time.Millisecond)
+		} else {
 			if ok := ep.waitForShoulderTapOrPollTimeout(eventCount); !ok {
 				return
 			}
--- a/internal/events/event_poller.go
+++ b/internal/events/event_poller.go
@@ -217,9 +217,7 @@ func (ep *eventPoller) eventLoop() {
 		}
 
 		// Once we run out of events, wait to be woken
-		if repoll {
-			time.Sleep(10 * time.Millisecond)
-		} else {
+		if !repoll {
 			if ok := ep.waitForShoulderTapOrPollTimeout(eventCount); !ok {
 				return
 			}
--- a/internal/events/persist_batch.go
+++ b/internal/events/persist_batch.go
@@ -246,6 +246,7 @@ func (em *eventManager) persistBatchContent(ctx context.Context, batch *fftypes.
 	// If we are sure we wrote the batch, then we do a cached lookup of each in turn - which is efficient
 	// because all of those should be in the cache as we wrote them recently.
 	if em.sentByUs(ctx, batch) {
+		log.L(ctx).Debugf("Batch %s sent by us", batch.ID)
 		allStored, err := em.verifyAlreadyStored(ctx, batch)
 		if err != nil {
 			return false, err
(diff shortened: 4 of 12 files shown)


internal/events/persist_batch.go
```diff
@@ -298,6 +299,8 @@ func (em *eventManager) persistBatchContent(ctx context.Context, batch *fftypes.
 				log.L(ctx).Errorf("Failed to insert message entry %d in batch '%s': %s", i, batch.ID, err)
 				return false, err // a persistence failure here is considered retryable (so returned)
 			}
+			mm := matchedMsgs[i]
+			em.data.UpdateMessageCache(mm.message, mm.data)
```
hfuss, line 303: We have a postcommit hook for updating the messages via the optimized insert, but if we resort to individual upserts because the optimization failed we should update the cache after those individual upserts is my understanding.

hfuss:
Repolling is proving to bit inefficient and seems to be causing other problems, so need to approach this a bit differently.

internal/events/event_poller.go
```diff
@@ -217,7 +217,9 @@ func (ep *eventPoller) eventLoop() {
 		}
 
 		// Once we run out of events, wait to be woken
-		if !repoll {
+		if repoll {
+			time.Sleep(10 * time.Millisecond)
```
hfuss, line 221: Without sleeping, if using `repoll` its a hot spin. Is that intended? If there should be some sort of repoll delay, I'm assuming that should be configurable?

internal/events/batch_pin_complete.go
```diff
@@ -68,8 +69,13 @@ func (em *eventManager) BatchPinComplete(bi blockchain.Plugin, batchPin *blockch
...
 			}
-			// Kick off a download for broadcast batches
-			if !private {
+
+			batch, err := em.database.GetBatchByID(ctx, batchPin.BatchID)
+			if err != nil {
+				return err
+			}
+			// Kick off a download for broadcast batches if the batch isn't already persisted
+			if !private && batch == nil {
```
hfuss, line 78: Alright, so the issue was not actually that the sender's aggregator was incorrectly receiving and confirming the pin before the batch was persisted - the issue was that the batch was being re-persisted because we were downloading the batch from IPFS as the sender which was incorrect.

The sender persists the batch before uploading to IPFS, so theres no need for it to unnecessarily download from IPFS and then re-persist the batch. So after ensuring the pins are persisted when receiving the `BatchPinComplete` event from the chain, we now do a check to see if the batch is already persisted which avoids the issue.

The reason why this didn't affect other batch / messages sent after reigsteration was because of the `sentByUs` function being used in `persist_batch.go` which used the `Node` of the batch to check if the node authored it. For registration batches, the node hasn't been registered so that didn't protect us.

The extra DB lookup prevents wasting a download worker for downloading batches when we are the sender, as well as if the batch pin events from etheconnect are ever replayed (i.e. the eventstream is deleted) which are both added benefits with this fix.

Only question remains: is this the most appropriate way to look up if the batch is already persisted? Should it be going through the operator manager or checking the cache first? @peterbroadhurst 

Want to make sure this small bit of code is correct before updating the unit tests for it.

internal/events/persist_batch.go
peterbroadhurst: Great point 👍 
However, this should be a post-commit on the `UpsertMessage` call if that's ok, because we don't want the cache to be present (on the non-sending node) and visible to the aggregator until after we've successfully committed the write.

peterbroadhurst: (You'll likely have to add the `hooks` option to the call in `pkg.database.Plugin` and `internal.database`)

internal/events/event_poller.go
peterbroadhurst: I'd like to understand this one a bit more. `repoll` is intended to be returned when the pluggable function knows there's more to do, so it shouldn't be a hot spin. But if you've found an implementation of `newEventsHandler` that has a hot spin condition would love to talk it through.

I believe the most common case to want to immediately read again, is if you've read a full page - so there's more potentially there that simply were beyond your page boundary.

hfuss: > But if you've found an implementation of newEventsHandler that has a hot spin condition would love to talk it through.

Only one that I created bc I was mis-using `repoll` 😄 . That makes sense, thanks for explaining that. Will remove this.

internal/events/batch_pin_complete.go
peterbroadhurst: Great investigation 👍 

We have a batch cache, that should be used here rather than going directly to the DB.
See `GetBatchForPin`.

One potential thing I've been thinking through here - we have considered the database design of FireFly such that it should be pluggable to document databases without SQL style transactions in the future.
I was digging around to think about if there's a possibility that because we write the batch (which just has the manifest) to the DB before all the contents whether there's a possibility for an issue when one of those DBs come along. Basically whether we should swap the order of `validateAndPersistBatchContent` and `UpsertBatch` in `persistBatch` for future proofing to when those DBs come along.

peterbroadhurst: e.g. do this line first:
https://github.com/hyperledger/firefly/blob/41a5a155121aaeeb076a245399f3cd1e397ba429/internal/events/persist_batch.go#L82
Then this line second:
https://github.com/hyperledger/firefly/blob/41a5a155121aaeeb076a245399f3cd1e397ba429/internal/events/persist_batch.go#L72

hfuss: > We have a batch cache, that I think should be used here rather than going directly to the DB.
See GetBatchForPin.

Makes sense, only concern I have is if somehow the batch is not in the cache but its been persisted when receiving the pin event. Should it be check cache, if not in cache check DB?

> Basically whether we should swap the order of validateAndPersistBatchContent and UpsertBatch in persistBatch for future proofing to when those DBs come along.

Sounds reasonable to me! Will go ahead and add that in as part of this.

hfuss: Saw that `GetBatchForPin` falls through the DB in the event of an empty cache so this was exactly was what's needed and is working.

hfuss:
Testing in a multi-orgs / multi-node network all orgs / nodes registered successfully and messages are properly reporting confirmed:

Screen Shot 2022-04-21 at 10 45 33 PM
Screen Shot 2022-04-21 at 10 45 42 PM

README.md
```diff
@@ -68,7 +68,8 @@ Other repositories you might be interested in containing those microservice comp
   - Public EVM compatible chains: Learn more in the [documentation](https://hyperledger.github.io/firefly)
 - Corda connector starter: https://github.com/hyperledger/firefly-cordaconnect
   - CorDapp specific customization is required
-- Firefly Performance CLI: https://github.com/hyperledger/firefly-perf-cli
+- FireFly Performance CLI: https://github.com/hyperledger/firefly-perf-cli
+- Helm Charts for Deploying to Kubernetes: https://github.com/hyperledger/firefly-helm-charts
```
peterbroadhurst, line 72: Thanks for spotting this 👍

peterbroadhurst (approved):
Super, thanks @hfuss for rounding this all out. Looks great.

PR #746, opened 2022-04-20
Resolve pool activation operations more cleanly

### LINKED ISSUE #743: Pool activation happens more than once
Token pool activation is sometimes happening twice, and only one of the operations gets resolved.

This is partly because activation operations are resolved only by TX ID, not by the actual operation ID.
https://github.com/hyperledger/firefly/blob/5afbecf2563d290f014e45b2542e037947ecd3ec/internal/events/token_pool_created.go#L68

Need to revisit this flow.

awrichar:
Fixes #743

### DIFF
--- a/internal/events/token_pool_created.go
+++ b/internal/events/token_pool_created.go
@@ -61,11 +61,6 @@ func (em *eventManager) confirmPool(ctx context.Context, pool *fftypes.TokenPool
 		}
 		em.emitBlockchainEventMetric(ev)
 	}
-	if op, err := em.findTXOperation(ctx, pool.TX.ID, fftypes.OpTypeTokenActivatePool); err != nil {
-		return err
-	} else if op == nil {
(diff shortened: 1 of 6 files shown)


internal/events/token_pool_created.go
```diff
@@ -65,8 +65,6 @@ func (em *eventManager) confirmPool(ctx context.Context, pool *fftypes.TokenPool
 		return err
```
awrichar, line 65: It occurs to me that this operation query is probably not needed at all now. I think it was only ever here to do the resolve.

peterbroadhurst: Agreed. Will you remove it in this PR?

peterbroadhurst (approved):
Looks good to me. Approved, but with a question on whether you want to remove that now redundant DB call before merging.

internal/events/token_pool_created.go
awrichar: Yes, will push a quick change to do that.

PR #747, opened 2022-04-20
Add table name to debug on every SQL call

peterbroadhurst:
Currently we take the first three words of the SQL to try and provide a debug statement that's useful, without printing the massive bundle of SQL. However, that is insufficient as for something like `SELECT` it just contains the first two fields from the select.

This PR introduces an explicit table name being passed on every call and put in the logs.

I did find what I think was one bug where `GetContractAPIs` was using the `contract_interfaces` table


internal/database/sqlcommon/contractapis_sql.go
```diff
@@ -173,7 +175,7 @@ func (s *SQLCommon) GetContractAPIs(ctx context.Context, ns string, filter datab
 		apis = append(apis, api)
 	}
 
-	return apis, s.queryRes(ctx, tx, "contract_interfaces", fop, fi), err
```
awrichar, line 176: 😳

internal/database/sqlcommon/sqlcommon.go
```diff
@@ -195,7 +177,7 @@ func (s *SQLCommon) queryTx(ctx context.Context, tx *txWrapper, q sq.SelectBuild
 	if err != nil {
 		return nil, tx, i18n.WrapError(ctx, err, coremsgs.MsgDBQueryBuildFailed)
 	}
-	l.Debugf(`SQL-> query: %s`, shortenSQL(sqlQuery))
+	l.Debugf(`SQL-> query %s`, table)
 	l.Tracef(`SQL-> query args: %+v`, args)
```
awrichar, line 181: We logs the args at trace level... but shouldn't we log the full SQL statement as well?

internal/database/sqlcommon/sqlcommon.go
```diff
@@ -226,12 +208,12 @@ func (s *SQLCommon) countQuery(ctx context.Context, tx *txWrapper, tableName str
...
 	}
-	q := sq.Select(fmt.Sprintf("COUNT(%s)", countExpr)).From(tableName).Where(fop)
+	q := sq.Select(fmt.Sprintf("COUNT(%s)", countExpr)).From(table).Where(fop)
 	sqlQuery, args, err := q.PlaceholderFormat(s.features.PlaceholderFormat).ToSql()
 	if err != nil {
 		return count, i18n.WrapError(ctx, err, coremsgs.MsgDBQueryBuildFailed)
 	}
-	l.Debugf(`SQL-> count query: %s`, shortenSQL(sqlQuery))
+	l.Debugf(`SQL-> count query %s`, table)
 	l.Tracef(`SQL-> count query args: %+v`, args)
```
awrichar, line 217: same comment

internal/database/sqlcommon/sqlcommon.go
```diff
@@ -266,25 +248,25 @@ func (s *SQLCommon) queryRes(ctx context.Context, tx *txWrapper, tableName strin
...
 	l := log.L(ctx)
 	q, useQuery := s.provider.ApplyInsertQueryCustomizations(q, requestConflictEmptyResult)
 
 	sqlQuery, args, err := q.PlaceholderFormat(s.features.PlaceholderFormat).ToSql()
 	if err != nil {
 		return i18n.WrapError(ctx, err, coremsgs.MsgDBQueryBuildFailed)
 	}
-	l.Debugf(`SQL-> insert %s`, shortenSQL(sqlQuery))
+	l.Debugf(`SQL-> insert %s`, table)
 	l.Tracef(`SQL-> insert query: %s (args: %+v)`, sqlQuery, args)
```
awrichar, line 270: and here

awrichar (approved):
Looks great. Just one comment on what's now logged (same comment in multiple places).

internal/database/sqlcommon/sqlcommon.go
peterbroadhurst: Thanks for catching that 👍

PR #748, opened 2022-04-21
Format JSON errors from connectors

awrichar:
```
Parse a JSON error of the form:
  {"error": "Bad Request", "message": "Field 'x' is required"}
into a message of the form:
  "Bad Request: Field 'x' is required"
```

See https://github.com/hyperledger/firefly/issues/753

### DIFF
--- a/internal/tokens/fftokens/fftokens.go
+++ b/internal/tokens/fftokens/fftokens.go
@@ -123,6 +123,11 @@ type tokenApproval struct {
 	Config      fftypes.JSONObject `json:"config"`
 }
 
+type tokenError struct {
+	Error   string `json:"error,omitempty"`
+	Message string `json:"message,omitempty"`
+}
+
 func (ft *FFTokens) Name() string {
 	return "fftokens"
 }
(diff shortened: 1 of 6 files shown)


internal/tokens/fftokens/fftokens.go
```diff
@@ -477,6 +477,27 @@ func (ft *FFTokens) eventLoop() {
 	}
 }
 
+// Parse a JSON error of the form:
+//   {"error": "Bad Request", "message": "Field 'x' is required"}
+// into a message of the form:
+//   "Bad Request: Field 'x' is required"
+func wrapError(ctx context.Context, res *resty.Response, err error) error {
```
peterbroadhurst, line 484: There is a `SetError` function on the Resty client, where you can pass in a `fftypes.RESTError` object (or a more customized one) as a way to not need some of this complexity. Did you see that as an option?

awrichar: Did not know about that - I'll give it a look. I do want to fall back to the old behavior if the error doesn't conform to the expected JSON format from the connector though (ie in case nginx or something hops in the middle and spits back some text).

awrichar:
Seems like maybe the ethconnect and fabconnect plugins should get similar treatment in order to parse the expected error formats from there as well.

awrichar:
Added formatting for ethconnect and fabconnect errors. Tested via unit test and manual test on a real stack to ensure that errors which match the expected formatting _and_ errors which don't match it will both show up in a suitable manner.

peterbroadhurst (approved):
This is great, thanks for chasing this all the way through 👍

PR #749, opened 2022-04-21
Misc cleanup for contract listeners

awrichar:


peterbroadhurst (approved):
Just noting as it wasn't in the PR description:

This PR means that now we have a `signature` field on all events, which is computed at the point of query.

I reviewed the APIs, and there's no place where you can query an `event` underneath an FFI directly, or filter on them. So this one place is the only one where we need to generate the signature for query.

PR #750, opened 2022-04-21
Rename docs pages for better hierarchy

nguyer:
This PR renames some of the files in the subdirectories of the docs site. It improves the page hierarchy, but it does run the risk of breaking some existing links that we may have.

For example:
https://hyperledger.github.io/firefly/gettingstarted/gettingstarted.html
Becomes:
https://hyperledger.github.io/firefly/gettingstarted

This also fixes the links on the main page that are currently broken, because they used relative links to directories, and there was no index.html file in those directories previously.


nguyer:
FYI @keeganhrobinson just wanted to give you a heads up in case you are aware of things that link to the pages I renamed.

keeganhrobinson:
Ah, cool. Thanks @nguyer. I'll run some link checkers on our other sites after this goes out and clean up anything that pops up.

peterbroadhurst (approved):
👍

PR #752, opened 2022-04-21
Remove server-side HTML sanitization of error messages

awrichar:
Errors may be displayed or used in many contexts. If they are displayed in HTML,
it should be the responsibility of the HTML client to sanitize them. Doing the
sanitization server-side makes the messages less readable in all non-HTML
contexts.

I want to be conscious about XSS and security, but I'd say that doing
an HTML-escaping pass on these errors before writing them to the database
is putting the processing in the wrong place. Characters such as `<` and `"`
become much less readable when they're converted to HTML entities.

And arguably this encourages bad design on an HTML client - as the client page would have
to pipe the values into the HTML _without_ sanitizing them client-side, in order
to have the HTML entities render back into readable characters.

One good discussion of this (anti-)pattern:
https://security.stackexchange.com/questions/42498/which-is-the-best-way-to-sanitize-user-input-in-php/42521#42521

Also see https://github.com/hyperledger/firefly/issues/753


peterbroadhurst (approved):
Ack. Don't see any strong counter points from review.
Pushing this responsibility out of the middleware tier seems like the right action.

PR #755, opened 2022-04-22
Firefly E2E Test updates

calbritt:
- Refactor tests to specify a Firefly namespace to run tests against 
- Create namespace test suite to use without identity tests 
- Fix string comparison in objectCheck
- Remove TestE2EContractEvents() test 
- Set pool name to random int instead of using len - len causes issues when completing more than 25 test runs on the same stack and will always make the poolname pool26


test/e2e/e2e_test.go
```diff
@@ -425,7 +427,7 @@ func checkObject(t *testing.T, expected interface{}, actual interface{}) bool {
 		}
 	default:
 		expectedString, expectedIsString := expected.(string)
-		actualString, actualIsString := expected.(string)
```
nguyer, line 428: 🫢 Thank you for fixing this!

test/e2e/e2e_test.go
```diff
@@ -240,6 +233,15 @@ func beforeE2ETest(t *testing.T) *testState {
 		}
 	}
 
+	// If no namespace is set to run tests against, use the default namespace
+	if os.Getenv("NAMESPACE") != "" {
+		namespace := os.Getenv("NAMESPACE")
+		CreateNamespaces(ts.client1, namespace)
+		CreateNamespaces(ts.client2, namespace)
+	} else {
+		os.Setenv("NAMESPACE", "default")
```
nguyer, line 242: Rather than setting an environment variable here, and always reading from an environment variable, maybe we should set a field on the `TestState`?

test/e2e/tokens_test.go
```diff
@@ -143,7 +145,7 @@ func (suite *TokensTestSuite) TestE2EFungibleTokensAsync() {
 			To:     suite.testState.org2key.Value,
 			Amount: *fftypes.NewFFBigInt(1),
 			From:   suite.testState.org1key.Value,
-			Key:    suite.testState.org2key.Value,
```
nguyer, line 146: @awrichar Can you take a quick look at this? If we're transferring tokens from `org1` to `org2` using `org2`'s signing key, shouldn't that have failed before?

nguyer (commented):
One minor suggestion, and a question for Andrew. Also it looks like there is a failing test still.

test/e2e/e2e_test.go
calbritt: will do!

test/e2e/tokens_test.go
peterbroadhurst: We're using a single blockchain node, with an embedded signer, in the CLI dev environment created for these tests. So the only reason this would have failed is if `org2` wasn't authorized to transfer the tokens... which it was a few lines earlier:
https://github.com/hyperledger/firefly/blob/5e1120d852aadbf75e37efdab8ea0c9b4f8c9766/test/e2e/tokens_test.go#L101-L108

So I think maybe this was _correct_ in the test before, and might just have needed a comment @calbritt

peterbroadhurst: Thanks for making the change @calbritt to keep the original test, but change the API endpoint it uses. So it signs with `org2key` as originally, but uses the org2 node to do it - important when using the E2E test in full k8s deployments against a blockchain with per-member key management.

peterbroadhurst (approved):

PR #756, opened 2022-04-25
Gauge Metrics for Network Identities

hfuss:
Adds a background loop to the `networkmap` manager which periodically fetches all identities, orgs, and nodes and then updates gauges to reflect their current counts.

The intent is that a network operator aggregating metrics could observe the number of identities / nodes / orgs according to each org / node and ensure they are in-sync to detect any issues. Failure to fetch the identities results in the gauge returning `-1`.

This is different from our other counters which are concerned with reflecting the number of times the FireFly process itself has processed a certain event or performed a certain operation. These gauges are meant to provide a Prometheus-native way to observe some of the database state.

### DIFF
--- a/internal/apiserver/route_get_identity_did.go
+++ b/internal/apiserver/route_get_identity_did.go
@@ -39,6 +39,6 @@ var getIdentityDID = &oapispec.Route{
 	JSONOutputValue: func() interface{} { return &networkmap.DIDDocument{} },
 	JSONOutputCodes: []int{http.StatusOK},
 	JSONHandler: func(r *oapispec.APIRequest) (output interface{}, err error) {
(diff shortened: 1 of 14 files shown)


hfuss:
Talking w/ @peterbroadhurst we recognize this is a monitoring smell. Needing special goroutines or callbacks w/in FireFly itself to query state stored in the DB so that it can be consumed via Prom isn't a FireFly problem. Its a more generic problem of monitoring application state via REST and reflecting those as metrics. We'll work on either finding an existing tool that does that (think https://github.com/percona/mongodb_exporter vs https://github.com/raffis/mongodb-query-exporter) or perhaps write our own eventually.

If its a separate tool with separate Prometheus metrics and `ServiceMonitor` then we also can configure the scrape interval to be significantly less often.

PR #763, opened 2022-04-25
Version Command

### LINKED ISSUE #132: Add version Command to Core Binary

hfuss:
Closes #132, resurrects #586. Uses new preferred way of versioning with module fields in the `cmd` package rather than an `internal` package, see https://github.com/hyperledger/firefly-cli/pull/164 for original implementation of this approach.

Example "default" usage:
```
make build
./firefly version
{
  "Version": "(devel)",
  "Date": "2022-04-25T19:52:28Z",
  "License": "Apache-2.0"
}

./firefly version -o yaml
Date: "2022-04-25T19:52:28Z"
License: Apache-2.0
Version: (devel)

./firefly version --short
(devel)

make docker
docker run -it --entrypoint /bin/sh hyperledger/firefly
/firefly # firefly version
{
  "Version": "(devel)",
  "Date": "2022-04-25T19:45:02Z",
  "License": "Apache-2.0"
}
/firefly # firefly version -o yaml
Date: "2022-04-25T19:45:02Z"
License: Apache-2.0
Version: (devel)

/firefly # firefly version --short
(devel)

```


peterbroadhurst (commented):
Really pleased to see this PR - thanks @hfuss 

Wonder if this file would be helpful to add too?
https://github.com/hyperledger/firefly-transaction-manager/blob/main/cmd/version_test.go

Note small refactor to allow testing:
https://github.com/hyperledger/firefly-transaction-manager/blob/281f7a4c1a5c2c0ebd5f1c4b49400d25565f39b1/cmd/version.go#L45-L49

peterbroadhurst (approved):
👍

Issue #773, opened 2022-04-27
Missing message_id in contractapis


mfolnovic:
Hello,

I'm running Firefly v1.0.0-rc.5 under PostgreSQL.

While creating API, I get an error in logs:

`[2022-04-27T13:29:17.185Z] ERROR: SQL query failed: pq: column "message_id" does not exist sql=[ SELECT id, interface_id, location, name, namespace, message_id FROM contractapis WHERE (namespace = $1 AND name = $2) ] dbtx=D_Ys4hok httpreq=6D4zMeUK req=8_Wrk2Xg`

It seems there's a small difference between SQLite and PostgreSQL migrations, where SQLite's migrations have column message_id while PostgreSQL's migrations don't:

https://github.com/hyperledger/firefly/blob/main/db/migrations/sqlite/000052_create_contractapis_table.up.sql
https://github.com/hyperledger/firefly/blob/main/db/migrations/postgres/000052_create_contractapis_table.up.sql

nguyer:
@mfolnovic Thanks for reporting this. This issue has been fixed in https://github.com/hyperledger/firefly/pull/774

PR #390, opened 2022-01-07
Add postgres migrations for onchain-logic

nguyer:
Signed-off-by: Nicko Guyer <nicko.guyer@kaleido.io>

PR #774, opened 2022-04-27
Add E2E test for contract APIs and fix postgres migration

nguyer:
This PR fixes a PostgreSQL migration issue and adds an additional E2E test to cover that scenario


hfuss (approved):

db/migrations/postgres/000086_alter_contractapis.up.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE contractapis ADD COLUMN message_id UUID NOT NULL;
```
awrichar, line 2: This will fail for existing deployments, right? Need to give a default or make it nullable.

test/e2e/ethereum_contract_test.go
```diff
@@ -247,3 +247,57 @@ func (suite *EthereumContractTestSuite) TestFFIInvokeMethod() {
...
+		"output": map[string]interface{}{
+			"_value": "42",
+			"_from":  suite.testState.org1key.Value,
+		},
+		"listener": listener.ID.String(),
+	}
+	event := waitForContractEvent(suite.T(), suite.testState.client1, received1, match)
+	assert.NotNil(suite.T(), event)
+	assert.NoError(suite.testState.t, err)
+	fmt.Println(createContractAPIResult)
```
awrichar, line 294: Maybe should use logging instead of a raw print.

awrichar (changes requested):
Couple of changes to consider.

db/migrations/postgres/000086_alter_contractapis.up.sql
nguyer: It looks like this has been missing since the first release that it appeared in: https://github.com/hyperledger/firefly/pull/390/files#diff-b20af3fc418d53f1d729cf53470d90ee03dfeb5ba2e1400650036c120aa107e5

test/e2e/ethereum_contract_test.go
nguyer: Actually this was just left over from debugging and can be removed

awrichar (approved):

nguyer:
I forgot to mention in my PR description that this PR also resolves https://github.com/hyperledger/firefly/issues/773. The fix is now available in https://github.com/hyperledger/firefly/releases/tag/v1.0.0-rc.7

PR #780, opened 2022-05-01
Token transfer/approval protocolId should be unique for each connector

awrichar:


nguyer (approved):
Changes look good, and I think you just need to merge `main` back into this branch to get a passing build now

PR #789, opened 2022-05-03
Remove non-definition items from DefinitionHandler

### LINKED ISSUE #614: Rename or split up DefinitionHandlers
At some point in the past, `SystemHandlers` was renamed to `DefinitionHandlers` - because the primary thing it does is handling system definition broadcasts.

https://github.com/hyperledger/firefly/blob/ab26d06a665472d426abd8608d9688b453e78f6d/internal/definitions/definition_handler.go#L37

Since that time, I've realized that this was probably incorrect, because it doesn't _only_ handle system definition broadcasts. It does handle a few other random tasks related to groups and replies. It should probably be split up or maybe just renamed back to what it was before...

awrichar:
Resolves #614


peterbroadhurst (approved):
The reason for the original split out didn't come back to me while doing the review 🤷 
This is much cleaner.

PR #790, opened 2022-05-04
Split definition methods out of Broadcast Manager into new Definition Sender

awrichar:
Not totally crazy about the names `defsender.Sender` and `defhandler.DefinitionHandler`, but they need to be separate packages and this is where I landed...


internal/defsender/manager.go
```diff
@@ -53,57 +88,36 @@ func (bm *broadcastManager) BroadcastIdentityClaim(ctx context.Context, ns strin
 	return bm.broadcastDefinitionCommon(ctx, ns, def, signingIdentity, tag, waitConfirm)
 }
 
-func (bm *broadcastManager) broadcastDefinitionCommon(ctx context.Context, ns string, def fftypes.Definition, signingIdentity *fftypes.SignerRef, tag string, waitConfirm bool) (*fftypes.Message, error) {
+func (bm *definitionSender) broadcastDefinitionCommon(ctx context.Context, ns string, def fftypes.Definition, signingIdentity *fftypes.SignerRef, tag string, waitConfirm bool) (*fftypes.Message, error) {
```
awrichar, line 91: Just want to note that this is probably the only method with functional changes beyond simple renaming. Due to the new separation between broadcast manager and definition sender, less of the data processing is done directly. We simply build a `MessageInOut` with `InlineData`, similar to any other broadcast. It feels clean overall, but I just wanted to point it out among the noise.

peterbroadhurst (commented):
Just checking @awrichar - could `DefinitionSender` go into the `broadcast` package? (hence leaving the `definitions` package with its current name, but only containing the handlers)

awrichar:
Yea, it probably could. But now I'm thinking it's best to pause on this until FIR-12 moves along a little further. The thought is that soon definitions will _not_ necessarily be a broadcast, so they don't totally belong in the broadcast package. Going to wait until that next piece of work queues up though before doing any more shuffling here.

peterbroadhurst:
Converted to draft based on last comment

awrichar:
Replaced by #878

PR #791, opened 2022-05-04
Refactor to move common utility code out to firefly-common

peterbroadhurst:
- Moves base `pkg/fftypes` (time, numeric, bytes etc.) out to `firefly-common`
- Moves all FireFly Core data types from `pkg/fftypes` to `pkg/core`
- Removes the following packages that all moved to `firefly-common`:
  - `pkg/config`
  - `pkg/ffresty`
  - `pkg/httpserver`
  - `pkg/i18n`
  - `pkg/log`
  - `pkg/retry`
  - `pkg/wsclient`
- Only functional change: `api.shutdownTimeout` moved to be consistent with the other `httpserver` config, so is now split into three separate configurations:
   - `api.shutdownTimeout`
   - `admin.shutdownTimeout`
   - `metrics.shutdownTimeout`
- I decided to keep `cors.` as a global config shared across `api`/`admin`/`metrics` - which means initializing the HTTP Server now passes in two separate config prefixes


Makefile
```diff
@@ -24,14 +24,17 @@ ${MOCKERY}:
 		$(VGO) install github.com/vektra/mockery/cmd/mockery@latest
 ${LINT}:
 		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+ffcommon:
+		$(eval FFCOMMON_PATH := $(shell $(VGO) list -f '{{.Dir}}' github.com/hyperledger/firefly-common/pkg/wsclient))
```
awrichar, line 28: nit: this is more accurately `WSCLIENT_PATH`

pkg/core/batch.go
```diff
@@ -14,11 +14,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package fftypes
+package core
```
awrichar, line 17: Why rename from fftypes to core? Is it just to differentiate it from the base fftypes in common?

test/e2e/identity_test.go
```diff
@@ -8,8 +8,8 @@
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// Unless required by applicable law or agreed nto in writing, software
+// distributed under the License is distributed o an "AS IS" BASIS,
```
awrichar, line 12: Randomly messed up a header here

pkg/core/nexthash.go
```diff
@@ -14,9 +14,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
awrichar, line 16: nit: just noticed that this file is called "nexthash" but the type is called "NextPin"

awrichar (approved):

pkg/core/batch.go
peterbroadhurst: Yep, lots of files pull in both - so this avoids the need to alias.

PR #792, opened 2022-05-04
Allow namespace to be omitted from URLs

awrichar:
Proposal to tweak our route generation so that namespaced routes will actually generate _two_ versions of the route:
* a base route like `/data` that implies use of the default namespace
* a namespaced route like `/namespaces/{ns}/data` that allows you to specify the namespace

To reduce clutter on the Swagger UI page, routes will also now be grouped into 3 category buckets. I'm open to names for these categories, but I've proposed "Global", "Default Namespace", and "Non-Default Namespace".

The main functional changes are in `internal/apiserver/routes.go`.

---

Swagger_UI

---

Swagger_UI

---

Swagger_UI

---

Swagger_UI


peterbroadhurst (approved):
This is great! You've just saved me loads of `curl` typing 🙇

PR #800, opened 2022-05-09
Move tokens config under `plugins` key

shorsher:
Updating the tokens configuration to be aligned with the format proposed
in `FIR-12: Namepsace Isolation`. The config is now under a top-level key called `plugins`,
with a slightly updated format. The existing config is now marked as
deprecated, but still supported.

There is not support for having both configration structures defined, the new
`plugins` config structure takes precedence over the deprecated one.

Example of the new configuration:
```
plugins:
  tokens:
      - type: fftokens
        name: erc20_erc721
        fftokens:
          url: http://tokens_0_0:3000
```


internal/orchestrator/orchestrator.go
```diff
@@ -67,7 +67,9 @@ var (
 	// For backward compatibility with the old "publicstorage" prefix
 	publicstorageConfig = config.NewPluginConfig("publicstorage")
 	dataexchangeConfig  = config.NewPluginConfig("dataexchange")
-	tokensConfig        = config.NewPluginConfig("tokens").Array()
+	// tokenConfig is deprecated in favor of tokensPluginConfig
```
awrichar, line 70: This comment confusingly doesn't match the variable names below. Might not be needed now that "deprecated" is in the var name?

internal/tokens/tifactory/factory.go
```diff
@@ -30,6 +30,16 @@ var pluginsByName = map[string]func() tokens.Plugin{
 	(*fftokens.FFTokens)(nil).Name(): func() tokens.Plugin { return &fftokens.FFTokens{} },
 }
 
+func InitPrefixWithSubPrefix(prefix config.PrefixArray) {
+	prefix.AddKnownKey(tokens.TokensConfigName)
+	prefix.AddKnownKey(tokens.TokensConfigType)
+	for name, plugin := range pluginsByName {
+		// Accept a superset of configs allowed by all plugins
```
awrichar, line 37: This comment is inaccurate now (we aren't squashing the config into a superset as before)

internal/tokens/tifactory/factory.go
```diff
@@ -30,6 +30,16 @@ var pluginsByName = map[string]func() tokens.Plugin{
 	(*fftokens.FFTokens)(nil).Name(): func() tokens.Plugin { return &fftokens.FFTokens{} },
 }
 
+func InitPrefixWithSubPrefix(prefix config.PrefixArray) {
```
awrichar, line 33: I think these names are going to change again when we bring in the latest firefly-common... but I think this one should probably land as "InitConfig" and the deprecated one below should be "InitConfigDeprecated", or something like that.

pkg/tokens/config.go
```diff
@@ -21,6 +21,8 @@ const (
 	TokensConfigName = "name"
 	// TokensConfigConnector is the connector plugin used for this token type (deprecated)
 	TokensConfigConnector = "connector" // TODO: remove
-	// TokensConfigPlugin is the connector plugin used for this token type
+	// TokensConfigPlugin is the connector plugin used for this token type (deprecated)
 	TokensConfigPlugin = "plugin"
+	// TokensConfigType is the name of the plugins used for this token type
```
awrichar, line 26: `plugins` -> `plugin`

internal/orchestrator/orchestrator.go
```diff
@@ -452,46 +456,88 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
...
+	}
+
+	if len(or.tokens) > 0 {
+		return nil
+	}
+
+	// If there still is no tokens config, check the deprecated structure for config
+	tokensConfigArraySize = deprecatedTokensConfig.ArraySize()
+	if tokensConfigArraySize > 0 {
+		log.L(ctx).Warnf("Your tokens config uses a deprecated configuration structure - the tokens configuration has been moved under the 'plugin' section")
```
awrichar, line 502: I think the section is called `plugins`

internal/orchestrator/orchestrator.go
```diff
@@ -452,46 +456,88 @@ func (or *orchestrator) initPlugins(ctx context.Context) (err error) {
...
+			return err
+		}
+		or.tokens[name] = plugin
+	}
+
+	if len(or.tokens) > 0 {
+		return nil
+	}
+
+	// If there still is no tokens config, check the deprecated structure for config
```
awrichar, line 499: So we only parse the new config _or_ the old config (we don't interpolate them). That seems reasonable, just need to make sure it's called out in the release notes.

internal/orchestrator/orchestrator.go
shorsher: Good catch, thanks!

internal/tokens/tifactory/factory.go
shorsher: Agreed, I'll include these in the PR for upgrading to latest firefly-common

awrichar (changes requested):
This looks good, but want to hold on merging until a few more pieces of functionality for FIR-12 are cued up.

shorsher:
closing, as this was included in https://github.com/hyperledger/firefly/pull/821/commits

PR #804, opened 2022-05-10
Move all identity APIs under /namespaces

awrichar:


pkg/core/identity.go
```diff
@@ -169,12 +171,25 @@ func (i *IdentityBase) Validate(ctx context.Context) (err error) {
...
+	if i.Type == IdentityTypeCustom {
+		options.extraAllowedChars = "/"
+	}
+	if err = ValidateFFNameFieldOptions(ctx, i.Name, "name", options); err != nil {
 		return err
 	}
+
+	var legacyDID string
+	if i.Type == IdentityTypeCustom {
+		if strings.HasPrefix(i.DID, FireFlyNodeDIDPrefix) || strings.HasPrefix(i.DID, FireFlyOrgDIDPrefix) {
```
awrichar, line 185: As noted on the FIR, this does mean we can never add additional "reserved" prefixes. I'm on the fence about whether that's a good move, so open to thoughts on whether we should have some other well-defined prefix for custom identities.

peterbroadhurst:
I have a migration question @awrichar. Take this example:

- Network has 10 members at v1.0.x
- Network has broadcast APIs, datatypes, and data from these members
- Network upgrades to v1.1.x
- New member joins the network

With the migration approach you described, of disabling `legacySystemIdentities` at some point, how will the new member receive/trust the data broadcast while the network was running at v1.0.x?

awrichar:
@peterbroadhurst I think the new member would have to join with `legacySystemIdentities` enabled, at least initially.

internal/apiserver/route_get_identity_by_did.go
```diff
@@ -25,22 +25,25 @@ import (
 	"github.com/hyperledger/firefly/pkg/core"
 )
 
-var postNewNamespace = &oapispec.Route{
-	Name:   "postNewNamespace",
-	Path:   "namespaces",
-	Method: http.MethodPost,
+var getIdentityByDID = &oapispec.Route{
+	Name:   "getIdentityByDID",
+	Path:   "identities/{did:did:.+}",
```
peterbroadhurst, line 30: Why is this `did:did:` @awrichar?

I'm assuming to query for `did:firefly:org/myorg` I'm going to query:
`/api/v1/identities/did:firefly:org/myorg` (_not_ `/api/v1/identities/did:did:firefly:org/myorg`)

peterbroadhurst: oh, got it - being dim. This is the Gorilla mux syntax confusing me - the first `did:` is saying "Treat this as a path param called `did`" 🤦

awrichar: Yea it reads a bit weird, but the regex matching `did:.+` will map into a gorilla path param named "did".

awrichar: Wonder if I could make that more obvious 😅

internal/coreconfig/namespace.go
```diff
@@ -0,0 +1,26 @@
...
+
+package coreconfig
+
+const (
+	// NamespaceName is a short name for a pre-defined namespace
+	NamespaceName = "name"
+	// NamespaceName is a long description for a pre-defined namespace
+	NamespaceDescription = "description"
+	// NamespaceOrgKey is a default signing key for blockchain transactions within a namespace (overrides top-level org.key)
+	NamespaceOrgKey = "org.key"
```
peterbroadhurst, line 25: Per comments on the FIR, I think this should be in a `multiparty.rootIdentity` (or similar) scoped sub-section of the config, in the new world.

awrichar: Yea, still an [open discussion] (https://github.com/hyperledger/firefly-fir/pull/12#discussion_r870661545) on exactly how to spell that. I think it does need to have "org" in the name though, since the root identity is ultimately always an org.

peterbroadhurst: I guess the question is whether we want to merge this and go through a period of flux in `main`, or hold it out until we close on the config design.

internal/namespace/manager.go
```diff
@@ -0,0 +1,143 @@
...
+			if err := nm.database.UpsertNamespace(ctx, newNS, true); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (nm *namespaceManager) GetConfigWithFallback(ns string, key config.RootKey) string {
+	if nsConfig, ok := nm.nsConfig[ns]; ok {
```
peterbroadhurst, line 136: Per other comments, I feel like this isn't top-level on the namespace, rather one level down - on the multi-party system identity specific section.

internal/orchestrator/orchestrator.go
```diff
@@ -393,7 +369,6 @@ func (or *orchestrator) initDataExchange(ctx context.Context) (err error) {
 	fb := database.IdentityQueryFactory.NewFilter(ctx)
 	nodes, _, err := or.database.GetIdentities(ctx, fb.And(
```
peterbroadhurst, line 370: I think I'm right in saying, this section will move to being one-per-namespace when @shorsher has completed the work to make DX plugins numerous.... so the namespace in the query query will come back at that point.

internal/orchestrator/orchestrator_test.go
```diff
@@ -83,6 +84,33 @@ type testOrchestrator struct {
 	msd *shareddownloadmocks.Manager
 	mae *admineventsmocks.Manager
 	mdh *definitionsmocks.DefinitionHandler
+	mns *namespacemocks.Manager
+}
+
+func (tor *testOrchestrator) cleanup(t *testing.T) {
+	tor.mdi.AssertExpectations(t)
```
peterbroadhurst, line 91: That's a lot of sub-components :)

... maybe the namespaces work (in a future PR) is the time to look at whether all of this attach at the same layer, or if some attach at the `orchestrator` layer, but most attach one level lower at a new namespace-specific orchestration layer tier.

pkg/core/identity.go
peterbroadhurst: What about _not_ allowing `/` in custom ones? Wouldn't that solve it.

- `did:firefly:bob/sally` - this is only acceptable if `bob` is an understood category
- `did:firefly:sally` - this is a custom one, because it does not allow `/`

internal/apiserver/route_get_identity_by_did.go
peterbroadhurst: Think it's good as it is.

internal/orchestrator/orchestrator_test.go
awrichar: I think it's already clear that the new `Namespace Manager` is going to shoulder a large part of orchestrator's responsibility by the time we're done. The final component tree is still foggy to me, but it has to end up a little better =)

internal/orchestrator/orchestrator.go
awrichar: Yes. For now data exchange is still global, so it needs to know about all the nodes across namespaces - but when the plugin moves to a namespace, it will only deal with the nodes in its own namespace.

awrichar:
This is now going to come in behind https://github.com/hyperledger/firefly/pull/820, as we look to be introducing a full "network rules migration" policy instead of the simple `legacySystemIdentities` config proposed here.

There's also at least one unrelated [naming question](https://github.com/hyperledger/firefly/pull/804#discussion_r872370750) unresolved on the comments above.

awrichar:
Updated on top of the latest, and replaced the `legacySystemIdentities` config option with a "network version" queried from the blockchain contract.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -1049,3 +1058,34 @@ func (e *Ethereum) getFFIType(solitidyType string) string {
...
 }
+
+func (e *Ethereum) getNetworkVersion(ctx context.Context, address string) (int, error) {
+	res, err := e.queryContractMethod(ctx, address, networkVersionMethodABI, []interface{}{})
+	if err != nil || !res.IsSuccess() {
+		// "Call failed" is interpreted as "method does not exist, default to version 1"
+		if strings.Contains(err.Error(), "FFEC100148") {
+			return 1, nil
+		}
+		return 0, ffresty.WrapRestErr(ctx, res, err, coremsgs.MsgEthconnectRESTErr)
```
awrichar, line 1069: Note to self - can just return `err` here, as it's already been wrapped

internal/blockchain/ethereum/ethereum.go
```diff
@@ -1049,3 +1058,34 @@ func (e *Ethereum) getFFIType(solitidyType string) string {
...
+		return 0, err
+	}
+	return strconv.Atoi(output.Output.(string))
+}
+
+func (e *Ethereum) NetworkVersion(ctx context.Context) (int, error) {
+	e.fireflyMux.Lock()
+	if e.networkVersion > 0 {
+		e.fireflyMux.Unlock()
+		return e.networkVersion, nil
```
awrichar, line 1082: Need to defer unlock until returning

internal/blockchain/ethereum/abi_definitions.go
```diff
@@ -96,3 +96,16 @@ var batchPinEventABI = ABIElementMarshaling{
 		},
 	},
 }
+
+var networkVersionMethodABI = ABIElementMarshaling{
```
peterbroadhurst, line 100: I note this might have a merge conflict with @nguyer who I believe is moving all of these to the [firefly-signer/pkg/api](https://pkg.go.dev/github.com/hyperledger/firefly-signer/pkg/abi) package.

... or if it doesn't then this is another place we should remove a duplicate interface definition.

internal/blockchain/ethereum/abi_definitions.go
```diff
@@ -96,3 +96,16 @@ var batchPinEventABI = ABIElementMarshaling{
...
+	Name:            "networkVersion",
+	Type:            "function",
+	StateMutability: "pure",
+	Inputs:          []ABIArgumentMarshaling{},
+	Outputs: []ABIArgumentMarshaling{
+		{
+			InternalType: "uint8",
```
peterbroadhurst, line 107: I wonder whether a `bytes32` would be the most efficient option here, and using a `semver` structure. `uint8` gets stored as a `unit256` anyway.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -49,21 +50,22 @@ const (
...
+	callbacks       blockchain.Callbacks
+	client          *resty.Client
+	fftmClient      *resty.Client
+	streams         *streamManager
+	streamID        string
+	fireflyContract struct {
+		mux            sync.Mutex
+		address        string
+		fromBlock      string
+		networkVersion int
```
peterbroadhurst, line 67: Note comment above over `int` vs. semver. Probably an open discussion, rather than a proposal.

peterbroadhurst: As I ramble with half-formed thoughts here... I wonder whether it's actually a network-options block, where we've reserved a byte for a version integer, but other contract implementations could express additional details in that.

internal/blockchain/ethereum/abi_definitions.go
peterbroadhurst: (added more discussion on this in a separate comment)

internal/blockchain/ethereum/ethereum.go
peterbroadhurst: I guess I may be completely overthinking this, as fundamentally this is just a constructor argument with a getter on the contract. So any extensibility to other options might be contract specific.

I'm just feeling an extensibility worry, and wondering if there's some way to do something now that would give future-us some more flexibility for some (currently unknown) migration challenge ahead, that could avoid the need for another contract migration by allowing more options to be encoded in the existing contract without change.

internal/coremsgs/en_error_messages.go
```diff
@@ -244,4 +244,5 @@ var (
 	MsgDuplicatePluginName                = ffe("FF10395", "Invalid plugin configuration - plugin with name %s already exists")
 	MsgInvalidFireFlyContractIndex        = ffe("FF10396", "No configuration found for FireFly contract at %s")
 	MsgUnrecognizedNetworkAction          = ffe("FF10397", "Unrecognized network action: %s", 400)
+	MsgTerminateNotSupported              = ffe("FF10398", "Termination is not supported on namespace %s", 400)
```
peterbroadhurst, line 247:
```suggestion
	MsgTerminateNotSupported              = ffe("FF10398", "The terminate operation to mark the switch of smart contracts is not supported on namespace %s", 400)
```

internal/namespace/manager.go
peterbroadhurst: @awrichar - did you have any thoughts on this spelling point?

internal/blockchain/ethereum/ethereum.go
awrichar: I think these are all good points, and probably worthy of a lengthy discussion in their own right. Perhaps if an `int` is "good enough" for this PR, that discussion can be deferred to a follow-up issue or FIR? We can make a separate call on whether it's all part of 1.1 (and thus can break/replace all the logic here).

internal/coreconfig/namespace.go
awrichar: Seems like the discussion on the FIR has resolved - will incorporate those changes back to here.

internal/blockchain/ethereum/ethereum.go
awrichar: I've opened #846 as a placeholder to continue this discussion.

pkg/core/identity.go
```diff
@@ -183,25 +188,19 @@ func (i *IdentityBase) Validate(ctx context.Context) (err error) {
...
-		if i.Namespace == SystemNamespace {
-			return "", i18n.NewError(ctx, i18n.MsgCustomIdentitySystemNS, SystemNamespace)
+		if i.Namespace == LegacySystemNamespace {
+			return "", i18n.NewError(ctx, i18n.MsgCustomIdentitySystemNS, LegacySystemNamespace)
 		}
 		if i.Parent == nil {
 			return "", i18n.NewError(ctx, i18n.MsgNilParentIdentity, i.Type)
 		}
-		return fmt.Sprintf("%s%s/%s", FireFlyCustomDIDPrefix, i.Namespace, i.Name), nil
+		return fmt.Sprintf("%s%s", FireFlyCustomDIDPrefix, i.Name), nil
```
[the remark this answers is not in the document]
awrichar: Whenever we call `Validate` on an identity, we verify that `i.Name` is a valid FireFly name, which means it cannot contain `/`. I didn't think we need to do any additional validation here when generating the DID.

internal/coreconfig/namespace.go
awrichar: Seems like the discussion on the FIR has converged for now, so I've reworked the config here to match. Changes in the new commit:

- Replace "mode: multiparty" with "multiparty.enabled: true"
- Default to multiparty mode if org name or key is set
- Remove previously-deprecated "org.identity" root key
- Add new "defaultKey" config to namespaces (valid for both multiparty and gateway namespaces, overrides "multiparty.org.key" for signing)

internal/blockchain/ethereum/abi_definitions.go
peterbroadhurst: looks like this one is ready first - sorry @nguyer

internal/identity/identitymanager.go
```diff
@@ -206,44 +216,41 @@ func (im *identityManager) firstVerifierForIdentity(ctx context.Context, vType c
...
+	if key, ok := im.nodeOwnerBlockchainKey[namespace]; ok {
+		return key, nil
 	}
 
-	orgKey := config.GetString(coreconfig.OrgKey)
-	if orgKey == "" {
-		orgKey = config.GetString(coreconfig.OrgIdentityDeprecated)
-		if orgKey != "" {
-			log.L(ctx).Warnf("The %s config key has been deprecated. Please use %s instead", coreconfig.OrgIdentityDeprecated, coreconfig.OrgKey)
+	defaultKey := im.namespace.GetDefaultKey(namespace)
```
peterbroadhurst, line 240: I think it's important that we do *not* use the `defaultKey` in the case of a root org registration.

e.g. when I have an NS config like this:

```yaml
  defaultKey: 0xabcd
  multiparty:
    enabled: true
    org:
      name: bob
      key: 0xbbb
```

Then my organization must be registered as `0xbbb` *not* `0xaaa`

awrichar: Alright, pushed another commit to differentiate and ensure that the "register org" path only looks at the `multiparty.org.key`, but all other paths should let `defaultKey` take precedence.

peterbroadhurst (approved):
👍 - sorry this one turned got so intertwined with other things. Looks great

PR #811, opened 2022-05-11
Add reference docs generation for types

nguyer:
This is a starting point for auto-generation of reference docs for types in FireFly. There is a lot more work to do to fill out the detailed descriptions of the types currently listed, and we need pages for more types. But this gives us the framework we need to create this pages, and a couple of lightweight examples to start with.

https://github.com/hyperledger/firefly/issues/569


peterbroadhurst (approved):
Thanks @nguyer - this infrastructure is great. I will make helping fill in the descriptions a priority 👍

PR #813, opened 2022-05-12
Return definition rejection reasons as errors

awrichar:
There is always a reason, so return it and let the caller decide how to log it.

I chose to return it in `err` - could also choose to add a new field to `HandlerResult` instead, but this felt logical as it does represent an exceptional/error state.


internal/definitions/definition_handler_identity_claim.go
```diff
@@ -30,18 +30,18 @@ func (dh *definitionHandlers) handleIdentityClaimBroadcast(ctx context.Context,
 	var claim core.IdentityClaim
 	valid := dh.getSystemBroadcastPayload(ctx, msg, data, &claim)
 	if !valid {
-		return HandlerResult{Action: ActionReject}, nil
+		return HandlerResult{Action: ActionReject}, fmt.Errorf("unable to process identity claim %s - invalid payload", msg.Header.ID)
```
peterbroadhurst, line 33: I feel like leaking in `fmt.Errorf` (rather than `i18n.NewError`) calls into the code is a bit of a slippery slope.
I understand we're making an assumption the caller won't return it anywhere... but the goal of the PR seemed to be that the raiser of an error wouldn't make assumptions about the caller.
Probably worth just popping these into `coremsg` - what do you think?

peterbroadhurst (changes requested):
Marking this as request changes - while we resolve the discussion on the strings vs. translated errors (not necessarily saying it has to result in a change, but do think we want an explicit decision).

internal/definitions/definition_handler_identity_claim.go
nguyer: I agree with this. Quite a few places in the code will simply continue to pass `err` back up until it is returned in an API response. I don't see any harm in returning a proper error code and translation even from lower level functions like this. If they get returned in an API response that's great. If they just get logged it's also great, and potentially in the operators native language.

internal/definitions/definition_handler.go
```diff
@@ -143,20 +144,19 @@ func (dh *definitionHandlers) HandleDefinitionBroadcast(ctx context.Context, sta
 	case core.SystemTagDefineContractAPI:
 		return dh.handleContractAPIBroadcast(ctx, state, msg, data, tx)
 	default:
-		l.Warnf("Unknown SystemTag '%s' for definition ID '%s'", msg.Header.Tag, msg.Header.ID)
-		return HandlerResult{Action: ActionReject}, nil
+		return HandlerResult{Action: ActionReject}, fmt.Errorf("unknown system tag '%s' for definition ID '%s'", msg.Header.Tag, msg.Header.ID)
```
nguyer, line 147: Does the fact that this now returns an error affect logic in the caller of this function?

internal/definitions/definition_handler_datatype.go
```diff
@@ -18,38 +18,33 @@ package definitions
...
 	if err := dt.Validate(ctx, true); err != nil {
-		l.Warnf("Unable to process datatype broadcast %s - validate failed: %s", msg.Header.ID, err)
-		return HandlerResult{Action: ActionReject}, nil
+		return HandlerResult{Action: ActionReject}, fmt.Errorf("unable to process datatype definition %s - validate failed: %s", msg.Header.ID, err)
 	}
 
 	if err := dh.data.CheckDatatype(ctx, dt.Namespace, &dt); err != nil {
-		l.Warnf("Unable to process datatype broadcast %s - schema check: %s", msg.Header.ID, err)
-		return HandlerResult{Action: ActionReject}, nil
+		return HandlerResult{Action: ActionReject}, fmt.Errorf("unable to process datatype definition %s - schema check: %s", msg.Header.ID, err)
```
nguyer, line 39: Same question on these since we're now returning errors where we didn't used to

internal/definitions/definition_handler.go
awrichar: No. The only caller is in `aggregator.go`, and in the case of `ActionReject` it simply logs the error.

internal/definitions/definition_handler_identity_claim.go
awrichar: Yea, in the current world they're only logged, just because this happens on a background event loop that has no connection to a user action.

But for the forthcoming "gateway namespaces", much of this logic will be reused in a synchronous way, directly in response to a user action (such as when creating local identities and token pools). So in that case, we could bubble them up (instead of a more generic "definition rejected" error), and localizing them is probably the right call. It means ~30 new error strings.

peterbroadhurst (approved):
👍

PR #818, opened 2022-05-16
Add language tag to FFM, FFE, FFC, etc

nguyer:
This PR depends on https://github.com/hyperledger/firefly-common/pull/7 and a new release of `firefly-common`.

This adds the AmericanEnglish language tag to each of our existing message keys. These will be used as fallbacks if a specific key is missing in another language.

For now, this does not _actually_ create multi-lingual documentation, but it lays the groundwork for it. I need to do some separate work in the docs generation to separate out pages for each language. That will be in a separate PR.


internal/coremsgs/en_api_translations.go
```diff
@@ -16,190 +16,193 @@
...
+	"github.com/hyperledger/firefly-common/pkg/i18n"
+	"golang.org/x/text/language"
+)
 
 var ffm = i18n.FFM
 
 //revive:disable
 var (
-	CoreSystemNSDescription = ffm("core.systemNSDescription", "FireFly system namespace")
+	CoreSystemNSDescription = ffm(language.AmericanEnglish, "core.systemNSDescription", "FireFly system namespace")
```
awrichar: Can we follow the same pattern from `firefly-common` to make the local `ffm` helper just add the language tag instead of including it on every line?

internal/coremsgs/es/en_struct_descriptions.go
```diff
@@ -0,0 +1,52 @@
+// Copyright © 2022 Kaleido, Inc.
```
awrichar, line 1: Should this file be name `es_struct_descriptions`?

internal/coremsgs/es/en_struct_descriptions.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/hyperledger/firefly-common/pkg/i18n"
+	"golang.org/x/text/language"
+)
+
+//revive:disable
+
+/*
+This file contains the English field level descriptions that are used in
```
awrichar, line 27: Not English

awrichar (approved):

PR #820, opened 2022-05-17
Support migration of the FireFly contract from one location to another

awrichar:
~~In a chain with #821.~~
~~Depends on https://github.com/hyperledger/firefly-common/pull/8 being merged and integrated here.~~

Summary of changes:

1. Support a list of contract locations in the blockchain config under a new `fireflyContract` key.
New config example (for Ethereum):
```
blockchain:
- name: blockchain0
  type: ethereum
  ethereum:
    ethconnect:
      url: http://ethconnect_0:8080
      topic: "0"
    fireflyContract:
    - address: 0x057b6ae997a77726a2cfed8cbef7fd94ce42b15a
      fromBlock: 0
    - address: 0xb347e9de7d9f312c24c375fff7632e189bf1dd43
```

This means deprecation of the previous keys `ethconnect.instance` and `fabconnect.chaincode`.

3. New `fireflyContract` field added to namespaces.
Records the config index and address of the active contract, as well as the final event for all previously-terminated contracts. New namespace example:
```
{
  "id": "3ee769e3-92c5-4695-acd6-75c6e362082c",
  "name": "ff_system",
  "description": "FireFly system namespace",
  "type": "system",
  "created": "2022-05-24T19:18:40.07772Z",
  "fireflyContract": {
    "active": {
      "index": 1,
      "info": {
        "address": "0xb347e9de7d9f312c24c375fff7632e189bf1dd43",
        "fromBlock": "oldest",
        "subscription": "sb-2f4dc8bb-b577-4b9b-582b-9b5bbd347706"
      }
    },
    "terminated": [
      {
        "index": 0,
        "finalEvent": "000000000007/000000/000000",
        "info": {
          "address": "0x057b6ae997a77726a2cfed8cbef7fd94ce42b15a",
          "fromBlock": "oldest",
          "subscription": "sb-b43ec7ed-b158-48dc-473a-05bda8e0fc28"
        }
      }
    ]
  }
}
```
Currently there is only one globally active FireFly contract, and its info is recorded against the "ff_system" namespace. In a future PR as part of https://github.com/hyperledger/firefly-fir/pull/12, each namespace will have its own blockchain plugin and therefore its own contract.

4. New `/network/operatoraction` API, with a single supported action type of `terminate`.
Sends a specially formatted "BatchPin" transaction to signal that all members should terminate listening to the current FireFly contract and migrate to the next one configured.

**Note that this is all separate from migrating the "network rules" of a namespace (which is also in scope for FIR-12). This first step just allows migrating from one contract to another. In a future PR we can introduce the capability for a new contract to expose a "network version", so that a contract migration may also cause a rules migration (such as changing the signature of BatchPin or the rules for resolving identities).**


awrichar:
I think this will now need to slot in after #826

awrichar:
Now rebased on mainline after the merge of #826

peterbroadhurst:
> each namespace will have its own blockchain plugin and therefore its own contract index.

I think I'm right interpreting this as:

> Each unique `remote` namespace will have it's own blockchain plugin. Noting that there might be multiple local namespaces on the node that have separate root identities registered in that business network, and they can choose to share a single blockchain plugin, or each have a separate blockchain plugin.

peterbroadhurst:
k - feel like we need to net that out in FIR-12. Some of it might be terminology, but if you really have to define an individual plugin for each namespace and those configurations can't be re-used... it feels odd that you don't just define them directly in the namespace.

internal/apiserver/route_post_network_migrate.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+
+	"github.com/hyperledger/firefly/internal/coremsgs"
+	"github.com/hyperledger/firefly/internal/oapispec"
+	"github.com/hyperledger/firefly/pkg/core"
+)
+
+var postNetworkMigrate = &oapispec.Route{
+	Name:            "postNetworkMigrate",
+	Path:            "network/migrate",
```
peterbroadhurst, line 29: I quite liked the "Operator Action" general terminology from the API you described, where this is just one special form of a command and we aren't tying ourselves to this being the only type of command that ever exists.

Illustrating the point, with some starter for 10 spelling suggestions:
What do you think about `network/governance/action`, `network/govaction` or `network/operatoraction`.
Then a `type` under the payload of `migration_v_1_0`.

internal/database/sqlcommon/namespace_sql.go
```diff
@@ -90,6 +91,7 @@ func (s *SQLCommon) UpsertNamespace(ctx context.Context, namespace *core.Namespa
 				Set("name", namespace.Name).
 				Set("description", namespace.Description).
 				Set("created", namespace.Created).
+				Set("contract_index", namespace.ContractIndex).
```
peterbroadhurst, line 94: ... sure I'll work this out as I go through, but reading this I assume this means "the last contract_index I detected an event from"

internal/apiserver/route_post_network_migrate.go
peterbroadhurst: ... or I guess is the action being performed here not actually a migration, rather just an `end`/`terminate`/`stop`/`swap`/`jump` type action.

The overall migration is a journey:
1. Deploy a new contract
2. All members add that contract to their config, agreeing who's going to say "go"
4. One member authorized to do so, does this action to signify we're ready to move

internal/blockchain/ethereum/ethereum.go
```diff
@@ -172,67 +174,89 @@ func (e *Ethereum) Init(ctx context.Context, config config.Section, callbacks bl
...
+func (e *Ethereum) initStreams(contractIndex int) (err error) {
+
+	ctx := e.ctx
+	e.streams = &streamManager{client: e.client}
+	if e.contractConf.ArraySize() > 0 || contractIndex > 0 {
+		// New config (array of objects under "fireflyContract")
+		if contractIndex >= e.contractConf.ArraySize() {
+			return i18n.NewError(ctx, coremsgs.MsgInvalidFireFlyContractIndex, fmt.Sprintf("blockchain.ethereum.fireflyContract[%d]", contractIndex))
+		}
+		e.contractAddress = e.contractConf.ArrayEntry(contractIndex).GetString(FireFlyContractAddress)
```
peterbroadhurst, line 205: If I read it right, this code doesn't do anything to clean up the old subscriptions that are defunct. Not objecting to that, but making a note as I think it's significant (assuming I don't find that elsewhere).

peterbroadhurst: I might have missed it, but I'm looking out for the code that records the conclusive `endBlock` for the contract index `0`. I see us calling `bi.Start()` which calls into this code, but I don't see anything that prevents further events being consumed.

Otherwise, in the middle of the processing of the `OperatorActionMigrate` message, we have delete the old stream in a way that prevents any more events possibly being processed. I don't know that we have that API contract with blockchain event listeners, that they have to be able to process a REST API to delete a stream in the middle of an ack.

awrichar:
> if you really have to define an individual plugin for each namespace

@peterbroadhurst yea I think it's mostly terminology here. From a configuration perspective, you can reuse a plugin configuration in many namespaces. From an internal instantiation perspective, there will be a separate plugin instance+event loop created for every one of those namespaces.

internal/blockchain/ethereum/ethereum.go
awrichar: Good point. I was actually planning to delete and recreate the stream and then forgot that piece. But your suggestion of an end block is probably better, if I can figure out where to record it...

internal/apiserver/route_post_network_migrate.go
awrichar: Still ruminating on the spelling for this API... but I did change the internal terminology to `terminate` over `migrate`, so that will likely bubble up here in some form.

internal/blockchain/ethereum/ethereum.go
awrichar: OK, made a significant additional commit to handle these two changes:
1. The "final event" of every previous FireFly contract is stored forever, and events from that contract past that point are ignored
2. When migrating to a new contract, the new subscription is simply added to the existing stream (no attempt to clean up old ones and no need to restart the websocket)

internal/apiserver/route_post_network_migrate.go
awrichar: For this next pass, I changed the API to `/network/operatoraction` with a `type` of `terminate`.

Still open if sliding in "governance" makes more sense. I'm very slightly concerned about "operator" vs the already-used term "operation" (though not terribly so).

internal/blockchain/ethereum/ethereum.go
```diff
@@ -173,79 +178,131 @@ func (e *Ethereum) Init(ctx context.Context, config config.Section, callbacks bl
...
+func (e *Ethereum) ConfigureContract(ctx context.Context, contracts *core.FireFlyContracts) (err error) {
+
+	log.L(ctx).Infof("Resolving FireFly contract at index %d", contracts.Active.Index)
+	address, fromBlock, err := e.resolveFireFlyContract(ctx, contracts.Active.Index)
+	if err != nil {
+		return err
+	}
+
+	e.fireflyContract = address
+	e.fireflyFromBlock = fromBlock
```
awrichar, line 277: Note that there's no validation of the current config values against the `contracts` info that was passed in from the database - the config always overwrites what is there (and ultimately gets stored back in the database by the caller). I toyed briefly with trying to flag errors for "invalid changes" in your local config, but that got sticky.

awrichar: Also now that I look at this again, I may need to add a mutex since `e.fireflyContract` is used in a few methods (across different Goroutines).

internal/blockchain/ethereum/ethereum.go
```diff
@@ -173,79 +178,131 @@ func (e *Ethereum) Init(ctx context.Context, config config.Section, callbacks bl
...
+	e.initInfo.sub, err = e.streams.ensureFireFlySubscription(ctx, e.fireflyContract, e.fireflyFromBlock, e.initInfo.stream.ID, batchPinEventABI)
+	if err == nil {
+		contracts.Active.Info = fftypes.JSONObject{
+			"address":      address,
+			"fromBlock":    fromBlock,
+			"subscription": e.initInfo.sub.ID,
+		}
```
awrichar, line 284: Note that there's no validation of the current config values against the contracts info that was passed in from the database - the config always overwrites what is there (and ultimately gets stored back in the database by the caller). I toyed briefly with trying to flag errors for "invalid changes" in your local config, but that got sticky.

internal/apiserver/route_post_network_migrate.go
awrichar: Or what about the simpler names `/network/action` and `SubmitNetworkAction`?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -173,79 +180,137 @@ func (e *Ethereum) Init(ctx context.Context, config config.Section, callbacks bl
...
+func (e *Ethereum) Start() (err error) {
 	return e.wsconn.Connect()
 }
 
+func (e *Ethereum) resolveFireFlyContract(ctx context.Context, contractIndex int) (address, fromBlock string, err error) {
+
+	if e.contractConfSize > 0 || contractIndex > 0 {
+		// New config (array of objects under "fireflyContract")
+		if contractIndex >= e.contractConfSize {
+			return "", "", i18n.NewError(ctx, coremsgs.MsgInvalidFireFlyContractIndex, fmt.Sprintf("blockchain.ethereum.fireflyContract[%d]", contractIndex))
```
awrichar, line 230: Another thing to note - when terminating a contract, this call is wrapped in a "retry" loop from event manager. Therefore it will constantly log failure if there is no "next" contract configured, until you fix the config and restart.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -599,7 +665,22 @@ func (e *Ethereum) SubmitBatchPin(ctx context.Context, operationID *fftypes.UUID
 		batch.BatchPayloadRef,
 		ethHashes,
 	}
-	return e.invokeContractMethod(ctx, e.instancePath, signingKey, batchPinMethodABI, operationID.String(), input)
+	e.fireflyMux.Lock()
```
peterbroadhurst, line 668: I think you could drop the scope of this lock to be only:

```go
e.fireflyMux.Lock()
instancePath := e.instancePath
e.fireflyMux.Unlock()
```

Rather than holding it for the full invocation path. I understand that means the contract could change while the submission is in-flight. But I believe there is a much longer time window that exists for that anyway - the time (on all the nodes) to detect the blockchain event.

internal/blockchain/fabric/fabric.go
```diff
@@ -170,61 +176,113 @@ func (f *Fabric) Init(ctx context.Context, config config.Section, callbacks bloc
...
+		}
+	}
+	return err
+}
+
+func (f *Fabric) TerminateContract(ctx context.Context, contracts *core.FireFlyContracts, termination *blockchain.Event) (err error) {
+
+	chaincode := termination.Info.GetString("chaincodeId")
+	f.fireflyMux.Lock()
+	if chaincode != f.fireflyChaincode {
```
peterbroadhurst, line 271: Just an observation about us having added quite a bit of extra code duplication of logic here. At some point it feels like we need a base plugin/framework for the common code between `ethereum` and `fabric`. However, I have no objection to this going in as is - as I'm expecting you were thinking about this duplication as you went through, and how interleaved the logic is.

internal/blockchain/fabric/fabric.go
```diff
@@ -585,9 +640,24 @@ func (f *Fabric) SubmitBatchPin(ctx context.Context, operationID *fftypes.UUID,
...
+	pinInput := map[string]interface{}{
+		"namespace":  "firefly:" + action,
+		"uuids":      hexFormatB32(nil),
+		"batchHash":  hexFormatB32(nil),
+		"payloadRef": "",
+		"contexts":   []string{},
+	}
 	input, _ := jsonEncodeInput(pinInput)
-	return f.invokeContractMethod(ctx, f.defaultChannel, f.chaincode, batchPinMethodName, signingKey, operationID.String(), batchPinPrefixItems, input)
+	f.fireflyMux.Lock()
```
peterbroadhurst, line 658: Same comment as for Ethereum on the scope of this lock

internal/blockchain/fabric/fabric.go
```diff
@@ -170,61 +176,113 @@ func (f *Fabric) Init(ctx context.Context, config config.Section, callbacks bloc
...
 	return nil
 }
 
-func (f *Fabric) Start() error {
+func (f *Fabric) resolveFireFlyContract(ctx context.Context, contractIndex int) (chaincode, fromBlock string, err error) {
+
+	if f.contractConfSize > 0 || contractIndex > 0 {
+		// New config (array of objects under "fireflyContract")
+		if contractIndex >= f.contractConfSize {
+			return "", "", i18n.NewError(ctx, coremsgs.MsgInvalidFireFlyContractIndex, fmt.Sprintf("blockchain.fabric.fireflyContract[%d]", contractIndex))
```
peterbroadhurst, line 221: Might be worth talking to @jimthematrix about the Fabric side of this, and whether in practice we use this feature you've added for jumping chaincode names, vs. just upgrading the Fabric chaincode in-place. Not a blocker to this merging, but something to think about in the next step when developing the v.next contract code, and testing this journey end to end with a Fabric environment.

internal/events/operator_action.go
```diff
@@ -0,0 +1,59 @@
...
+		if err := bi.TerminateContract(ctx, &ns.Contracts, event); err != nil {
+			return err
+		}
+		return em.database.UpsertNamespace(ctx, ns, true)
+	})
+}
+
+func (em *eventManager) BlockchainOperatorAction(bi blockchain.Plugin, action string, event *blockchain.Event, signingKey *core.VerifierRef) error {
+	return em.retry.Do(em.ctx, "handle operator action", func(attempt int) (retry bool, err error) {
+		// TODO: verify signing identity
```
peterbroadhurst, line 42: Feels like this one should probably close out before this goes in, assuming it's pretty straightforward to pop code in similar to `checkOnchainConsistency`. I believe any resolved root Organization identity should be able to do this today, but future PRs might enhance that (with a gatekeeper / network permissioning model).

db/migrations/postgres/000090_add_namespace_contractindex.up.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE namespaces ADD COLUMN firefly_contracts TEXT;
```
peterbroadhurst, line 2: Can we rename the filenames for the new richer JSON object we're storing `contractindex`=>`fireflycontracts`

internal/apiserver/route_post_network_migrate.go
peterbroadhurst: I like these simpler names 👍 

... notice we still have `operatoraction` in the code - worth changing before we merge?

internal/blockchain/fabric/fabric.go
awrichar: Yea, my gut said it wouldn't improve overall readability to factor out utilities, because so many little details are just slightly different ("address" vs "chaincodeId", for instance). It's probably an exercise worth doing at some point just to see what's possible, but I've punted (again) for now.

internal/blockchain/ethereum/ethereum.go
awrichar: I also want to note that with this model we never delete the old subscriptions (just ignore any events they produce). That leaves a bit of unnecessary overhead on ethconnect, but it was just difficult enough that I didn't want to mess with it at this time (as noted above, deleting a subscription while in the process of processing/acking an event _from_ that subscription seems like it could cause problems).

peterbroadhurst (approved):
👍  - really pleased this all worked through, and we have an upgrade path in place to adopt the new per-namespace contracts.

PR #822, opened 2022-05-18
Docs internationalization

nguyer:
This PR adds the foundations for internationalization of the docs site. It adds a new Jekyll plugin which handles translating string templates into the appropriate language. It also has a graceful fallback by rendering English pages / sections where a native local translation does not exist for the currently selected language.

Here are the major things in this PR:

- `make docs` command that will build the docs site and check for broken links
- `jekyll-multiple-languages-plugin` added for internationalization
- Introduction of a new `i18n_page` Jekyll layout - use this if you want your page to be translatable
- Theme templates are updated to translate page titles or fallback in several key places
  - The left nav bar
  - Page breadcrumbs
  - Tables of contents
- Several Spanish page titles and the content for a single page were added for testing and as an example of how to add additional translations. **Disclaimer:** These were generated with Google Translate. They are probably not good translations.
- A translation picker dropdown has been added to the top right of the page
- We are using a new GitHub action that builds the docs site and pushes static files to the `gh-pages` branch. Going forward, we will serve the docs out of this branch. This will also allow us to version the docs in the very near future.
- There are quite a few new static files added here because we no longer use a remote theme for Jekyll. The theme has been embedded to allow for complete customization, and integration with the internationalization plugin. This also allows for a couple of new visual improvements:
  - Blockquotes now stand out a lot more
  - Official FireFly accent colors are used throughout the docs


docs/_i18n/en.yml
```diff
@@ -0,0 +1,27 @@
+langs:
```
peterbroadhurst, line 1: Is this a generated directory? Or do we have to keep this updated with translations?

docs/_sass/color_schemes/dark.scss
```diff
@@ -0,0 +1,17 @@
+$body-background-color: $grey-dk-300;
```
peterbroadhurst, line 1: 🕶️  coming soon ™️

docs/architecture/index.md
```diff
@@ -1,6 +1,6 @@
 ---
-layout: default
-title: Architecture
+layout: i18n_page
+title: pages.architecture
```
peterbroadhurst, line 3: Related to earlier questions on whether we need to ensure updates to the `en.yml` need to be checked in... for generated pages do we need to generate the english translations for the pages too?

peterbroadhurst:
Could we remove the image assets that we're not using @nguyer ? - would have commented in-line, but Github wouldn't let me add comments to binary files.

nguyer:
> Could we remove the image assets that we're not using @nguyer ? - would have commented in-line, but Github wouldn't let me add comments to binary files.

Yeah that sounds good. I noticed that after I opened this PR. I'll try to scrub some stuff we're not using.

docs/_i18n/en.yml
nguyer: This is not autogenerated. If we add a new page **and** we want it to be "translatable", the page should use the `i18n_page` layout and we need to add the appropriate title and / or content to this directory. If we make it an `i18n_page` and forget to add stuff to the `en.yml` or `en` directory the build will fail, so that part is enforced.

docs/architecture/index.md
peterbroadhurst: Thanks for clarifying on chat, that this phase will add the infrastructure to allow contributions, but the switch to having it on across the board only really makes sense once the first set of contributions is in place. So we will defer that work for now.

peterbroadhurst (approved):

PR #823, opened 2022-05-19
Add additional test suite for running E2E tests without tokenTestSuite against specific Firefly namespace

calbritt:
Can call test suite directly to run without token test suite.

### DIFF
--- a/test/e2e/ethereum_namespace_no_tokens_test.go
+++ b/test/e2e/ethereum_namespace_no_tokens_test.go
@@ -1,0 +1,28 @@
+// Copyright © 2022 Kaleido, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


awrichar (changes requested):
I'm slightly wary of having a suite that reads a config file to decide which tests it will or will not run. Feels like it's better to explicitly say which tests need to run, or you'd have to carefully inspect the output to ensure a test wasn't unintentionally skipped.

Can you give some background on the motivation for this?

calbritt:
> I'm slightly wary of having a suite that reads a config file to decide which tests it will or will not run. Feels like it's better to explicitly say which tests need to run, or you'd have to carefully inspect the output to ensure a test wasn't unintentionally skipped.
> 
> Can you give some background on the motivation for this?

For firefly stacks that don't have a token connector running we want to be able to still complete a successful E2E run without the tokens test failing.

awrichar:
OK - would it make sense to have a different top-level suite for that?

calbritt:
I will change existing '*E2ESuite* files to use `Runner` as new suffix to indicate these are running a compilation of test suites. As part of that change I will also add something like a `NoTokenConnectorRunner` file to contain all Ethereum E2E tests not including TokensTestSuite instead of skipping the tests.

PR #824, opened 2022-05-19
Include component name when initialization fails

awrichar:
I needed this briefly for debugging; might as well keep it around.


peterbroadhurst (approved):
📛

PR #830, opened 2022-05-23
Add version support to docs site

nguyer:
This PR adds significant new functionality to the docs GitHub action to manage versions of the docs site. This will create a directory tree that ends up looking like this:

```
+ / (latest release)
|-- v1.0.0
|-- v1.0.1
|-- ...
|-- head (current commit in main)
```

It will also build a JSON file that contains the full list of all these versions to build the dropdown menu that shows up in the docs rendered pages (with the addition of a `latest` link that links to `/`).

These changes allow us to always keep the docs up to date with the code in the same commit, but users navigating to the docs site will see the latest release by default. They can also use the dropdown menu to browse older release as well.

> **NOTE** There will be a one time manual action necessary to generate docs for older versions. I can take care of this after this PR is approved and merged.


.github/workflows/docs.yml
```diff
@@ -26,13 +36,77 @@ jobs:
...
+        run: echo "DOCS_VERSION=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Set version in _config.yml
+        run: |
+          echo "version: ${{ env.DOCS_VERSION }}" >> ./docs/_config.yml
+
+      # Only need to build the full list of versions every time there is a new release
+      - name: Update version.json list
+        if: github.event_name == 'release'
+        run: ls ./gh-pages | grep -E '^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$' | sed 's/^/"/;s/$/"/' | tr '\n' ',' | sed 's/.$//' | sed 's/^/[/;s/$/]/' | jq '. += ["${{ env.DOCS_VERSION }}"] | sort_by(.) | reverse | . += ["latest", "head"]' > ./docs/assets/js/versions.json
```
peterbroadhurst, line 56: I've fully validated every character of this regex 😇

peterbroadhurst: ... I wonder if it's worth a comment describing what the regex is doing, and how it was derived

.github/workflows/docs.yml
```diff
@@ -26,13 +36,77 @@ jobs:
...
+      ## FINAL STEPS TO COMPLETE IF THIS WAS A NEW RELEASE ##
+
+      - name: Copy new version into version directory
+        if: github.event_name == 'release'
+        working-directory: docs
+        run: mv _site ${{ env.DOCS_VERSION }}
+
+      - name: Copy existing versions into working directory
+        if: github.event_name == 'release'
+        run: ls ./gh-pages | grep -E '^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$' | xargs -I '{}' mv './gh-pages/{}' ./docs
```
peterbroadhurst, line 94: I see this is the same regex - does github pages have a variable construct we can use to have this in the code once?

peterbroadhurst (approved):
Approved - as this looks sound, and I could follow the logic.
One item to review with regards to future maintainability.

.github/workflows/docs.yml
nguyer: Sure. It actually comes from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string I'll add a comment in the action explaining that.

.github/workflows/docs.yml
nguyer: Yeah I can assign it to a variable. I'll make that change.

.github/workflows/docs.yml
peterbroadhurst: Thanks for adding the comment/variable

PR #833, opened 2022-05-24
Namespace config validation

shorsher:
Adding additional configuration validation to the namespace manager, as specified in [FIR-12](https://github.com/hyperledger/firefly-fir/pull/12)

* `name` must be unique on this node
 * for historical reasons, "ff_system" is a reserved string and cannot be used as a `name` or `remoteName`
 * a `database` plugin is required for every namespace
 * if `mode: multiparty` is specified, plugins _must_ include one each of `blockchain`,
  `dataexchange`, and `sharedstorage`
 * if `mode: gateway` is speicified, plugins _must not_ include `dataexchange` or `sharedstorage`
 * at most one of each type of plugin is allowed per namespace, except for tokens (which
  may have many per namespace)

depends on changes in https://github.com/hyperledger/firefly-common/pull/14


internal/namespace/config.go
```diff
@@ -34,8 +34,12 @@ var (
 func InitConfig(withDefaults bool) {
 	namespacePredefined.AddKnownKey(coreconfig.NamespaceName)
 	namespacePredefined.AddKnownKey(coreconfig.NamespaceDescription)
+	namespacePredefined.AddKnownKey(coreconfig.NamespaceRemoteName)
+	namespacePredefined.AddKnownKey(coreconfig.NamespacePlugins)
+	// do we need to set the default mode?
```
awrichar, line 39: Yes, there should be an `AddKnownKey` here for `NamespaceMode`, with a default of "multiparty"

internal/namespace/config.go
```diff
@@ -34,8 +34,12 @@ var (
...
 	if withDefaults {
 		namespaceConfig.AddKnownKey(NamespacePredefined+".0."+coreconfig.NamespaceName, "default")
 		namespaceConfig.AddKnownKey(NamespacePredefined+".0."+coreconfig.NamespaceDescription, "Default predefined namespace")
+		namespaceConfig.AddKnownKey(NamespacePredefined+".0."+coreconfig.NamespaceMode, "multiparty")
```
awrichar, line 43: No need to set the default for element 0, as it should take the global default noted above.

internal/coremsgs/en_config_descriptions.go
```diff
@@ -253,6 +253,9 @@ var (
 	ConfigNamespacesDefault               = ffc("config.namespaces.default", "The default namespace - must be in the predefined list", i18n.StringType)
 	ConfigNamespacesPredefined            = ffc("config.namespaces.predefined", "A list of namespaces to ensure exists, without requiring a broadcast from the network", "List "+i18n.StringType)
 	ConfigNamespacesPredefinedName        = ffc("config.namespaces.predefined[].name", "The name of the namespace (must be unique)", i18n.StringType)
+	ConfigNamespacesPredefinedPlugins     = ffc("config.namespaces.predefined[].plugins", "The list of plugins for this namespace", i18n.StringType)
+	ConfigNamespacesPredefinedMode        = ffc("config.namespaces.predefined[].mode", "The namespace mode", i18n.StringType)
```
awrichar, line 257: We need to enumerate the choices here ("multiparty" or "gateway")

internal/namespace/manager.go
```diff
@@ -124,3 +139,129 @@ func (nm *namespaceManager) initNamespaces(ctx context.Context, di database.Plug
 	}
 	return nil
 }
+
+func (nm *namespaceManager) validateNamespaceConfig(ctx context.Context, name string, index int, conf config.Section) error {
+	if err := core.ValidateFFNameField(ctx, name, fmt.Sprintf("namespaces.predefined[%d].name", index)); err != nil {
+		return err
+	}
+
+	if strings.ToLower(name) == "ff_system" || strings.ToLower(conf.GetString(coreconfig.NamespaceRemoteName)) == "ff_system" {
```
awrichar, line 148: Can use the `core.SystemNamespace` constant here

awrichar: Might want to parameterize the error message as well, ie `%s is a reserved name` and then you can pass in `core.SystemNamespace`

internal/namespace/manager.go
```diff
@@ -124,3 +139,129 @@ func (nm *namespaceManager) initNamespaces(ctx context.Context, di database.Plug
...
+		// Still check to ensure at least one bc, dx, ss, and db plugin exist
+		if len(nm.bcPlugins) == 0 || len(nm.dxPlugins) == 0 || len(nm.ssPlugins) == 0 || len(nm.dbPlugins) == 0 {
+			return i18n.NewError(ctx, coremsgs.MsgNamespaceMultipartyConfiguration, name)
+		}
+		return nil
+	}
+
+	switch mode {
+	// Multiparty is the default mode when none is provided
+	case "multiparty", "":
```
awrichar, line 166: If you properly set the default, I _think_ the mode will always be set so you don't have to check for `""`

internal/namespace/manager.go
```diff
@@ -124,3 +139,129 @@ func (nm *namespaceManager) initNamespaces(ctx context.Context, di database.Plug
...
+	if strings.ToLower(name) == "ff_system" || strings.ToLower(conf.GetString(coreconfig.NamespaceRemoteName)) == "ff_system" {
+		return i18n.NewError(ctx, coremsgs.MsgFFSystemReservedName)
+	}
+
+	mode := conf.GetString(coreconfig.NamespaceMode)
+	plugins := conf.GetStringSlice(coreconfig.NamespacePlugins)
+
+	// If the no plugins are found when querying the config, assume older config file
+	if len(plugins) == 0 {
+		// Still check to ensure at least one bc, dx, ss, and db plugin exist
```
awrichar, line 157: In this branch maybe we want to actually iterate all the defined plugins and add them to the `plugins` list, then let it go through the normal validation switch-case below?

awrichar:
It's worth calling out explicitly that this is config validation only - none of the newly added config is actually functional yet (which is fine for now).

Also, I don't see that this bullet is enforced yet for the plugins:
* `name` must be unique on this node

It might be on orchestrator to enforce that before passing the list of plugins to Namespace Manager...

shorsher:
Thanks for the comments and added some new changes which rely on your changes in https://github.com/kaleido-io/firefly-common/tree/arrays

internal/namespace/manager.go
```diff
@@ -124,3 +139,139 @@ func (nm *namespaceManager) initNamespaces(ctx context.Context, di database.Plug
...
+		}
+		if _, ok := nm.tokensPlugins[pluginName]; ok {
+			continue
+		}
+
+		return i18n.NewError(ctx, coremsgs.MsgNamespaceUnknownPlugin, name, pluginName)
+	}
+
+	if !dbPlugin || !ssPlugin || !dxPlugin || !bcPlugin {
+		fmt.Printf("plugins: %+v\n", nm)
```
awrichar, line 233: Remove this debug print?

internal/orchestrator/orchestrator.go
```diff
@@ -358,6 +360,7 @@ func (or *orchestrator) initDatabasePlugins(ctx context.Context, plugins []datab
 		}
 		name := config.GetString(coreconfig.PluginConfigName)
 		or.databases[name] = plugin
+		or.databases[name] = plugin
```
awrichar, line 363: Duplicate line?

internal/coreconfig/coreconfig.go
```diff
@@ -29,6 +29,12 @@ const (
 	PluginConfigName = "name"
 	// PluginConfigType is the type of the plugin to be loaded
 	PluginConfigType = "type"
+	// NamespaceMode is a the configured mode of a namespace
+	NamespaceMode = "mode"
+	// NamespaceRemoteName is a the configured mode of a namespace
```
awrichar, line 34: Comment is wrong

internal/coreconfig/coreconfig.go
```diff
@@ -29,6 +29,12 @@ const (
...
+	NamespaceRemoteName = "remoteName"
+	// NamespacePlugins is a the list of namespace plugins
```
awrichar, line 36: `a the`

internal/namespace/manager.go
```diff
@@ -124,3 +139,138 @@ func (nm *namespaceManager) initNamespaces(ctx context.Context, di database.Plug
 	}
 	return nil
 }
+
+func (nm *namespaceManager) validateNamespaceConfig(ctx context.Context, name string, index int, conf config.Section) error {
+	if err := core.ValidateFFNameField(ctx, name, fmt.Sprintf("namespaces.predefined[%d].name", index)); err != nil {
+		return err
+	}
+
+	if strings.ToLower(name) == core.SystemNamespace || strings.ToLower(conf.GetString(coreconfig.NamespaceRemoteName)) == core.SystemNamespace {
```
awrichar, line 148: Not sure `ToLower` is needed - our namespaces are case-sensitive, right? Although we could choose to disallow any variation of ff_system here.

internal/orchestrator/orchestrator_test.go
```diff
@@ -218,6 +218,7 @@ func TestBadDeprecatedDatabaseInitFail(t *testing.T) {
 
 func TestDatabaseGetPlugins(t *testing.T) {
 	or := newTestOrchestrator()
+	or.pluginNames = make(map[string]bool)
```
awrichar, line 221: This could maybe be moved into `newTestOrchestrator`

awrichar (approved):
Validation looks good. I left a few minor comments which you can check on, but marking approved now.

internal/coreconfig/coreconfig.go
shorsher: spelling is hard

internal/namespace/manager.go
shorsher: wasn't sure, I guess since FF_SYSTEM is technically allowed I'll remove `toLower()`

PR #835, opened 2022-05-25
Update README with missing links, and bit of a restructure

peterbroadhurst:
I found a few links were missing, and given how many repos we've grown to I proposed a bit of organization too.


nguyer (approved):
Looks good. Thanks!

PR #836, opened 2022-05-27
default fabconnect calls to async

shorsher:
The fabconnect default for `/transactions` is synchronous. The fabric plugin currently expects async responses, so adding the `x-firefly-sync` header.


peterbroadhurst (approved):
👍

PR #837, opened 2022-05-27
Use firefly-signer library and allow numbers expressed as strings

nguyer:
This PR introduces `firefly-signer`'s ABI parsing functions and uses them in FireFly's ABI -> FFI generation code. It also replaces any existing ABI data structures that were originally copied over from Ethconnect, with those of `firefly-signer`.

This PR also makes some enhancements to the ABI -> FFI generation code itself to allow numbers to be expressed in a JSON string or a JSON number. This is reflected in the generated OpenAPI Spec for custom contracts, including a description in the Swagger UI that describes this.

Further improvements have also been made to the OpenAPI Spec / Swagger UI for custom smart contracts, to indicate if a function is payable, etc. The `options` object also appears in the schema to make it more obvious how to pass in additional options.

A new `details` field has been added to `FFIMethod` and `FFIEventDefinition` structs. In the same way that a `details` field has always wrapped blockchain specific information for a `FFIParam`, we now have this ability on a function and event level as well. For an Ethereum smart contract, this is used to indicate whether a function is payable, whether an event is anonymous, etc. The details are simply stored as a `fftypes.JSONObject`. In the rendered Swagger UI, a table is constructed to show each field and its corresponding value in the Function / Event description.


internal/blockchain/ethereum/ethereum.go
```diff
@@ -930,121 +920,152 @@ func (e *Ethereum) GenerateFFI(ctx context.Context, generationRequest *core.FFIG
...
+			Params: params,
+		},
 	}
-	return ffiParams
 }
 
-func (e *Ethereum) getSchema(arg ABIArgumentMarshaling) *Schema {
-	s := &Schema{
-		Type: e.getFFIType(arg.Type),
+func (e *Ethereum) getSchemaForABIInput(ctx context.Context, input *abi.Parameter) *Schema {
```
peterbroadhurst, line 996: This function still seems to have a bunch of `[]` related processing, and I could not convince myself that something like `(bytes32[5][][10],fixed128x18)` would be handled completely.

None of this should be required with the new API available, it should be possible to simply recurse the parsed component type tree structure:
https://pkg.go.dev/github.com/hyperledger/firefly-signer/pkg/abi#ParameterArray.TypeComponentTreeCtx

peterbroadhurst (changes requested):
@nguyer - I paused the review when I hit quite a big question, about whether you'd used the functions introduced in the ABI parsing structure specifically to handle features like OpenAPI generation. Could you take a look and see if there's more to do, either in Core, or in firefly-signer, to ensure identical recursive processing on the metadata generation and runtime?

nguyer:
Just realized there are some major merge conflicts here, including some changes that actually need to go into `firefly-common`. I'll sort those before requesting a re-review.

internal/apiserver/ffi2swagger.go
```diff
@@ -73,34 +77,45 @@ func (og *ffiSwaggerGen) Generate(ctx context.Context, baseURL string, api *core
...
+		Description:           ffi.Description,
+		BaseURL:               baseURL,
+		DefaultRequestTimeout: config.GetDuration(coreconfig.APIRequestTimeout),
 	}).Generate(ctx, routes)
 }
 
 func (og *ffiSwaggerGen) addMethod(routes []*ffapi.Route, method *core.FFIMethod, hasLocation bool) []*ffapi.Route {
+	description := method.Description
+	if method.Details != nil && len(method.Details) > 0 {
+		description = fmt.Sprintf("%s\n\nAdditional smart contract details:\n\n%s", description, buildDetailsTable(method.Details))
```
peterbroadhurst, line 91: Struggling a little to understand this. What is the end to end of the `details`? Is it the place where `devdocs` would be specified?

One specific item (along with the wider question) is the `Additional smart contract details` string needs to be translated.

nguyer: Sure! I did forget to mention that in the description, sorry about that. The short version is that it's a way to bubble up blockchain specifics about a function, like if it's `payable` or not.

Good call out on the string translation, thanks.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -797,38 +794,59 @@ func (e *Ethereum) GetFFIParamValidator(ctx context.Context) (core.FFIParamValid
...
+	if method.Details != nil {
+		stateMutability, ok := method.Details["stateMutability"]
+		if ok {
+			stateMutabilityString, ok := stateMutability.(string)
+			if ok {
+				abiEntry.StateMutability = abi.StateMutability(stateMutabilityString)
+			}
+		}
+		if payable, ok := method.Details["payable"]; ok {
+			abiEntry.Payable = payable.(bool)
```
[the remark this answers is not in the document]
nguyer: Ah good suggestion. I'll change this one and the similar checks for other fields.

internal/blockchain/ethereum/ffi_param_validator.go
```diff
@@ -33,31 +33,40 @@ var bytesRegex, _ = regexp.Compile("^bytes([0-9]{1,2})?")
 func (v *FFIParamValidator) Compile(ctx jsonschema.CompilerContext, m map[string]interface{}) (jsonschema.ExtSchema, error) {
```
peterbroadhurst, line 33: I was trying to work out where this is used. I think we agreed that the connector would be responsible for verifying input, and that we should ensure that happens synchronously before we give back a `202` to an invoke.

Understand this PR might not be the final step in the chain, but I want to make sure we're not trying to replicate the connector input validation logic in core.

nguyer: The most important use of this is in the contracts manager https://github.com/hyperledger/firefly/blob/e8419846ccd8c6e80791540a06881f3385ace2da/internal/contracts/manager.go#L112-L118

Once this JSON Schema extension is registered, the JSON Schema library calls this function when a request is made to broadcast a new FFI. This ensures that the schema specified in the FFI for each field adheres to the specific rules defined by each blockchain connector. As an example, the Ethereum connector requires the `details` object in most places, and it requires that a `type` field is set inside there (in addition to a bunch of other schema rules).

In other words, this is the metaschema (the schema for the schema). This code is _not_ for validating blockchain invocation request input.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -956,121 +981,167 @@ func (e *Ethereum) GenerateFFI(ctx context.Context, generationRequest *core.FFIG
...
+	schema := &Schema{
+		Details: &paramDetails{
+			Type:         typeComponent.Parameter().Type,
+			InternalType: typeComponent.Parameter().InternalType,
+			Indexed:      typeComponent.Parameter().Indexed,
+		},
+	}
+	switch typeComponent.ComponentType() {
+	case abi.ElementaryComponent:
+		t := e.getFFIType(typeComponent.ElementaryType().String())
```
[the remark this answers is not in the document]
nguyer: I think we still need this bit of code, because somewhere we need a mapping from ABI type (e.g. `uint256`) to the JSON input type we will accept on the API (e.g. `integer`)

PR #840, opened 2022-05-27
Split Batch Manager to have one Assembler per database plugin

awrichar:
Part of https://github.com/hyperledger/firefly-fir/pull/12


awrichar:
Closing as I think the latest direction may just involve multiple batch managers, instead of introducing this middle layer.

PR #842, opened 2022-05-27
Move /admin/api to /spi and PUT->PATCH

peterbroadhurst:
> Leaving this as a draft until the updates to FFTM are complete, because otherwise early adopters trying out the feature will not have a functioning combination.

- [x] Rename the `/admin/api/v1` API to `/spi/v1`
- [x] Include `namespace:` prefix on all request IDs to all plugins, so that FFTM can pass the NS back to find/update the op
  - [x] Blockchain
  - [x] Data Exchange
  - ~Shared Storage~ - no async ops
  - [x] Tokens
- [x] `PUT`s change to `PATCH`s, because there are race conditions between delete/update that mean PUT semantics wouldn't be safe without optimistic concurrency locking
- [x] Make sure all APIs FFTM needs to function are on this API, so it doesn't need to use two
  - [x] `GET` `/spi/v1/operations`
  - [x] `GET` `/spi/v1/operations/{nsopid}`
  - [x] `PATCH` `/spi/v1/operations/{nsopid}`
  - [x] `GET` `/spi/v1/namespaces`
  - [x] `GET` `/spi/v1/namespaces/{ns}`
- [x]  Continue to allow FFTM to use cross-namespace calls on the internal API
- [x] Add namespaces to the API that FFTM has

> Previously we discussed additional changes, to allow event listeners hosted in FFTM to be stateless - and avoid the
> need to have a REST API + LevelDB in FFTM to manage listeners/checkpoints. This worked out to be too complex
> to change right now. Specifically because:
> 1. BatchPin is not tracked as a listener
> 2. Token connectors use the APIs of EthConnect today to directly create event streams

Note thread in discord: https://discord.com/channels/905194001349627914/979733575496790086

### DIFF
--- a/internal/coremsgs/en_api_translations.go
+++ b/internal/coremsgs/en_api_translations.go
@@ -23,7 +23,7 @@ var (
 	APIParamsConfigRecordKeyUpdate          = ffm("api.params.configRecordKey.update", "The configuration key to update. This should use dot notation to reference a key documented in https://hyperledger.github.io/firefly/reference/config.html")
 	APIParamsConfigRecordKeyGet             = ffm("api.params.configRecordKey.get", "The configuration key to get. This should use dot notation to reference a key documented in https://hyperledger.github.io/firefly/reference/config.html")
 	APIParamsOperationIDGet                 = ffm("api.params.operationID.get", "The operation ID key to get")
-	APIParamsOperationNamespacedID          = ffm("api.params.spiOperationID", "The operation ID as passed to the connector when the operation was performed, including the 'ns!' prefix")
+	APIParamsOperationNamespacedID          = ffm("api.params.spiOperationID", "The operation ID as passed to the connector when the operation was performed, including the 'namespace:' prefix")
 	APIParamsNamespace                      = ffm("api.params.namespace", "The namespace which scopes this request")
 	APIParamsContractListenerNameOrID       = ffm("api.params.contractListenerNameOrID", "The contract listener name or ID")
 	APIParamsContractListenerID             = ffm("api.params.contractListenerID", "The contract listener ID")
--- a/internal/operations/manager.go
+++ b/internal/operations/manager.go
@@ -45,7 +45,6 @@ type Manager interface {
 	AddOrReuseOperation(ctx context.Context, op *core.Operation) error
 	SubmitOperationUpdate(plugin core.Named, update *OperationUpdate)
 	TransferResult(dx dataexchange.Plugin, event dataexchange.DXEvent)
-	ResolveOperationByID(ctx context.Context, ns, id string, op *core.OperationUpdateDTO) error
 	ResolveOperationByNamespacedID(ctx context.Context, nsOpID string, op *core.OperationUpdateDTO) error
 	Start() error
 	WaitStop()
@@ -211,15 +210,6 @@ func (om *operationsManager) writeOperationFailure(ctx context.Context, ns strin
 
 func (om *operationsManager) ResolveOperationByNamespacedID(ctx context.Context, nsOpID string, op *core.OperationUpdateDTO) error {
 	ns, u, err := core.ParseNamespacedOpID(ctx, nsOpID)
-	if err != nil {
-		return err
-	}
-	err = om.database.ResolveOperation(ctx, ns, u, op.Status, op.Error, op.Output)
-	return err
-}
-
-func (om *operationsManager) ResolveOperationByID(ctx context.Context, ns, id string, op *core.OperationUpdateDTO) error {
-	u, err := fftypes.ParseUUID(ctx, id)
 	if err != nil {
 		return err
 	}
--- a/internal/operations/manager_test.go
+++ b/internal/operations/manager_test.go
@@ -545,46 +545,6 @@ func TestTransferResultBatchLookupFail(t *testing.T) {
 
 }
 
-func TestResolveOperationByIDOk(t *testing.T) {
-	om, cancel := newTestOperations(t)
-	defer cancel()
-
-	ctx := context.Background()
-	opID := fftypes.NewUUID()
-	errStr := "my error"
-	opUpdate := &core.OperationUpdateDTO{
-		Status: core.OpStatusSucceeded,
-		Error:  &errStr,
-		Output: fftypes.JSONObject{
-			"my": "data",
-		},
-	}
-
-	mdi := om.database.(*databasemocks.Plugin)
-	mdi.On("ResolveOperation", ctx, "ns1", opID, core.OpStatusSucceeded, &errStr, fftypes.JSONObject{
-		"my": "data",
-	}).Return(nil)
-
-	err := om.ResolveOperationByID(ctx, "ns1", opID.String(), opUpdate)
-
-	assert.NoError(t, err)
-
-	mdi.AssertExpectations(t)
-}
-
-func TestResolveOperationBadID(t *testing.T) {
-	om, cancel := newTestOperations(t)
-	defer cancel()
-
-	ctx := context.Background()
-	op := &core.OperationUpdateDTO{}
-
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -69,7 +69,7 @@ $(eval $(call makemock, internal/namespace,        Manager,            namespace
 $(eval $(call makemock, internal/networkmap,       Manager,            networkmapmocks))
 $(eval $(call makemock, internal/assets,           Manager,            assetmocks))
 $(eval $(call makemock, internal/contracts,        Manager,            contractmocks))
-$(eval $(call makemock, internal/adminevents,      Manager,            admineventsmocks))
+$(eval $(call makemock, internal/spievents,      Manager,            spieventsmocks))
```
awrichar, line 72: nit: indentation

internal/database/sqlcommon/contractlisteners_sql.go
```diff
@@ -42,6 +42,7 @@ var (
 		"signature",
 		"topic",
 		"options",
+		"state",
 		"created",
 	}
```
awrichar, line 47: I expected to see an "updated" field here (it's in the database migration and in the query factory).

internal/contracts/manager.go
```diff
@@ -58,7 +58,10 @@ type Manager interface {
 	AddContractAPIListener(ctx context.Context, ns, apiName, eventPath string, listener *core.ContractListener) (output *core.ContractListener, err error)
 	GetContractListenerByNameOrID(ctx context.Context, ns, nameOrID string) (*core.ContractListener, error)
 	GetContractListeners(ctx context.Context, ns string, filter database.AndFilter) ([]*core.ContractListener, *database.FilterResult, error)
+	GetContractListenerByIDGlobal(ctx context.Context, nameOrID string) (*core.ContractListener, error)
+	GetContractListenersGlobal(ctx context.Context, filter database.AndFilter) ([]*core.ContractListener, *database.FilterResult, error)
```
awrichar, line 62: Global collections are going away very soon (everything needs to be retrieved within a namespace)

pkg/core/contract_listener.go
```diff
@@ -37,6 +37,12 @@ type ContractListener struct {
 	Signature string                   `ffstruct:"ContractListener" json:"signature" ffexcludeinput:"true"`
 	Topic     string                   `ffstruct:"ContractListener" json:"topic,omitempty"`
 	Options   *ContractListenerOptions `ffstruct:"ContractListener" json:"options,omitempty"`
+	State     *fftypes.JSONAny         `ffstruct:"ContractListener" json:"state,omitempty"`
+}
```
awrichar, line 41: Shouldn't this have an "Updated" timestamp?

internal/database/sqlcommon/operation_sql.go
```diff
@@ -181,22 +181,29 @@ func (s *SQLCommon) UpdateOperation(ctx context.Context, ns string, id *fftypes.
...
 func (s *SQLCommon) ResolveOperation(ctx context.Context, ns string, id *fftypes.UUID, status core.OpStatus, errorMsg string, output fftypes.JSONObject) (err error) {
-	update := database.OperationQueryFactory.NewUpdate(ctx).
-		Set("status", status).
-		Set("error", errorMsg)
+	update := database.OperationQueryFactory.NewUpdate(ctx).S()
+	if status != "" {
+		update = update.Set("status", status)
+	}
+	if errorMsg != "" {
+		update = update.Set("error", errorMsg)
```
awrichar, line 203: Looks like it is not possible to clear an error message once one is set. It's probably fine, but just pointing it out.

pkg/core/contract_listener.go
peterbroadhurst: Sorry, this one needs to come out before this merges. I had started adding the ability for FFTM to perform updates to contract listeners... before realizing that wasn't practical as a next step due to the direct creation of listeners in EthConnect today by tokens. So I will remove this field.

internal/database/sqlcommon/operation_sql.go
peterbroadhurst: This is a good point. Now splitting out to an update DTO, we can have a string pointer to make this distinction.

internal/database/sqlcommon/contractlisteners_sql.go
peterbroadhurst: This is removed now

internal/contracts/manager.go
peterbroadhurst: This is going (only the cross-namespace operations query will be retained at the end of this PR - which FFTM relies upon for the very short term)

peterbroadhurst:
ok - merge battled, and I've removed all the unnecessary bits around contract listeners. Also addressed the ability to set the error message to the empty string.

peterbroadhurst:
I've put this back in draft, as per thread https://discord.com/channels/905194001349627914/983341711742885948 we need to make the `id` passed to connectors when an operation is performed, include the `namespace` as well as the `UUID` of the operation.

So `0df3d864-2646-4e5d-8585-51eb154a8d23` becomes `default:0df3d864-2646-4e5d-8585-51eb154a8d23`

internal/coremsgs/en_api_translations.go
```diff
@@ -23,6 +23,7 @@ var (
 	APIParamsConfigRecordKeyUpdate          = ffm("api.params.configRecordKey.update", "The configuration key to update. This should use dot notation to reference a key documented in https://hyperledger.github.io/firefly/reference/config.html")
 	APIParamsConfigRecordKeyGet             = ffm("api.params.configRecordKey.get", "The configuration key to get. This should use dot notation to reference a key documented in https://hyperledger.github.io/firefly/reference/config.html")
 	APIParamsOperationIDGet                 = ffm("api.params.operationID.get", "The operation ID key to get")
+	APIParamsOperationNamespacedID          = ffm("api.params.spiOperationID", "The operation ID as passed to the connector when the operation was performed, including the 'ns!' prefix")
```
awrichar, line 26: Prefix is now `ns:`, right?

internal/operations/manager.go
```diff
@@ -46,6 +46,7 @@ type Manager interface {
 	SubmitOperationUpdate(plugin core.Named, update *OperationUpdate)
 	TransferResult(dx dataexchange.Plugin, event dataexchange.DXEvent)
 	ResolveOperationByID(ctx context.Context, ns, id string, op *core.OperationUpdateDTO) error
```
awrichar, line 48: Is this method used, or is it totally replaced by the new one below?

internal/orchestrator/orchestrator.go
```diff
@@ -899,5 +900,10 @@ func (or *orchestrator) SubmitNetworkAction(ctx context.Context, ns string, acti
 	} else {
 		return i18n.NewError(ctx, coremsgs.MsgUnrecognizedNetworkAction, action.Type)
 	}
-	return or.blockchain.SubmitNetworkAction(ctx, fftypes.NewUUID(), key, action.Type)
+	// TODO: This should be a new operation type
```
awrichar, line 903: ack

awrichar (approved):
Looks good for a step forward that doesn't break core functionality 👍 Just noted some minor things inline.

internal/apiserver/server.go
```diff
@@ -300,6 +299,11 @@ func (as *apiServer) createMuxRouter(ctx context.Context, o orchestrator.Orchest
 	return r
 }
 
+func (as *apiServer) notFoundHandler(res http.ResponseWriter, req *http.Request) (status int, err error) {
```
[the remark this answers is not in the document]
peterbroadhurst: Yep, we set it on the router here: 
https://github.com/hyperledger/firefly/blob/85f10558ef898cdef4e1b3eca7c7636e94e6c335/internal/apiserver/server.go#L298

PR #843, opened 2022-05-31
Verify namespace for all broadcast/private message requests

### LINKED ISSUE #817: Broadcast on non-existing namespace returns 202
broadcasting a message to a namespace that has not yet been created should return an error. Currently the API call returns 202.

awrichar:
Fixes #817

### DIFF
--- a/internal/broadcast/message.go
+++ b/internal/broadcast/message.go
@@ -115,6 +115,10 @@ func (s *broadcastSender) resolveAndSend(ctx context.Context, method sendMethod)
 func (s *broadcastSender) resolve(ctx context.Context) error {
 	msg := s.msg.Message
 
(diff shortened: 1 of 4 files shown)


peterbroadhurst (changes requested):
Would you mind adding the LRU cache for NS to this code at this point?
Should be very simple to do according to the pattern we have in place for other caches.
https://github.com/hyperledger/firefly/blob/4b99553381a382cbf097cce4d9a124239c7bc233/internal/data/data_manager.go#L142

awrichar:
Added caching for namespace existence checks only (simple since namespaces cannot be deleted).

peterbroadhurst (approved):
Thanks for adding the cache

PR #844, opened 2022-05-31
Allow extra options to pass through to blockchain connectors

nguyer:
This PR adds the ability to pass additional options through to blockchain connectors. This is useful for controlling other parameters when submitting a blockchain transaction besides the specific arguments that a smart contract function takes. For example, this can now be used to set a custom gas price

Screen Shot 2022-05-31 at 11 39 25 AM

Anywhere that a smart contract can be invoked or queried a new `options` JSON object (treated as a `map[string]interface{}` in Go) can be optionally specified. Any fields in this structure will be included in the request to the blockchain connector.

Overriding existing fields in the request is **not** allowed. For example, if using Ethconnect, the following request body would be rejected because `headers` is a field that FireFly itself needs to set:

```json
{
  "input": {
    "newValue": "x"
  },
  "options": {
    "headers": "not allowed"
  }
}
```
This would result in an HTTP 400 with an error message indicating which option was not allowed.

### DIFF
--- a/docs/config_docs_generate_test.go
+++ b/docs/config_docs_generate_test.go
@@ -14,6 +14,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build reference
+// +build reference
+
 package config
 
 import (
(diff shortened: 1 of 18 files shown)


peterbroadhurst (approved):
This is great. I like the override protection

PR #849, opened 2022-06-02
Enhancements to migration/network version behavior

awrichar:
Network version is loaded during blockchain plugin init, instead of lazily. This is (soon) going to be needed at manager initialization time, so might as well load it once upfront.

Since we currently have a single blockchain plugin that services all namespaces, any termination/migration event is recorded against ALL namespaces (rather than arbitrarily recording it against ff_system only).

### DIFF
--- a/internal/events/network_action.go
+++ b/internal/events/network_action.go
@@ -17,20 +17,34 @@
 package events
 
 import (
+	"context"
+
 	"github.com/hyperledger/firefly-common/pkg/log"
 	"github.com/hyperledger/firefly/pkg/blockchain"
 	"github.com/hyperledger/firefly/pkg/core"
+	"github.com/hyperledger/firefly/pkg/database"
 )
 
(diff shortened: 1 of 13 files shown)


internal/events/network_action.go
```diff
@@ -22,18 +22,28 @@ import (
...
+	}
+	// Currently, a termination event is implied to apply to ALL namespaces
 	return em.database.RunAsGroup(em.ctx, func(ctx context.Context) error {
-		ns, err := em.database.GetNamespace(ctx, core.LegacySystemNamespace)
-		if err != nil {
-			return err
-		}
-		if err := bi.TerminateContract(ctx, &ns.Contracts, event); err != nil {
-			return err
+		for _, ns := range namespaces {
```
peterbroadhurst, line 40: Do we have a "current version is v1" check around this?

This feels like an action we should reject apart from the one case of `ff_system` (v1) to per-namespace migration.

peterbroadhurst (approved):
Thanks @awrichar - the picture helped 🖍️ 💡

PR #852, opened 2022-06-03
Fix github.ref in prereleased action

nguyer:


awrichar (approved):
Sure. Do we need this on any other branch?

PR #855, opened 2022-06-09
Use a separate orchestrator for each namespace

awrichar:
Changed in this PR:
- Namespace manager is now the root manager in the system
- There is a separate orchestrator for each namespace, along with a separate instance of (almost) every manager
- Config for including different plugins in each namespace is now honored
- Each plugin supports multiple listeners, and each listener function (mostly on event manager) will filter out events from just the one namespace to which it's assigned

Broken in this PR:
- You cannot have multiple namespaces that point at the same blockchain FireFly contract, or the same token pool. They will end up sharing an event stream and therefore only one of them will process events. Fixing this will require splitting a lot of the contract/event stream logic out of the blockchain plugins into a new per-namespace component.
- Performance of event processing degrades almost linearly with the number of created namespaces, due to the fact that most events are fired to _all_ instances of the event manager and then examined in order to process or ignore them. In the future, we can push namespace knowledge down into lower layers to make this decision earlier and avoid this extra overhead.
- SPI route for "get all operations" is moved from `/operations` to `/namespaces/{ns}/operations`, which represents a breaking change against FFTM.

~~Also includes the changes from #856.~~


internal/dataexchange/ffdx/dxevent.go
```diff
@@ -154,8 +156,14 @@ func (h *FFDX) dispatchEvent(msg *wsEvent) {
 		var hash *fftypes.Bytes32
 		hash, err = fftypes.ParseBytes32(h.ctx, msg.Hash)
 		if err == nil {
+			var ns string
+			pathParts := strings.Split(msg.Path, "/")
+			if len(pathParts) >= 2 {
+				ns = pathParts[len(pathParts)-2]
```
awrichar, line 162: This seems to work, but feels a bit fragile. I wonder if namespaces should be more explicitly included on dx blob events?

internal/namespace/manager.go
```diff
@@ -21,271 +21,830 @@ import (
...
+		identity      map[string]identityPlugin
+		database      map[string]databasePlugin
+		sharedstorage map[string]sharedStoragePlugin
+		dataexchange  map[string]dataExchangePlugin
+		tokens        map[string]tokensPlugin
+	}
+	metricsEnabled bool
+	metrics        metrics.Manager
+	adminEvents    spievents.Manager
+	utOrchestrator orchestrator.Orchestrator
```
awrichar, line 97: I don't love this testing workaround, but I was struggling to find a better way.

internal/apiserver/route_spi_get_ops.go
```diff
@@ -37,6 +37,7 @@ var spiGetOps = &ffapi.Route{
 	Extensions: &coreExtensions{
 		FilterFactory: database.OperationQueryFactory,
 		CoreJSONHandler: func(r *ffapi.APIRequest, cr *coreRequest) (output interface{}, err error) {
+			// TODO: cr.or will be nil (this must be converted to a namespaced query)
```
peterbroadhurst, line 40: Should we not do this in this PR? Having a broken endpoint seems bad, and it should be fine to make it `operations/{ns}`

peterbroadhurst: (We still need the `operations/{nsopid}` routes for `GET` and `PATCH`, where the namespace is embedded into the ID)

internal/dataexchange/ffdx/dxevent.go
peterbroadhurst: It's a good point, in-line with the work on other microservice connectors to make them namespace aware at the API boundary.

internal/events/dx_callbacks.go
```diff
@@ -218,6 +222,10 @@ func (em *eventManager) privateBlobReceived(dx dataexchange.Plugin, event dataex
 		event.Ack() // Still confirm the event
 		return
 	}
+	if br.Namespace != em.namespace {
+		log.L(em.ctx).Debugf("Ignoring blob from wrong namespace '%s'", br.Namespace)
```
peterbroadhurst, line 226: Thing a change from `wrong` to `different` might help users here, considering how many of these log entries they could see in their logs.

Comment applies to all the `wrong namespace` event entries, not just DX

peterbroadhurst (approved):
Thanks for calling out all the considerations for where we end up after this (big!) step forwards. Looks good. I did have a couple of very minor comments along the way to review before merge.

internal/apiserver/route_spi_get_ops.go
awrichar: Yep that works - I can replace that route. Just wanted to get your feedback on how to spell it in the context of the other SPI routes and the usage by FFTM.

PR #856, opened 2022-06-09
Add namespace to pins

awrichar:
Existing pins will all be categorized against "ff_system", just so the field has a non-null value.

### DIFF
--- a/db/migrations/postgres/000092_add_pin_namespace.up.sql
+++ b/db/migrations/postgres/000092_add_pin_namespace.up.sql
@@ -1,0 +1,5 @@
+BEGIN;
+ALTER TABLE pins ADD COLUMN namespace VARCHAR(64);
+UPDATE pins SET namespace = 'ff_system';
+ALTER TABLE pins ALTER COLUMN namespace SET NOT NULL;
+COMMIT;
--- a/db/migrations/sqlite/000092_add_pin_namespace.up.sql
+++ b/db/migrations/sqlite/000092_add_pin_namespace.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE pins ADD COLUMN namespace VARCHAR(64);
+UPDATE pins SET namespace = "ff_system";
--- a/internal/apiserver/route_get_pins.go
+++ b/internal/apiserver/route_get_pins.go
@@ -25,13 +25,13 @@ import (
 	"github.com/hyperledger/firefly/pkg/database"
 )
 
-var getStatusPins = &ffapi.Route{
-	Name:            "getStatusPins",
-	Path:            "status/pins",
+var getPins = &ffapi.Route{
+	Name:            "getPins",
+	Path:            "pins",
 	Method:          http.MethodGet,
 	PathParams:      nil,
 	QueryParams:     nil,
-	Description:     coremsgs.APIEndpointsGetStatusPins,
+	Description:     coremsgs.APIEndpointsGetPins,
 	JSONInputValue:  nil,
 	JSONOutputValue: func() interface{} { return []core.Pin{} },
 	JSONOutputCodes: []int{http.StatusOK},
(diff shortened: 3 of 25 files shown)


db/migrations/postgres/000092_add_pin_namespace.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+ALTER TABLE pins ADD COLUMN namespace VARCHAR(64);
+UPDATE pins SET namespace = 'ff_system';
```
peterbroadhurst, line 3: I've added a `migration_consideration` tag, as we've chosen `ff_system` really only because it's a reserved namespace name that can be used to query these values back (definitely *not* because these pre-upgrade pins were all for the `ff_system` namespace).

So we should document that.

db/migrations/sqlite/000092_add_pin_namespace.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE pins ADD COLUMN namespace VARCHAR(64);
+UPDATE pins SET namespace = "ff_system";
```
peterbroadhurst, line 2: Do you think we should we update the indexes to have `namespace` as a dimension?

I've noticed we did that elsewhere when we introduced a namespace, and I guess it should help with isolation of performance considerations between namespaces in the case of a shared DB.

internal/apiserver/route_get_pins.go
```diff
@@ -25,20 +25,20 @@ import (
 	"github.com/hyperledger/firefly/pkg/database"
 )
 
-var getStatusPins = &ffapi.Route{
-	Name:            "getStatusPins",
-	Path:            "status/pins",
+var getPins = &ffapi.Route{
+	Name:            "getPins",
```
peterbroadhurst, line 29: Also need to cover the moving of the API in the migration docs

db/migrations/sqlite/000092_add_pin_namespace.up.sql
peterbroadhurst: As part of answering this, I think it would be good to look at this code:
https://github.com/hyperledger/firefly/blob/eb61c4c6d21b4d85152f69ee94fd946c9b8802d4/internal/events/aggregator.go#L224-L232

If the event aggregator is to become namespace aware, then I would expect the `filter` passed into this function by the parent to include a `namespace` filter. I'm not sure if that's true today.

peterbroadhurst (commented):
Code looks great to me. I did have a couple of questions for you to ponder whether there's more to do before this merges.

db/migrations/postgres/000092_add_pin_namespace.up.sql
awrichar: Yea, originally I was going to leave it blank but then realized the API would give you no way to query them (and that seemed bad).

Pins were treated as "global" in 1.0, and the aggregator loop that processes pins even logs them as if they are part of `ff_system` (see [here](https://github.com/hyperledger/firefly/blob/30b0c569639a8b74c5b18144f5779ff39190ac16/internal/events/aggregator.go#L93)), so it didn't feel terribly wrong to pick `ff_system` for this purpose. But it was always a bit inaccurate, as they are specific to a namespace even if they weren't treated that way.

Anyway, bit of a ramble but I agree we need to document it.

db/migrations/sqlite/000092_add_pin_namespace.up.sql
awrichar: Good catch, and I think this should be rolled into #855 where the aggregator becomes namespace-aware.

peterbroadhurst: Can we do that without introducing yet another migration file?

awrichar: Yes, can do. We've said migrations aren't guaranteed to be stable in mainline (only between releases).

awrichar: See https://github.com/hyperledger/firefly/pull/855/commits/f8ef5664f83fb03f71bd5951438bcab49079c20d

PR #862, opened 2022-06-14
Remove namespace from identity manager, network map, and group manager calls

awrichar:

Each manager is initialized for a single namespace and can assume all calls are scoped within that namespace This PR tackles identity manager, network manager, and group manager so they store and use that single namespace.

The one exception is for "network version 1", where identities may have been registered on the legacy "ff_system" namespace. In this case, the identity manager will query "ff_system" instead of its assigned namespace (noting that this will only work if "ff_system" shares a database with that namespace).


internal/identity/identitymanager.go
```diff
@@ -389,12 +388,13 @@ func (im *identityManager) cachedIdentityLookupByVerifierRef(ctx context.Context
 		return nil, err
 	} else if verifier == nil {
 		if namespace != core.LegacySystemNamespace && im.blockchain.NetworkVersion() == 1 {
-			// For V1 networks, fall back to SystemNamespace for looking up identities
+			// For V1 networks, fall back to LegacySystemNamespace for looking up identities
+			// This assumes that the system namespace shares a database with this manager's namespace!
 			return im.cachedIdentityLookupByVerifierRef(ctx, core.LegacySystemNamespace, verifierRef)
```
awrichar, line 393: This is the one location where a manager is performing a lookup _across_ namespaces to something other than the one to which it is scoped. As noted, it will only work if the database plugin is shared between this manager's namespace and the ff_system namespace. ff_system (when enabled) always uses a copy of the plugins for the default namespace.

The alternative I considered was giving some sort of global access to the ff_system orchestrator and identity manager, so that other identity managers could specifically query them... but it seemed heavyweight for what is a soon-to-be deprecated behavior.

peterbroadhurst (approved):
👍  - logic makes sense to me on the `ff_system` cross-ns identity lookup.

PR #864, opened 2022-06-15
Further cleanup of namespace params to managers and database plugin

awrichar:
In a chain with #862

Each manager is specific to a namespace, which should be received and stored during init. All operations conducted by that manager should use that namespace, and therefore no manager methods should accept a namespace as a parameter.

The database plugin, on the other hand, may potentially service multiple namespaces. Therefore, all query operations should take a namespace parameter to ensure that only results from a single namespace are returned.

This PR makes these true for many cases, but others are still to be addressed with a follow-up PR.


peterbroadhurst (approved):
Reviewed all the changes in the commits after #862 👍  (looks like Git is happy to work out the merge)

PR #872, opened 2022-06-21
Resolve DX operations via requestID (not event ID)

### LINKED ISSUE #871: All DX operations get stuck in pending state
Currently all data exchange operations (send batch, etc) remain stuck in "Pending" state indefinitely. Seems to be a side-effect of https://github.com/hyperledger/firefly/pull/842.

awrichar:
Fixes #871 

Also enhance all E2E test suites to check for stuck operations, as this seems to be a recurring theme that slips through the cracks.

### DIFF
--- a/internal/dataexchange/ffdx/dxevent.go
+++ b/internal/dataexchange/ffdx/dxevent.go
@@ -51,6 +51,10 @@ type dxEvent struct {
 	transferResult      *dataexchange.TransferResult
 }
 
+func (e *dxEvent) EventID() string {
+	return e.id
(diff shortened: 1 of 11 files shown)


awrichar:
Looks like for the unpinned message case, the message may get marked "confirmed" before the data exchange operation has recorded success. I assume it's due to operation updates being handle by a separate processing loop. I'm not sure of the best way to handle this... perhaps polling or listening to database change events...

peterbroadhurst (approved):
👍

PR #865, opened 2022-06-15
Create firefly contract subscription per namespace

shorsher:
Each namespace now has a unique firefly contract subscription

To support a subscription per namespace, the `fireflyContract` config section has been replaced with the `contracts` array section under the `multiparty` config.

Example:
```
blockchain:
- name: blockchain0
  type: ethereum
  ethereum:
    ethconnect:
      url: http://ethconnect_0:8080
      topic: "0"
      
namespaces:
  default: default
  predefined:
  - name: default
    remoteName: default
    description: Default predefined namespace
    plugins: [database0, blockchain0, dataexchange0, sharedstorage0, erc20_erc721]
    multiparty:
      enabled: true
      org:
        name: org0
        description: org0
        key: 0x123456
      contract:
        - location:
            address: 0x4ae50189462b0e5d52285f59929d037f790771a6 
          firstEvent: oldest
```

 * location is a yaml object containing blockchain plugin configuration
 * `firstEvent` is replacing `fromBlock`, supporting values of newest or oldest

This PR also introduces a new component, multiparty manager, that is responsible for managing the active firefly contract and interfacing with the blockchain plugin on behalf of that contract.

The subscription names will now also contain the name of the namespace it's associated with:
 * Ethereum: `namespaceName_abiName_instanceUniqueHash`
 * Fabric: `namespaceName_event`

PR #874, opened 2022-06-22
Convert more callback handlers to be namespace-specific

awrichar:
Part of [FIR-12](https://github.com/hyperledger/firefly-fir/pull/12).
Follow-up to #863.

Only the blockchain plugin remains, and will need to be addressed after #865.


internal/events/dx_callbacks.go
```diff
@@ -224,6 +211,7 @@ func (em *eventManager) privateBlobReceived(dx dataexchange.Plugin, event dataex
 	}
 	if br.Namespace != em.namespace {
 		log.L(em.ctx).Debugf("Ignoring blob from different namespace '%s'", br.Namespace)
+		event.Ack() // Still confirm the event
```
peterbroadhurst, line 214: Are we guaranteed to only pass the event to a single DX Callback handler? ... if not, what is the implication that `event.Ack()` could be called multiple times?

peterbroadhurst: ... I remember we made changes to the event ack interface to FFDX, but I can't remember if that included an ID in the ack such that a duplicate was ignored. Feel like maybe I should check that

peterbroadhurst: Yep - all good, the FFDX API was upgraded to ack-by-id:
https://github.com/hyperledger/firefly-dataexchange-https/blob/10b23db379753b83c40da210fb86ceea4226cb42/src/handlers/events.ts#L100-L120

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -50,13 +51,33 @@ type FFDX struct {
...
 
-func (cb *callbacks) DXEvent(event dataexchange.DXEvent) {
-	for _, cb := range cb.handlers {
-		cb.DXEvent(event)
+func (cb *callbacks) DXEvent(namespace string, event dataexchange.DXEvent) error {
+	if handler, ok := cb.handlers[namespace]; ok {
+		handler.DXEvent(event)
+		return nil
 	}
+	return fmt.Errorf("unknown namespace on event '%s'", event.EventID())
```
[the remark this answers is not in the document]
awrichar: Pushed a tweak to address this

PR #876, opened 2022-06-22
Handle namespace validation at the route level rather than at data manager

awrichar:
Data manager is already part of a namespace, so determination of whether the namespace is valid needs to happen earlier in the router, when we're attempting to lookup the proper orchestrator to handle the request.

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -200,7 +200,11 @@ func (as *apiServer) swaggerGenerator(routes []*ffapi.Route, apiBaseURL string) 
 func (as *apiServer) contractSwaggerGenerator(mgr namespace.Manager, apiBaseURL string) func(req *http.Request) (*openapi3.T, error) {
 	return func(req *http.Request) (*openapi3.T, error) {
 		vars := mux.Vars(req)
-		cm := mgr.Orchestrator(vars["ns"]).Contracts()
+		or := mgr.Orchestrator(vars["ns"])
+		if or == nil {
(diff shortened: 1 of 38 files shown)


internal/apiserver/server.go
```diff
@@ -218,17 +222,22 @@ func (as *apiServer) contractSwaggerGenerator(mgr namespace.Manager, apiBaseURL
 	}
 }
 
-func getOrchestrator(mgr namespace.Manager, tag string, r *ffapi.APIRequest) orchestrator.Orchestrator {
-	if tag == routeTagDefaultNamespace {
-		return mgr.Orchestrator(config.GetString(coreconfig.NamespacesDefault))
-	}
-	if tag == routeTagNonDefaultNamespace {
+func getOrchestrator(ctx context.Context, mgr namespace.Manager, tag string, r *ffapi.APIRequest) (or orchestrator.Orchestrator, err error) {
+	switch {
```
shorsher, line 226: a nit, but I think you could do

```
switch tag {
	case routeTagDefaultNamespace:
		or = mgr.Orchestrator(config.GetString(coreconfig.NamespacesDefault))
	case routeTagNonDefaultNamespace:
		vars := mux.Vars(r.Req)
		if ns, ok := vars["ns"]; ok {
			or = mgr.Orchestrator(ns)
		}
	default:
		return nil, nil
}

shorsher (approved):
Looks good to me, with a very minor style comment

🚀

peterbroadhurst (approved):
👍

PR #879, opened 2022-06-24
surface URI as parameter for token mint

shorsher:
These changes also include added an optional `config` parameter, that is passed to directly to the token connectors without validation.


peterbroadhurst (changes requested):
Really like the approach that for `uri` we use the fully API modeled field to allow it to be set on `mint`.

I note that for other actions (pool creation, approval) we have an opaque `config` that is passed through FireFly to the Token connector, allowing extensibility without needing to change Core.

Feels like to avoid needing future changes in Core for other things that might need to be customized on mint/burn/transfer, we should add a similar pass-through field now?

Would you mind adding that into this PR before we merge?

peterbroadhurst (approved):
Thanks for adding that @shorsher

PR #880, opened 2022-06-24
Collapse batchpin.Submitter into multiparty.Manager

awrichar:


shorsher (approved):
Looks good to me.

We spoke offline about moving the `SubmitNetworkAction` login from `orchestrator` into `multiparty`, but I'll leave that with you.

awrichar:
Thanks - after a quick look, moving `SubmitNetworkAction` down will introduce an import cycle between multiparty and identity... so while I think it's correct to do, I'm going to come back to it later.

PR #883, opened 2022-06-29
Track blockchain callback handlers per namespace

awrichar:

For blockchain batch pin events, direct them to an appropriate namespace:
* If the multiparty contract is V1, the namespace is passed explicitly
* If the multiparty contract is V2+, the namespace is packed into the subscription name (and is not sent in the event)

For blockchain network actions, direct them to an appropriate namespace:
* If the multiparty contract is V1, send them to all handlers (the handler will filter based on whether the contract address matches the active contract)
* If the multiparty contract is V2+, the namespace is packed into the subscription name

Caveat for pre-existing subscriptions: Old subscription names will _not_ have a namespace included, so multiple namespaces will actually be sharing a single subscription. This is OK for V1 networks, as batch pin events always contain a namespace and network actions are sent to all handlers. A V2 network cannot function in this situation, so V2 contracts will be fundamentally incompatible with releases of FireFly prior to 1.1.

Tokens are also subject to the same problem where pre-existing subscriptions will not include namespace information, so they always send events to all handlers if namespace is unset.

Unanswered question: how does ff_system "go away" cleanly when migrating to a V2 contract? There is no entry for ff_system in the namespace config, so migration is weird and probably tied to the default namespace instead. UPDATE - see #884


internal/blockchain/ethereum/ethereum.go
```diff
@@ -68,39 +68,56 @@ type Ethereum struct {
...
 
-func (cb *callbacks) BlockchainNetworkAction(action string, event *blockchain.Event, signingKey *core.VerifierRef) error {
-	for _, cb := range cb.handlers {
-		if err := cb.BlockchainNetworkAction(action, event, signingKey); err != nil {
-			return err
+func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, action string, location *fftypes.JSONAny, event *blockchain.Event, signingKey *core.VerifierRef) error {
+	if namespace == "" {
+		// V1 networks don't populate namespace, so deliver the event to every handler
+		for _, handler := range cb.handlers {
+			if err := handler.BlockchainNetworkAction(action, location, event, signingKey); err != nil {
```
peterbroadhurst, line 104: I feel like there should be a check for delivery of this to a V2 handler, and immediate rejection (worried about V1 pre-migrated contracts leaking actions to an unrelated V2 namespace that's new/migrated and happens to share a remote-name like `default`). Will look out for this later in the review

internal/blockchain/ethereum/ethereum.go
```diff
@@ -68,39 +68,56 @@ type Ethereum struct {
...
 		}
+		log.L(ctx).Errorf("No handler found for blockchain network action on namespace '%s'", namespace)
 	}
 	return nil
 }
 
 func (cb *callbacks) BlockchainEvent(event *blockchain.EventWithSubscription) error {
 	for _, cb := range cb.handlers {
+		// Send the event to all handlers and let them match it to a contract listener
+		// TODO: can we push more listener/namespace knowledge down to this layer?
```
peterbroadhurst, line 120: Seems like the same namespace-encoded-in-sub-name trick should work here, right? (with a migration consideration I guess)

internal/blockchain/fabric/fabric.go
```diff
@@ -59,39 +59,57 @@ type Fabric struct {
...
+	log.L(ctx).Errorf("No handler found for blockchain batch pin on namespace '%s'", batch.Namespace)
 	return nil
 }
 
-func (cb *callbacks) BlockchainNetworkAction(action string, event *blockchain.Event, signingKey *core.VerifierRef) error {
-	for _, cb := range cb.handlers {
-		if err := cb.BlockchainNetworkAction(action, event, signingKey); err != nil {
-			return err
+func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, action string, location *fftypes.JSONAny, event *blockchain.Event, signingKey *core.VerifierRef) error {
+	if namespace == "" {
```
peterbroadhurst, line 93: Can't help but (re) note the code duplication here - fine by me if your assessment is it's still the right thing.

internal/blockchain/ethereum/ethereum.go
peterbroadhurst: Ok, found it, in the caller of this function:
https://github.com/hyperledger/firefly/blob/5d31178c72cd0749df64b7639f40889fdadf7501/internal/blockchain/ethereum/ethereum.go#L404-L409

internal/blockchain/ethereum/ethereum.go
peterbroadhurst: (not suggesting the TODO needs to be addressed in this PR)

peterbroadhurst (approved):
Looks great - none of my comments explicitly need action, but won't merge so you can review them first.

internal/blockchain/ethereum/ethereum.go
awrichar: Yes, but you make a good point that if we _can't_ infer the namespace from the subscription name, then V2 handlers should reject it outright. I think there should be a sanity check for that when we initialize the subscription info, to ensure that whenever we get to the point you noted above, `subInfo.namespace` is guaranteed to be valid for V2+ contracts.

awrichar: OK I pushed another commit to include this sanity check during init.

internal/blockchain/fabric/fabric.go
awrichar: Yea, we were able to commonize a chunk of the init and ingress logic for the FireFly multiparty contract when Multiparty Manager came into being - but you're right that the other end (handling the events that pop out of the multiparty contract) still has a lot of duplication. I'll note it for another pass on whether more of it can be commonized into Event Manger or Multiparty Manager (or some new place).

PR #884, opened 2022-07-06
Stop orchestrator for ff_system when moving to network V2

awrichar:
~~In a chain with #883~~
~~In a chain with #890~~

Background: The legacy `ff_system` namespace is created (ie an orchestrator and managers are initialized) whenever the default namespace is a V1 multiparty network. The `ff_system` namespace is initialized with an exact duplicate of the default namespace config (same plugins, same contract addresses, etc).

With this PR, if the default namespace (and therefore `ff_system`) migrates to V2 rules, the `ff_system` orchestrator will be stopped and removed, along with all its children.

This should allow legacy networks to cleanly migrate to V2 rules and retire `ff_system`. Notable caveat: once the default namespace is on V2, no other V1 namespaces will be supported.


internal/blockchain/fabric/fabric.go
```diff
@@ -217,6 +221,10 @@ func (f *Fabric) Init(ctx context.Context, config config.Section, metrics metric
 	f.capabilities = &blockchain.Capabilities{}
 	f.callbacks.handlers = make(map[string]blockchain.Callbacks)
 
+	// Hard-coded cache values for now, as we're only likely to query a given contract a few times in a short span anyway
+	f.versionCache = ccache.New(ccache.Configure().MaxSize(100))
+	f.versionCacheTTL = time.Minute * 5
```
awrichar, line 226: Some context - Multiparty Manager calls `GetNetworkVersion` and then `AddFireflySubscription` back-to-back. It so happens that `AddFireflySubscription` also calls `GetNetworkVersion`. It felt inefficient to query the blockchain twice, so this theoretically avoids _that_ problem, but maybe it's overkill...

internal/blockchain/ethereum/ethereum.go
```diff
@@ -209,6 +213,10 @@ func (e *Ethereum) Init(ctx context.Context, config config.Section, metrics metr
 	e.capabilities = &blockchain.Capabilities{}
 	e.callbacks.handlers = make(map[string]blockchain.Callbacks)
 
+	// Hard-coded cache values for now, as we're only likely to query a given contract a few times in a short span anyway
+	e.versionCache = ccache.New(ccache.Configure().MaxSize(100))
+	e.versionCacheTTL = time.Minute * 5
```
awrichar, line 218: Some context - Multiparty Manager calls `GetNetworkVersion` and then `AddFireflySubscription` back-to-back. It so happens that `AddFireflySubscription` also calls `GetNetworkVersion`. It felt inefficient to query the blockchain twice, so this theoretically avoids _that_ problem, but maybe it's overkill...

internal/blockchain/fabric/fabric.go
```diff
@@ -217,6 +221,10 @@ func (f *Fabric) Init(ctx context.Context, config config.Section, metrics metric
 	f.capabilities = &blockchain.Capabilities{}
 	f.callbacks.handlers = make(map[string]blockchain.Callbacks)
 
+	// Hard-coded cache values for now, as we're only likely to query a given contract a few times in a short span anyway
+	f.versionCache = ccache.New(ccache.Configure().MaxSize(100))
```
shorsher, line 225: same comment as above

awrichar:
I think this should go in after #891 in order to leverage the caching logic there. Maybe after #890 as well. Moving to draft until some of these others clear out.

awrichar:
Alright - this is rebased and ready now.

PR #887, opened 2022-07-06
Add E2E tests for gateway mode

nguyer:


.github/workflows/go.yml
```diff
@@ -29,23 +29,46 @@ jobs:
...
-          - blockchain-provider: fabric
-            test-suite: TestEthereumE2ESuite
-          - blockchain-provider: fabric
-            token-provider: erc1155
-          - blockchain-provider: fabric
-            token-provider: erc20_erc721
-          - blockchain-provider: geth
-            token-provider: none
-      fail-fast: false
+        exclude: [ blockchain-provider: geth, blockchain-provider: fabric ]
```
awrichar, line 35: Confused about why 2 blockchain providers are defined and then both excluded... is this necessary for the "whitelist" below to work?

manifest.json
```diff
@@ -44,6 +44,6 @@
     "release": "v1.1.0-alpha.1"
   },
   "cli": {
-    "tag": "v1.0.2"
+    "tag": "namespace-isolation"
```
awrichar, line 47: May want to merge https://github.com/hyperledger/firefly-cli/pull/195 first and tag a release

.github/workflows/go.yml
nguyer: Yeah this was a bit strange. Basically I wanted a list of all the permutations spelled out exactly (fewer lines of code). But you can't just give `include` by itself without first having a matrix to start from. This was basically a hack to create an empty matrix.

manifest.json
nguyer: Yep. Good idea.

awrichar: https://github.com/hyperledger/firefly-cli/releases/tag/v1.1.0-alpha.1

.github/workflows/go.yml
awrichar: OK, will leave it since it seems to work. Perhaps we'll find a cleaner way some other day 😁

awrichar (approved):

PR #890, opened 2022-07-07
Change default contracts to network version 2

awrichar:

Split out separate "pinBatch" and "networkAction" methods in the V2 contract, and make it the default. Adjust the blockchain plugins to handle both versions of the contract cleanly.

Also fix some bugs found along the way with network actions being broken.


shorsher (approved):
👋🏻

awrichar:
Based on offline discussion, might be worth also splitting the method into separate calls for batch pin / network action (but still both emitting a single shared event). Moving to draft until I can add that change.

shorsher (approved):

jimthematrix (approved):
solidity and chaincode updates look good to me

PR #891, opened 2022-07-07
Custom contract subscriptions now utilize namespaces

shorsher:
Now, both ethconnect and fabconnect subscription names contain the namespace name. This means that events will now be delivered only to the callback handler for that namespace, instead of every single callback handler. Older subscriptions will still be delivered to all handlers.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -280,6 +280,13 @@ nav_order: 2
 |size|The maximum number of messages that can be packed into a batch|`int`|`<nil>`
 |timeout|The timeout to wait for a batch to fill, before sending|[`time.Duration`](https://pkg.go.dev/time#Duration)|`<nil>`
 
+## cache.blockchain
+
+|Key|Description|Type|Default Value|
+|---|-----------|----|-------------|
+|size|test|`string`|`<nil>`
+|ttl|test|`string`|`<nil>`
+
 ## cors
 
 |Key|Description|Type|Default Value|
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -284,8 +284,8 @@ nav_order: 2
 
 |Key|Description|Type|Default Value|
 |---|-----------|----|-------------|
-|size|test|`string`|`<nil>`
-|ttl|test|`string`|`<nil>`
+|size|Size of blockchain cache|`string`|`<nil>`
+|ttl|Time to live for blockchain cache items|`string`|`<nil>`
 
 ## cors
 
--- a/internal/blockchain/ethereum/ethereum.go
+++ b/internal/blockchain/ethereum/ethereum.go
@@ -106,8 +106,9 @@ func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, act
 	return nil
 }
 
-func (cb *callbacks) BlockchainEvent(ctx context.Context, event *blockchain.EventWithSubscription, namespace string) error {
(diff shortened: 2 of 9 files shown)


docs/reference/config.md
```diff
@@ -280,6 +280,13 @@ nav_order: 2
 |size|The maximum number of messages that can be packed into a batch|`int`|`<nil>`
 |timeout|The timeout to wait for a batch to fill, before sending|[`time.Duration`](https://pkg.go.dev/time#Duration)|`<nil>`
 
+## cache.blockchain
+
+|Key|Description|Type|Default Value|
+|---|-----------|----|-------------|
+|size|test|`string`|`<nil>`
```
awrichar, line 287: Need descriptions

internal/blockchain/ethereum/ethereum.go
```diff
@@ -106,13 +106,19 @@ func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, act
...
-func (cb *callbacks) BlockchainEvent(event *blockchain.EventWithSubscription) error {
-	for _, cb := range cb.handlers {
-		// Send the event to all handlers and let them match it to a contract listener
-		// TODO: can we push more listener/namespace knowledge down to this layer?
-		if err := cb.BlockchainEvent(event); err != nil {
-			return err
+func (cb *callbacks) BlockchainEvent(ctx context.Context, event *blockchain.EventWithSubscription, namespace string) error {
+	if namespace == "" {
+		for _, cb := range cb.handlers {
+			// Send the event to all handlers and let them match it to a contract listener
```
awrichar, line 112: Worth noting (as we do in fftokens) that older/pre-existing subscriptions won't have populated namespace

internal/blockchain/ethereum/ethereum.go
```diff
@@ -106,13 +106,19 @@ func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, act
 	return nil
 }
 
-func (cb *callbacks) BlockchainEvent(event *blockchain.EventWithSubscription) error {
-	for _, cb := range cb.handlers {
-		// Send the event to all handlers and let them match it to a contract listener
-		// TODO: can we push more listener/namespace knowledge down to this layer?
-		if err := cb.BlockchainEvent(event); err != nil {
-			return err
+func (cb *callbacks) BlockchainEvent(ctx context.Context, event *blockchain.EventWithSubscription, namespace string) error {
```
awrichar, line 109: nit: can we order `namespace` right after `ctx` like on the other methods above?

internal/blockchain/fabric/fabric.go
```diff
@@ -106,13 +106,19 @@ func (cb *callbacks) BlockchainNetworkAction(ctx context.Context, namespace, act
...
-func (cb *callbacks) BlockchainEvent(event *blockchain.EventWithSubscription) error {
-	for _, cb := range cb.handlers {
-		// Send the event to all handlers and let them match it to a contract listener
-		// TODO: can we push more listener/namespace knowledge down to this layer?
-		if err := cb.BlockchainEvent(event); err != nil {
-			return err
+func (cb *callbacks) BlockchainEvent(ctx context.Context, event *blockchain.EventWithSubscription, namespace string) error {
+	if namespace == "" {
+		for _, cb := range cb.handlers {
+			// Send the event to all handlers and let them match it to a contract listener
```
awrichar, line 112: Same note here as ethereum

PR #892, opened 2022-07-08
Fix logs from E2E test invocation

awrichar:
Disable parallelization and reduce test timeout to 1m.

This re-enables streaming logs during E2E runs (which is disabled by default when running suites in parallel). No real impact since our test jobs are already divided such that they run 1 suite per job.


awrichar:
Wow... is 1m really not long enough for some of these things to confirm?

nguyer (approved):

PR #895, opened 2022-07-08
Map namespace between local name and remote name

awrichar:

Orchestrator is now aware of the namespace local name _and_ remote name, as are some of the managers. Plugins are intentionally not aware of the mapping, but are passed the local or remote namespace names as appropriate.

The only two items which _record_ a remote namespace name are `Message` and `Group`. These now have an additional `localNamespace` field, which will be populated with the local namespace, while the existing `namespace` field will be populated with the remote namespace (because it is part of the hash in both of these cases, it must remain the same for all members). In addition, `Batch` and `Data` items will be populated with the remote namespace name for network transit (but will be stored with the local namespace name only). Beyond these, every other existing object simply continues to populate its `namespace` with a local namespace name.

Some more notes on how the local namespace is filled for other items:
* All definition types (datatypes, contract FFIs/APIs, identities, and token pools) are transmitted _without_ a namespace filled. Each receiving node will map the definition based on the enclosing message's namespace, and then will fill in the local namespace name before storing the item.
* Token transfers and approvals continue to inherit their namespace from the associated token pool.
* Contract listeners are still local-only, and blockchain events tied to them will inherit their namespace.


internal/definitions/sender_contracts.go
```diff
@@ -34,14 +33,17 @@ func (bm *definitionSender) DefineFFI(ctx context.Context, ffi *fftypes.FFI, wai
 	}
 
 	if bm.multiparty {
+		ffi.Namespace = bm.namespace // TODO: change validation to not check namespace
```
shorsher, line 36: Is this still a `TODO`?

PR #896, opened 2022-07-11
Support tokens only namespaces

shorsher:
Now, multiparty-disabled namespaces can run without a blockchain plugin:
 * All contract routes will be disabled
 * Identity normalizations that rely on a blockchain plugin will either default to the specified input/default key or fail
 * `aggregator` is not started within `EventManager`

### DIFF
--- a/internal/coremsgs/en_error_messages.go
+++ b/internal/coremsgs/en_error_messages.go
@@ -256,4 +256,5 @@ var (
 	MsgActionNotSupported                 = ffe("FF10414", "This action is not supported in this namespace", 400)
 	MsgMessagesNotSupported               = ffe("FF10415", "Messages are not supported in this namespace", 400)
 	MsgInvalidSubscriptionForNetwork      = ffe("FF10416", "Subscription name '%s' is invalid according to multiparty network rules in effect (network version=%d)")
+	MsgBlockchainNotConfigured            = ffe("FF10417", "No blockchain plugin configured", 400)
 )
--- a/internal/coremsgs/en_error_messages.go
+++ b/internal/coremsgs/en_error_messages.go
@@ -256,5 +256,5 @@ var (
 	MsgActionNotSupported                 = ffe("FF10414", "This action is not supported in this namespace", 400)
 	MsgMessagesNotSupported               = ffe("FF10415", "Messages are not supported in this namespace", 400)
 	MsgInvalidSubscriptionForNetwork      = ffe("FF10416", "Subscription name '%s' is invalid according to multiparty network rules in effect (network version=%d)")
-	MsgBlockchainNotConfigured            = ffe("FF10417", "No blockchain plugin configured", 400)
+	MsgBlockchainNotConfigured            = ffe("FF10417", "No blockchain plugin configured")
 )
--- a/internal/definitions/handler.go
+++ b/internal/definitions/handler.go
@@ -80,16 +80,16 @@ type definitionHandler struct {
 	namespace  string
 	multiparty bool
 	database   database.Plugin
-	blockchain blockchain.Plugin
+	blockchain blockchain.Plugin   // optional
 	exchange   dataexchange.Plugin // optional
 	data       data.Manager
 	identity   identity.Manager
 	assets     assets.Manager
-	contracts  contracts.Manager
+	contracts  contracts.Manager // optional
 }
 
 func newDefinitionHandler(ctx context.Context, ns string, multiparty bool, di database.Plugin, bi blockchain.Plugin, dx dataexchange.Plugin, dm data.Manager, im identity.Manager, am assets.Manager, cm contracts.Manager) (*definitionHandler, error) {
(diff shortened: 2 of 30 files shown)


internal/identity/identitymanager.go
```diff
@@ -56,7 +56,7 @@ type Manager interface {
```
shorsher, line 56: I could benefit from some double checking on how I handled the situations with no blockchain plugin. I wasn't sure when to error out or just use the default key/given input.

internal/definitions/handler.go
```diff
@@ -80,16 +80,16 @@ type definitionHandler struct {
...
 	multiparty bool
 	database   database.Plugin
-	blockchain blockchain.Plugin
+	blockchain blockchain.Plugin   // optional
 	exchange   dataexchange.Plugin // optional
 	data       data.Manager
 	identity   identity.Manager
 	assets     assets.Manager
-	contracts  contracts.Manager
+	contracts  contracts.Manager // optional
```
awrichar, line 88: Since two more items became "optional" (and may now be nil), are there any places that need additional nil checks? Or they're on paths that are guaranteed not to be hit?

internal/definitions/sender.go
```diff
@@ -53,7 +53,7 @@ type definitionSender struct {
 	broadcast  broadcast.Manager // optional
 	identity   identity.Manager
 	data       data.Manager
-	contracts  contracts.Manager
+	contracts  contracts.Manager // optional
```
awrichar, line 56: Same question as on handler.go

internal/events/event_manager.go
```diff
@@ -99,7 +99,7 @@ type eventManager struct {
 	data                  data.Manager
 	subManager            *subscriptionManager
 	retry                 retry.Retry
-	aggregator            *aggregator
+	aggregator            *aggregator              // optional
```
awrichar, line 102: blobReceiver is also optional now

internal/coremsgs/en_error_messages.go
```diff
@@ -256,4 +256,5 @@ var (
 	MsgActionNotSupported                 = ffe("FF10414", "This action is not supported in this namespace", 400)
 	MsgMessagesNotSupported               = ffe("FF10415", "Messages are not supported in this namespace", 400)
 	MsgInvalidSubscriptionForNetwork      = ffe("FF10416", "Subscription name '%s' is invalid according to multiparty network rules in effect (network version=%d)")
+	MsgBlockchainNotConfigured            = ffe("FF10417", "No blockchain plugin configured", 400)
```
awrichar, line 259: nit: not sure this is an HTTP 400, as it's not explicitly a problem with the request format (more of a server config error, so probably can omit the error code and keep the default of 500)

internal/identity/identitymanager.go
```diff
@@ -105,6 +105,16 @@ func ParseKeyNormalizationConfig(strConfigVal string) int {
...
+		if inputKey == "" {
+			log.L(ctx).Info("Blockchain plugin missing, falling back to using default key")
+			return im.defaultKey, nil
+		}
+		log.L(ctx).Info("Blockchain plugin missing, falling back to using input key")
+		return inputKey, nil
+	}
+
 	if inputKey == "" {
 		verifierRef, err := im.getDefaultVerifier(ctx)
```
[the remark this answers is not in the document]
shorsher: I solved this a bit differently, because I didn't see the benefit of adding `keyNormalizationMode` to `getDefaultVerifier`. Let me know what you think.

internal/identity/identitymanager.go
```diff
@@ -115,7 +119,7 @@ func (im *identityManager) NormalizeSigningKey(ctx context.Context, inputKey str
 	// If the caller is not confident that the blockchain plugin/connector should be used to resolve,
 	// for example it might be a different blockchain (Eth vs Fabric etc.), or it has its own
 	// verification/management of keys, it should set `assets.keyNormalization: "none"` in the config.
-	if keyNormalizationMode != KeyNormalizationBlockchainPlugin {
+	if keyNormalizationMode != KeyNormalizationBlockchainPlugin || im.blockchain == nil {
```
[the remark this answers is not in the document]
awrichar: I think removing this check would be enough to make that happen... since it would fall through to `normalizeKeyViaBlockchainPlugin` and error from there.

internal/identity/identitymanager.go
```diff
@@ -106,6 +106,10 @@ func ParseKeyNormalizationConfig(strConfigVal string) int {
 // or when the author is known by the caller and should not / cannot be confirmed prior to sending (identity claims)
 func (im *identityManager) NormalizeSigningKey(ctx context.Context, inputKey string, keyNormalizationMode int) (signingKey string, err error) {
 	if inputKey == "" {
+		if im.blockchain == nil {
+			return im.defaultKey, nil
```
awrichar, line 110: Shouldn't we error if `defaultKey` is unset?

PR #899, opened 2022-07-12
Add E2E tests for contract termination

awrichar:

~~Depends on https://github.com/hyperledger/firefly-cli/pull/197~~
~~In a chain with #904~~

Add a new E2E test which dynamically:
* deploys two instances of the FireFly multiparty contract
* dynamically adds a new namespace to the config of both nodes
* resets both nodes to read the config
* sends one message on the first contract
* performs a network migration
* sends one message on the new contract

There are variations of this test for both V1->V2 contract migrations and V2->V2. Notably, the V1->V2 migration can only be performed once per stack, so it is not part of the main suites. **We may need to add a dedicated nightly job for this test.**

This PR also includes loads of fixes for things exposed by the E2E tests, including major changes to how the legacy ff_system namespace is managed. There can now only be one multiparty V1 contract _EVER_ used in the history of a FireFly node, and attempting to configure any namespace with a different V1 contract is an error. The ff_system orchestrator will be initialized with that V1 config (if any is found) and will live forever (walking back some of the changes from #884), because we can never be fully sure when all pins from ff_system have been aggregated.


awrichar:
Factored some of the general enhancements out to https://github.com/hyperledger/firefly/pull/904 as a pre-req before adding the new test here.

awrichar:
Further, honing of this test has revealed that the migration feature is still a bit broken. Moving back to draft state while addressing the remaining issues.

awrichar:
Fixed up the test so that it passes now, so I'm opening this back up for review. I think there may still be work to do to reliably terminate the ff_system namespace though.

shorsher:
Trying to fix the merge conflicts + update tests proved to get a bit nasty. Going to fix the tests in this branch, make the `TODO` changes, and then fix the conflicts.

shorsher:
Currently investigating the contract migration test and why firefly does not seem to pick up new namespaces after a reset.

awrichar:
This is now ready for review again. Concluded that the ff_system namespace actually can never be "stopped". Once it exists, we have to keep the listener around forever.

shorsher (approved):
🚀

PR #900, opened 2022-07-12
Tokens plugins now support a remote name

shorsher:
- introduce a new `remotename` config key for tokens plugins
 - `remotename` must be unique
 - If not provided, `remotename` will default to the plugin `name`
 - `remotename` should never be stored in the local database.

The definitions package now utilizes the `remotename` of tokens plugins. `defsender` will replace the local token plugin name with the remote name before broadcasting the definition. The `defhandler` will then replace the `remotename` with the local name after receiving a broadcast.

### DIFF
--- a/internal/definitions/handler_tokenpool.go
+++ b/internal/definitions/handler_tokenpool.go
@@ -37,7 +37,8 @@ func (dh *definitionHandler) handleTokenPoolBroadcast(ctx context.Context, state
 	if localName, ok := dh.tokenNames[pool.Connector]; ok {
 		pool.Connector = localName
 	} else {
-		log.L(ctx).Infof("Could not find local name for token connector remote name: %s", pool.Connector)
+		log.L(ctx).Infof("Could not find local name for token connector: %s", pool.Connector)
+		return HandlerResult{Action: ActionReject}, i18n.NewError(ctx, coremsgs.MsgInvalidConnectorName, pool.Connector, "token")
 	}
 
 	pool.Message = msg.Header.ID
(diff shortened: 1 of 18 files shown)


internal/identity/identitymanager.go
```diff
@@ -123,7 +123,7 @@ func (im *identityManager) NormalizeSigningKey(ctx context.Context, inputKey str
 	// If the caller is not confident that the blockchain plugin/connector should be used to resolve,
 	// for example it might be a different blockchain (Eth vs Fabric etc.), or it has its own
 	// verification/management of keys, it should set `assets.keyNormalization: "none"` in the config.
-	if keyNormalizationMode != KeyNormalizationBlockchainPlugin {
+	if keyNormalizationMode != KeyNormalizationBlockchainPlugin || im.blockchain == nil {
```
awrichar, line 126: Did this creep in from another PR? I don't think this should be here.

internal/definitions/handler_tokenpool.go
```diff
@@ -33,6 +33,13 @@ func (dh *definitionHandler) handleTokenPoolBroadcast(ctx context.Context, state
 	}
 
 	pool := announce.Pool
+	// Map remote connector name -> local name
+	if localName, ok := dh.tokenNames[pool.Connector]; ok {
+		pool.Connector = localName
+	} else {
+		log.L(ctx).Infof("Could not find local name for token connector remote name: %s", pool.Connector)
```
awrichar, line 40: I think this should reject the message with an error.

shorsher:
Seems I accidentally pulled in some changes from #896, gonna fix them now.

awrichar (approved):
Looks good. We should probably add an E2E test.

shorsher:
I believe an e2e test would require some CLI changes to start a stack without a blockchain plugin. Holding off on merging this.

PR #904, opened 2022-07-14
E2E suite cleanup

awrichar:
Factor out a struct for FireFlyClient (bringing it closer to being exportable as an actual client SDK).

Reorganize and clean up various pieces between `run.sh` and the suite itself (preferring to keep `run.sh` as thin as possible).


pkg/blockchain/plugin.go
```diff
@@ -36,7 +36,7 @@ type Plugin interface {
 	Init(ctx context.Context, config config.Section, metrics metrics.Manager) error
 
 	// SetHandler registers a handler to receive callbacks
-	// If namespace is set, plugin will attempt to deliver only events for that namespace
+	// Plugin will attempt (but is not guaranteed) to deliver events only for the given namespace
```
nguyer, line 39: Does this mean that it might accidentally deliver events for a different namespace?

nguyer (approved):

pkg/blockchain/plugin.go
shorsher: In `1.1`, ethconnect/fabconnect subscriptions will be in the format `ff-sub-<namespace>-<listener ID>`. The plugins will attempt to parse the `namespace` from the subscription name and then deliver to the correct handler. If not, it will deliver to all plugins.

One scenario where this is likely is existing subscriptions after upgrading from `1.0` to `1.1`. 

You can see https://github.com/hyperledger/firefly/pull/891 if you're interested.

awrichar: @nguyer In short, yes. The managers that register callbacks must therefore always check each event and discard any that are not relevant to them.

Originally there was no filtering at the plugin level (each plugin would deliver all events it received to every subscribed manager). We added namespace-level filtering as a best effort to make things more performant, but as @shorsher mentioned there are situations where it just isn't tenable at the plugin level.

PR #909, opened 2022-07-26
Build Docker once to share with all E2E jobs

awrichar:
Each E2E job in the workflow matrix currently takes ~10 min, where the first ~5 min is just spent building the FireFly Docker image.

By building the Docker once upfront and sharing it across all jobs, each test job can be cut in half. This shaves 20-30 min off the combined runtime of the workflow jobs and makes more efficient use of the available workers.

As a small side benefit, the Docker image for each run is also stored as an artifact available for download.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -26,8 +26,33 @@ jobs:
       - name: Upload coverage
         run: bash <(curl -s https://codecov.io/bash)
 
+  docker:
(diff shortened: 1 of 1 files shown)


shorsher (approved):
🐳

shorsher:
should we also pull in https://github.com/hyperledger/firefly/pull/873?

PR #911, opened 2022-07-27
Properly handle remote->local namespace mapping in blockchain plugin

awrichar:
In the "new world" (network version 2), the blockchain plugin (like other plugins) should only deal with the local namespace name.

For network version 1, the (remote) namespace must be written on chain, so it's necessary to map back and forth from local to remote namespace name within the plugin. This does break the paradigm of keeping namespace mapping at a higher level only, but seems like the cleanest solution for old environments.

Note that this also alters the subscription creation logic for V1 vs. V2. For V1, there will be a shared subscription for all namespaces on a given contract (which is how FireFly 1.0 already behaves - it was changed in #865, but it's actually easiest to keep the old behavior for V1 contracts). For V2+ only, there will be a separate subscription created per namespace.

Includes a basic E2E test for namespace mapping.


internal/operations/manager.go
```diff
@@ -165,37 +163,6 @@ func (om *operationsManager) RetryOperation(ctx context.Context, opID *fftypes.U
 	return op, err
 }
 
-func (om *operationsManager) TransferResult(dx dataexchange.Plugin, event dataexchange.DXEvent) {
```
shorsher, line 168: What was this used for?

shorsher (approved):
❤️ PR's with more red. This looks really good.

PR #915, opened 2022-07-30
Allow multiple copies of things on different namespaces

awrichar:
Everything that can be sent (messages, data, all definition types) needs to only be unique within a namespace. The IDs can no longer be considered globally unique, and the database needs to reflect that.

Move "node" configuration to the "multiparty" section (next to "org"), as different namespaces may want to advertise different names for their local node.

Clean up operation handling in DX to make it consistent with other plugins.


internal/events/dx_callbacks.go
```diff
@@ -175,15 +175,15 @@ func (em *eventManager) markUnpinnedMessagesConfirmed(ctx context.Context, batch
...
+func (em *eventManager) DXEvent(ctx context.Context, dx dataexchange.Plugin, event dataexchange.DXEvent) {
 	switch event.Type() {
 	case dataexchange.DXEventTypePrivateBlobReceived:
 		em.privateBlobReceived(dx, event)
 	case dataexchange.DXEventTypeMessageReceived:
 		// Batches are significant items of work in their own right, so get dispatched to their own routines
 		go em.messageReceived(dx, event)
 	default:
-		log.L(em.ctx).Errorf("Invalid DX event type from %s: %d", dx.Name(), event.Type())
+		log.L(em.ctx).Errorf("Invalid data exchange event type from %s: %d", dx.Name(), event.Type())
```
shorsher, line 186: if we're now passing in a context, should we use that one for logging?

internal/events/dx_callbacks.go
```diff
@@ -208,7 +208,7 @@ func (em *eventManager) privateBlobReceived(dx dataexchange.Plugin, event dataex
 	log.L(em.ctx).Infof("Blob received event from data exchange %s: Peer='%s' Hash='%v' PayloadRef='%s'", dx.Name(), br.PeerID, &br.Hash, br.PayloadRef)
 
 	if br.PeerID == "" || len(br.PeerID) > 256 || br.PayloadRef == "" || len(br.PayloadRef) > 1024 {
-		log.L(em.ctx).Errorf("Invalid blob received event from data exhange: Peer='%s' Hash='%v' PayloadRef='%s'", br.PeerID, &br.Hash, br.PayloadRef)
+		log.L(em.ctx).Errorf("Invalid blob received event from data exchange: Peer='%s' Hash='%v' PayloadRef='%s'", br.PeerID, &br.Hash, br.PayloadRef)
```
shorsher, line 211: same comment as above

shorsher (approved):

internal/events/dx_callbacks.go
awrichar: I started to do this, but it proved to be a fairly large change to the way the event handling (and particularly the unit tests) are laid out. I altered the commit to not pass the context, since there's no compelling reason we need to use it right now vs. the context on event manager.

PR #916, opened 2022-08-01
Add node-specific suffix to DX endpoint "id"

awrichar:
~~In a chain with #915~~
Depends on https://github.com/hyperledger/firefly-dataexchange-https/pull/62

UPDATE: found a number of bugs while adding E2E tests - so #918 contains everything from this PR, some additional fixes, and an E2E test. The comments below are still relevant though.

FireFly can now support many local namespaces that map to a single remote namespace (for multi-tenancy purposes). Each of these local namespaces may have their own "node" within FireFly, but they may all be sharing a single DX plugin. This change adds a "nodeName" qualifier to many of the dataexchange plugin calls, along with a specific implementation in the FFDX plugin code, to allow many nodes to exist behind a single DX.

Implementation outline:
* For newly created nodes, the FFDX connector will append a suffix to the member ID advertised by the plugin before storing or broadcasting it. This means each node will have a unique verifier ID even if they are sharing a DX member ID and certificate. The FFDX connector and FFDX plugin have agreed on `/` as a separator.
* The FFDX connector will ignore this suffix when routing messages (paying attention only to the member ID before the separator) - but will pass the full recipient string for further routing by FireFly on the far end.
* FireFly will use the combination of the received message's remote namespace and recipient ID to route it uniquely to a single event manager.

Should remain fully backwards compatible for nodes that were already broadcast/stored in the database with a non-suffixed peer ID.

### DIFF
--- a/internal/blockchain/ethereum/ethereum.go
+++ b/internal/blockchain/ethereum/ethereum.go
@@ -201,13 +201,6 @@ func (e *Ethereum) AddFireflySubscription(ctx context.Context, namespace core.Na
 		return "", err
 	}
 
-	switch firstEvent {
-	case string(core.SubOptsFirstEventOldest):
-		firstEvent = "0"
-	case string(core.SubOptsFirstEventNewest):
-		firstEvent = "latest"
-	}
-
 	version, err := e.GetNetworkVersion(ctx, location)
 	if err != nil {
 		return "", err
--- a/internal/dataexchange/ffdx/dxevent.go
+++ b/internal/dataexchange/ffdx/dxevent.go
@@ -86,6 +86,7 @@ func (h *FFDX) dispatchEvent(msg *wsEvent) {
 	var namespace string
 	var err error
 	e := &dxEvent{ffdx: h, id: msg.EventID}
+
 	switch msg.Type {
 	case messageFailed:
 		h.callbacks.OperationUpdate(h.ctx, &core.OperationUpdate{
@@ -118,23 +119,6 @@ func (h *FFDX) dispatchEvent(msg *wsEvent) {
 			OnComplete:     e.Ack,
 		})
 		return
-	case messageReceived:
-		// De-serialize the transport wrapper
-		var wrapper *core.TransportWrapper
-		err = json.Unmarshal([]byte(msg.Message), &wrapper)
-		switch {
-		case err != nil:
-			err = fmt.Errorf("invalid transmission from peer '%s': %s", msg.Sender, err)
-		case wrapper.Batch == nil:
-			err = fmt.Errorf("invalid transmission from peer '%s': nil batch", msg.Sender)
-		default:
-			namespace = wrapper.Batch.Namespace
-			e.dxType = dataexchange.DXEventTypeMessageReceived
-			e.messageReceived = &dataexchange.MessageReceived{
-				PeerID:    msg.Sender,
-				Transport: wrapper,
-			}
-		}
 	case blobFailed:
 		h.callbacks.OperationUpdate(h.ctx, &core.OperationUpdate{
 			NamespacedOpID: msg.RequestID,
@@ -156,6 +140,35 @@ func (h *FFDX) dispatchEvent(msg *wsEvent) {
 			OnComplete:     e.Ack,
 		})
 		return
+	case blobAcknowledged:
+		h.callbacks.OperationUpdate(h.ctx, &core.OperationUpdate{
+			NamespacedOpID: msg.RequestID,
+			Status:         core.OpStatusSucceeded,
+			Output:         msg.Info,
+			VerifyManifest: h.capabilities.Manifest,
+			DXHash:         msg.Hash,
+			OnComplete:     e.Ack,
+		})
+		return
+
+	case messageReceived:
+		// De-serialize the transport wrapper
+		var wrapper *core.TransportWrapper
+		err = json.Unmarshal([]byte(msg.Message), &wrapper)
+		switch {
+		case err != nil:
+			err = fmt.Errorf("invalid transmission from peer '%s': %s", msg.Sender, err)
+		case wrapper.Batch == nil:
+			err = fmt.Errorf("invalid transmission from peer '%s': nil batch", msg.Sender)
+		default:
+			namespace = wrapper.Batch.Namespace
+			e.dxType = dataexchange.DXEventTypeMessageReceived
+			e.messageReceived = &dataexchange.MessageReceived{
+				PeerID:    msg.Sender,
+				Transport: wrapper,
+			}
+		}
+
 	case blobReceived:
 		var hash *fftypes.Bytes32
 		hash, err = fftypes.ParseBytes32(h.ctx, msg.Hash)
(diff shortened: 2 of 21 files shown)


internal/dataexchange/ffdx/ffdx.go
```diff
@@ -67,11 +74,17 @@ func (cb *callbacks) OperationUpdate(ctx context.Context, update *core.Operation
 	}
 }
 
-func (cb *callbacks) DXEvent(ctx context.Context, namespace string, event dataexchange.DXEvent) {
-	if handler, ok := cb.handlers[namespace]; ok {
-		handler.DXEvent(cb.plugin, event)
+func (cb *callbacks) DXEvent(ctx context.Context, namespace, recipient string, event dataexchange.DXEvent) {
+	if node, ok := cb.plugin.nodes[recipient]; ok {
```
awrichar, line 78: Is it possible for a DX message to arrive _before_ the node identity has been confirmed? Here `plugin.nodes` is populated with any nodes that exist on startup or any nodes that are confirmed by an identity claim broadcast. I don't see any way to pick the correct handler unless the node has been claimed and confirmed.

internal/dataexchange/ffdx/dxevent.go
```diff
@@ -170,25 +183,16 @@ func (h *FFDX) dispatchEvent(msg *wsEvent) {
...
 
 	// If we couldn't dispatch the event we received, we still ack it
-	if err == nil {
-		h.callbacks.DXEvent(h.ctx, namespace, e)
-	} else {
+	if err != nil {
 		log.L(h.ctx).Warnf("Failed to dispatch DX event: %s", err)
 		e.Ack()
+	} else {
+		h.callbacks.DXEvent(h.ctx, namespace, msg.Recipient, e)
```
awrichar, line 196: Can we always assume `Recipient` will be set now, or do we need a fallback if it's not?

gabriel-indik: Yes, there recipient should always be present in the blob received events.

awrichar: Confirmed with @gabriel-indik - yes it's always set.

internal/dataexchange/ffdx/ffdx.go
awrichar: Basically we have a validation on the new "recipient" field provided by DX, to see that it matches a known node/verifier in the database. Previously this field didn't exist, so any message received was just put into the database.

pkg/dataexchange/plugin.go
```diff
@@ -97,7 +97,7 @@ type Plugin interface {
 
 	// SendMessage sends an in-line package of data to another network node.
 	// Should return as quickly as possible for parallelism, then report completion asynchronously via the operation ID
-	SendMessage(ctx context.Context, nsOpID, peerID string, data []byte) (err error)
+	SendMessage(ctx context.Context, nsOpID, peerID string, senderID string, data []byte) (err error)
 
 	// TransferBlob initiates a transfer of a previously stored blob to another node
 	TransferBlob(ctx context.Context, nsOpID, peerID string, payloadRef string) (err error)
```
awrichar, line 103: Should probably add a sender ID to this call too...

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -47,12 +49,17 @@ type FFDX struct {
 	needsInit    bool
 	initialized  bool
 	initMutex    sync.Mutex
-	nodes        []fftypes.JSONObject
+	nodes        map[string]*dxNode
```
awrichar, line 52: Does this need a mutex?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -201,13 +201,6 @@ func (e *Ethereum) AddFireflySubscription(ctx context.Context, namespace core.Na
 		return "", err
 	}
 
-	switch firstEvent {
```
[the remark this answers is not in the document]
peterbroadhurst: ... I didn't (or I just missed it) - mind adding an explanation of this one?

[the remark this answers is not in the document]
awrichar: Looks like this is just a general cleanup item that crept into this branch. Nothing to do with the PR really, but this same check is performed again in the `createSubscription()` method ultimately called from this flow. I can keep it here or pull it out to a separate PR if it seems extraneous.

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -183,12 +203,17 @@ func (h *FFDX) Init(ctx context.Context, config config.Section) (err error) {
 	return nil
 }
 
-func (h *FFDX) SetNodes(nodes []fftypes.JSONObject) {
-	h.nodes = nodes
+func (h *FFDX) InitPeer(nodeName string, peer fftypes.JSONObject) {
+	id := peer.GetString("id")
```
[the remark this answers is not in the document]
awrichar: Technically there shouldn't be any contention before `Start()` is called. Everything in the "init" phase happens in serial. Open to whether that can be made more clear or if the mutex should be here too just for consistency.

internal/dataexchange/ffdx/ffdx.go
peterbroadhurst: So the question is if you were to:
1. As `B` broadcast my node to the network
2. As `A` triggers an event off of the broadcast of `B` and _immediately_ send a message to a group containing `A`
3. As `A` receive the message from `B` _before_ processing the broadcast from (1) - even though it was successfully processed by the remote node.

I believe it is (at least theoretically) possible to engineer this scenario.
It doesn't seem like one to worry about in practice.

awrichar: Yea, I think it would mainly be if the blockchain listener on `B` was disconnected or behind when (3) is received. It would ignore the off-chain data since there's no orchestrator lined up to handle it until (1) is confirmed (on- and off-chain portions) on my local node. It feels like an edge case though.

Perhaps the more important angle to this is that there's no way to communicate back to `A` whether the message was processed. DX will indicate success as long as the message was delivered to FireFly - but if FireFly decides to drop it on the floor, `A` will never know.

awrichar: In the more general case, you could send a message to `peer0/jack` and DX would report success as long as `peer0` is valid - regardless of whether `/jack` exists at all.

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -210,8 +235,14 @@ func (h *FFDX) beforeConnect(ctx context.Context) error {
 	if h.needsInit {
 		h.initialized = false
 		var status dxStatus
+		var body []fftypes.JSONObject
+		h.nodesMutex.Lock()
+		for _, node := range h.nodes {
```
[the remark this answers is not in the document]
awrichar: OK, I updated to fully separate the node lists per remote namespace - hopefully that makes it a bit more deterministic.

internal/dataexchange/ffdx/ffdx.go
peterbroadhurst: spoke to @gabriel-indik about this, and we agreed the MTLS implementation of DX should reject a HTTP POST to an unregistered node ID

awrichar: OK - can table it in the context of this PR then, and come back after scoping what it would take to push that responsibility into DX

awrichar:
Regarding the changes to the "Skipping send of batch for local node" logic - that's in the next PR (#918). Didn't discover the need for it until I added the E2E test.

internal/dataexchange/ffdx/ffdx.go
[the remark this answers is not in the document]
awrichar: moot point now as I've removed this method

internal/dataexchange/ffdx/ffdx.go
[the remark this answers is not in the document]
peterbroadhurst: Per https://discord.com/channels/905194001349627914/971132225355657267/1004818976876007445 there's going to be follow-on work between @awrichar and @gabriel-indik to work this one through. We are moving forwards with this update to FF Core for now as it's holding up the line, and this is well understood.

peterbroadhurst (approved):
Thanks @awrichar - sorry for taking a while with the review 🙇

PR #917, opened 2022-08-02
Add admin and config fields to E2E testState

awrichar:


shorsher (approved):

awrichar:
Yup, think it makes sense to continue moving them up if they're shared.

PR #920, opened 2022-08-02
add E2E test for token plugins with remote names

### LINKED ISSUE #929: Tokens-only test moves forward too quickly after reset
Failure here: https://github.com/hyperledger/firefly/runs/7675078438?check_suite_focus=true#step:6:923

Because this test polls for the `default` namespace to be ready, it may actually report success immediately (before the reset). It's better to poll for the new namespace, which won't be ready until reset has been successful.

shorsher:
Adds a new test suite for tokens plugins with remote names. 

Fixes #929


test/e2e/multiparty/tokens.go
```diff
@@ -370,3 +370,191 @@ func (suite *TokensTestSuite) TestE2ENonFungibleTokensSync() {
...
+
+func (suite *TokensTestSuite) TestE2EFungibleTokensWithRemoteNameAsync() {
+	defer suite.testState.done()
+	config1 := e2e.ReadConfig(suite.T(), suite.testState.configFile1)
+	e2e.AddPluginRemoteName(config1, "tokens", "testremote")
+	e2e.WriteConfig(suite.T(), suite.testState.configFile1, config1)
+
+	config2 := e2e.ReadConfig(suite.T(), suite.testState.configFile2)
+	e2e.AddPluginRemoteName(config2, "tokens", "testremote")
+	e2e.WriteConfig(suite.T(), suite.testState.configFile2, config2)
```
awrichar, line 382: We probably want the local plugin names to be different on the two nodes (otherwise it might be ignoring the remote name and still just happen to work because the local names match).

This test also might need to be in a suite of its own, since it makes pretty invasive changes to the config.

test/e2e/runners/ethereum_multiparty_test.go
```diff
@@ -27,6 +27,7 @@ func TestEthereumMultipartyE2ESuite(t *testing.T) {
 	suite.Run(t, new(multiparty.IdentityTestSuite))
 	suite.Run(t, new(multiparty.OnChainOffChainTestSuite))
 	suite.Run(t, new(multiparty.TokensTestSuite))
+	suite.Run(t, new(multiparty.TokensRemoteNameTestSuite))
```
awrichar, line 30: When I set "suite" I actually meant "runner". I think this test should be in its own runner since it makes changes to the token plugin config.

awrichar: I suppose "suite" was the first step though.

test/e2e/multiparty/tokens_remote_name.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (suite *TokensRemoteNameTestSuite) TestE2EFungibleTokensWithRemoteNameAsync() {
+	defer suite.testState.done()
+
+	newLocalName := "tokens1"
+
+	config1 := e2e.ReadConfig(suite.T(), suite.testState.configFile1)
+	e2e.AddPluginRemoteName(config1, "tokens", "testremote")
+
+	// Change plugin local name and update plugin list
```
awrichar, line 58: Could make this a helper function

awrichar (approved):
Seems good, but left some comments for consideration.

awrichar:
Since this is a new runner, it may need to be added to the GitHub Workflow test matrixes for PR jobs and integration.

shorsher:
I see the integration tests haven't been updated, so I'll file an issue

PR #923, opened 2022-08-03
Updates for EVMConnect support (with EthConnect cross-compatibility)

peterbroadhurst:
- Support the `TransactionUpdate` type of websocket payload - which just updates the operation, while leaving it in `Pending` status (rather than marking it success/fail)
- Support events with an `0x12345:` address prefix on the signature


nguyer (approved):
Looks good to me. Thanks

PR #926, opened 2022-08-04
Destroy sysmessaging package 💥

awrichar:
It was created to stash random interfaces just to avoid circular dependencies - but all of these interfaces can be relocated now without introducing a circle.


peterbroadhurst (approved):
🧹 🧹

PR #932, opened 2022-08-04
Pin golangci-lint to v1.47.3 instead of latest

awrichar:
Avoids surprising changes to linter rules.

Also looks like the current "latest" version of golangci-lint is Go 1.19 only, and we don't want to upgrade just yet.


shorsher (approved):

peterbroadhurst (approved):
👍

PR #938, opened 2022-08-08
update integration test to use new e2e suites

shorsher:


.github/workflows/integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        test-suite: [TestEthereumE2ESuite, TestFabricE2ESuite]
+        test-suite: [TestEthereumMultipartyE2ESuite, TestEthereumGatewayE2ESuite, TestEthereumMultipartyTokensRemoteNameE2ESuite, TestEthereumNamespaceE2ESuite, TestFabricGatewayE2ESuite, TestFabricMultipartyE2ESuite]
```
shorsher, line 15: Should we include the migration tests?

awrichar: There's already a separate job (line 60)

.github/workflows/integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
...
         blockchain-provider: [geth, besu, fabric]
         database-type: [sqlite3, postgres]
         exclude:
```
awrichar, line 18: The excluded combos need to be adjusted below for the new suite names. And you definitely need to exclude the combos of "fabric" with any test suite that uses tokens.

awrichar (approved):
Not sure of the best way to test this, but it LGTM.

.github/workflows/integration.yml
```diff
@@ -12,16 +12,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        test-suite: [TestEthereumE2ESuite, TestFabricE2ESuite]
+        test-suite: [TestEthereumMultipartyE2ESuite, TestEthereumGatewayE2ESuite, TestEthereumMultipartyTokensRemoteNameE2ESuite, TestEthereumNamespaceE2ESuite, TestFabricGatewayE2ESuite, TestFabricMultipartyE2ESuite]
```
awrichar, line 15: Actually can you pull any reference to `TestEthereumNamespaceE2ESuite`? It's just a subset of the main multiparty suite, and I think it's likely broken at the moment, and also probably not needed going forward due to the way identities work now.

PR #942, opened 2022-08-08
Advertise full node name (with suffix) to DX

awrichar:
Depends on https://github.com/hyperledger/firefly-dataexchange-https/pull/67


awrichar:
This is failing because the DX node doesn't trust its own cert (when sending messages to itself).

peterbroadhurst (approved):
👍

PR #947, opened 2022-08-10
Replace namespace.remotename with namespace.multiparty.networkname

awrichar:
This is a proposed rename for discussion.

Currently a namespace can have a `name` and a `remoteName` - `name` is how it's known locally (on APIs and most database objects), while `remoteName` is how it's written into objects that transmit the namespace in a multi-party system (specifically messages, batches, and groups on dataexchange, and the BatchPin method of the V1 blockchain contract).

Since the thing we've called `remoteName` is entirely specific to multiparty networks, it could be renamed to `multiparty.networkNamespace`, which seems more accurate. It would still default to the value of the local namespace `name` if unset.

Counterpoints:
- These fields will continue to show as `localNamespace` and `namespace` on messages and groups when queried (and the latter can't easily be modified due to the way hashes are computed).
- ~~Tokens plugins have a similar `name` and `remoteName` (one is stored locally and one is sent in messages), so this would be a departure from that terminology.~~ Update: decided to rename the tokens plugin `remoteName` to `broadcastName` based on comments below.


peterbroadhurst:
I like `multiparty.networkNamespace` 👍 

As the field is called `namespace` in all Multi-party data structures, calling it just `multiparty.networkName` seems like it might confuse things.

On the Tokens front, that's an interesting separate conversation. Because it's a name specific to the combination of a Token connector, and how it should be broadcast within a `multiparty` system. I wonder if `broadcastName` (rather than `remoteName`) might make sense there.

awrichar:
Yea, I like that suggestion. Will update the PR.

peterbroadhurst (approved):
I think this is a real positive step forwards in the terminology.

> Tokens plugins have a similar name and remoteName (one is stored locally and one is sent in messages), so this would be a departure from that terminology.

I think we've talked about `broadcastName` being an alternative there - but I'm guessing that's a discussion for another PR

PR #955, opened 2022-08-10
Properly match operation updates to the plugin that generated them

awrichar:


nguyer (approved):
This makes sense to me. Thanks!

PR #956, opened 2022-08-11
Messages that fail validation should be rejected

### LINKED ISSUE #952: Messages stuck in sent state after creating invalid identity
I added a new e2e test which created an invalid custom identity (using an already registered key and with a parent in a different org). The identity did not get created, but I did not get any events or messages that indicated that anything had gone wrong. All messages sent for creating the identity are stuck in `sent` state.

e.g.
Screenshot 2022-08-10 at 17 54 19

Logs:
* [firefly_core_0.log](https://github.com/hyperledger/firefly/files/9302448/firefly_core_0.log)
* [firefly_core_1.log](https://github.com/hyperledger/firefly/files/9302449/firefly_core_1.log)

awrichar:
There was some confusion between "message is invalid" and "message could not yet
be dispatched" in the event aggregator. This attempts to separate those two
situations, as the former should be an immediate rejection.

Fixes #952


internal/events/aggregator.go
```diff
@@ -476,14 +476,10 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 
 	}
 
-	dispatched := false
-	var newState core.MessageState
-	if dataAvailable {
```
awrichar, line 481: I removed this check because there's no way to get this far if data was missing.

internal/events/aggregator_test.go
```diff
@@ -1634,20 +1669,38 @@ func TestDispatchBroadcastQueuesLaterDispatch(t *testing.T) {
 	mim.On("FindIdentityForVerifier", ag.ctx, mock.Anything, mock.Anything).Return(org1, nil)
 
 	mdm := ag.data.(*datamocks.Manager)
-	mdm.On("GetMessageWithDataCached", ag.ctx, msg1.Header.ID, data.CRORequirePublicBlobRefs).Return(msg1, core.DataArray{}, true, nil).Once()
-	mdm.On("GetMessageWithDataCached", ag.ctx, msg2.Header.ID, data.CRORequirePublicBlobRefs).Return(msg2, core.DataArray{}, true, nil).Once()
+	data1 := core.DataArray{}
```
awrichar, line 1672: Honestly the aggregator tests are a bit rusty. When I started debugging, they were following paths that I did not expect based on the test names. I suspect they've never been fully overhauled in the wake of lots of identity changes and lots of batch state changes.

I had to touch a few tests to fix them after my changes, so I also tried to make those more consistent with what I _think_ they're supposed to be testing - but just wanted to note that there are quite a few others that are stale and might really need to do extra steps to validate things that end up in `batchState`.

internal/events/aggregator.go
peterbroadhurst: @awrichar - why do we *not* need to call `state.SetContextBlockedBy(ctx, *unmaskedContext, pin.Sequence)` in the case of `!dataAvailable`

That change wasn't made in this PR, and I'm not paged back in enough to answer this myself. But I read the comments on the `if dispatched {` block below and it strikes me it applies on face value:

```
	// - dispatched=false: we need to prevent dispatch of any subsequent messages on the same topic in the batch
```

awrichar: That question also occurred to me, and I was somewhat hoping you knew the answer 😅 sounds like one or both of us needs to page back in more fully on the difference between waiting, rejecting, and blocking.

awrichar: After looking again, it does feel like we should mark the context blocked when the message/data is unavailable. I rearranged the code blocks so it will fall through to that path now.

peterbroadhurst (approved):
Thanks Andrew for working through this and making the additional change. This all looks correct to me.

PR #960, opened 2022-08-11
Changes to support E2E tests with confirmations, where requests might be slow

peterbroadhurst:
Biggest change is to the HTTP Server to allow requests longer than 15s to complete without the client getting an EOF.

See https://github.com/hyperledger/firefly-common/pull/31

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -729,6 +729,14 @@ github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpO
 github.com/hyperledger/firefly-common v0.1.13/go.mod h1:2NqPi5Ud9H6rSlZXkLbotxW7z4EAD89p3/8oNOpm9Gs=
 github.com/hyperledger/firefly-common v0.1.16 h1:21xidDEKrJhtGdBSRqHN4PfDi7aYxF0HOFuAa04V1AE=
 github.com/hyperledger/firefly-common v0.1.16/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811192657-e63ab6827844 h1:DDWL4Tv2vZbcKgc/1dSjgei0QnWIQnu3ikRqg7tBnkQ=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811192657-e63ab6827844/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811194331-db95d0f1bc31 h1:ijIhHR9uDwGBXOBydaQxAVyyReNmUWcTHW5R6oJYc5U=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811194331-db95d0f1bc31/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811195438-ef727fa10629 h1:id4k0rBHXzcez+1qy+E8xP4J3ZudNtuSF7dJaBhPqMI=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811195438-ef727fa10629/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811200324-b6c4c44d2d0d h1:9s1MI4RIfKgftrGQ6sI2gvQBEQElq2CQOs3uaiB44s8=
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -729,6 +729,16 @@ github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpO
...
+github.com/hyperledger/firefly-common v0.1.19-0.20220811192657-e63ab6827844/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811194331-db95d0f1bc31 h1:ijIhHR9uDwGBXOBydaQxAVyyReNmUWcTHW5R6oJYc5U=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811194331-db95d0f1bc31/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811195438-ef727fa10629 h1:id4k0rBHXzcez+1qy+E8xP4J3ZudNtuSF7dJaBhPqMI=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811195438-ef727fa10629/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811200324-b6c4c44d2d0d h1:9s1MI4RIfKgftrGQ6sI2gvQBEQElq2CQOs3uaiB44s8=
+github.com/hyperledger/firefly-common v0.1.19-0.20220811200324-b6c4c44d2d0d/go.mod h1:MNbaI2spBsdZYOub6Duj9xueE7Qyu9itOmJ4vE8tjYw=
+github.com/hyperledger/firefly-common v0.1.20 h1:0dShkjlIShyBxkXRmu3vLmpEK6xrqmfc8GhF6k0Vgbg=
+github.com/hyperledger/firefly-common v0.1.20/go.mod h1:gMlv4Iy5JjnzXmSEdb+tWVDIc/2GhL9MRcgNX+VmI4M=
```
awrichar, line 741: I don't understand go.sum but why are there so many firefly-common entries??

peterbroadhurst: I was testing along the way - it seems to just add in each one you use, and never remove it

peterbroadhurst: e.g. I was doing `go get github.com/hyperledger/firefly-common@b6c4c44d2d0dc4b1c511aab629facf67113e3708` from branches in common, to test my changes

test/e2e/client/restclient.go
```diff
@@ -100,14 +102,18 @@ func NewFireFly(l Logger, hostname, namespace string) *FireFlyClient {
 func NewResty(l Logger) *resty.Client {
 	client := resty.New()
 	client.OnBeforeRequest(func(c *resty.Client, req *resty.Request) error {
-		l.Logf("==> %s %s %s", req.Method, req.URL, req.QueryParam)
+		var b []byte
+		if _, isReader := req.Body.(io.Reader); !isReader {
+			b, _ = json.Marshal(req.Body)
+		}
+		l.Logf("%s: ==> %s %s %s: %s", fftypes.Now(), req.Method, req.URL, req.QueryParam, string(b))
```
awrichar, line 109: Feels a little odd that we add timestamps inside these logs (shouldn't that be a higher-level job delegated to the logging framework for all logs?)

test/e2e/e2e.go
```diff
@@ -223,7 +223,7 @@ func checkObject(t *testing.T, expected interface{}, actual interface{}) bool {
 }
 
 func VerifyAllOperationsSucceeded(t *testing.T, clients []*client.FireFlyClient, startTime time.Time) {
-	tries := 3
+	tries := 30
 	delay := 2 * time.Second
```
awrichar, line 227: 60 seconds seems quite long

test/e2e/client/restclient.go
peterbroadhurst: Yes, but I couldn't find an option in the Go test framework to make `t.Logf` add timestamps... so this is a workaround that didn't mean changing every log line

go.sum
awrichar: `go mod tidy`?

test/e2e/e2e.go
peterbroadhurst: I'm happy to reduce - I think I need 30seconds with 2 confirmations, obviously more with more confirmations

awrichar (approved):
Left a few questions, but overall it seems OK

test/e2e/e2e.go
awrichar: OK, if we're sure 30 seconds is "happy path" in this world, 60 seems reasonable. It does make the suite take quite a long time then, right?

peterbroadhurst: Decision: Leave at 60 for now

go.sum
peterbroadhurst: Decision: Will do this in next PR, to keep the pipeline moving as this is way into the E2Es

PR #961, opened 2022-08-12
Update readme for local dev set up

Chengxuan:
Fixed an error and added link in the core readme to link to component dev readme

Signed-off-by: Chengxuan Xing <chengxuan.xing@kaleido.io>


docs/contributors/advanced_cli_usage.md
```diff
@@ -112,7 +112,7 @@ docker build -t ghcr.io/hyperledger/firefly .
 ```
 
 ## Edit your `manifest.json` file
```
nguyer, line 114: Editing the manifest file is actually no longer necessary, because the FireFly CLI will default to this same behavior when you pass in a manifest file. I think we can remove the whole `## Edit your manifest.json file` section

nguyer (changes requested):
Thanks for updating this. There are actually a few more changes that we should make to this as well, while we're at it.

Chengxuan:
@nguyer Thank you. I've addressed both of the comments.

nguyer (approved):

PR #964, opened 2022-08-15
Add /api/version route returning same info as CLI, with APIVersion

peterbroadhurst:
This PR proposes a simple way for clients (such as the Sandbox) to know what API version is hosted by the runtime server.
- Adds an `APIVersion` that comes for source code, to the existing version structure returned by the CLI
- The first two digits of the fully qualified semver - the "Minor" version, such as `1.1`
- Enforces that in builds where the `Version` is set, that must line up with the APIVersion
   - This is to avoid us forgetting to update it int he future
- Disables the check for `(devel)` builds

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -319,6 +319,14 @@ func (as *apiServer) handlerFactory() *ffapi.HandlerFactory {
 	}
 }
 
+func (as *apiServer) versionInfo(mgr namespace.Manager) func(res http.ResponseWriter, req *http.Request) (status int, err error) {
+	return func(res http.ResponseWriter, req *http.Request) (status int, err error) {
+		res.Header().Set("Content-Type", "application/json")
+		v := mgr.Version()
+		return 200, json.NewEncoder(res).Encode(&v)
+	}
+}
+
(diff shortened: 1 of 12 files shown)


internal/apiserver/server.go
```diff
@@ -347,6 +355,7 @@ func (as *apiServer) createMuxRouter(ctx context.Context, mgr namespace.Manager)
 
 	r.HandleFunc(`/api/swagger{ext:\.yaml|\.json|}`, hf.APIWrapper(as.swaggerHandler(as.swaggerGenerator(routes, apiBaseURL))))
 	r.HandleFunc(`/api`, hf.APIWrapper(hf.SwaggerUIHandler(publicURL+"/api/swagger.yaml")))
+	r.HandleFunc(`/api/version`, hf.APIWrapper(as.versionInfo(mgr)))
```
awrichar, line 358: I'm debating whether this should be a "normal" route that shows up in `routes.go` and lives under `/api/v1/version`. But maybe it's "global" enough to be outside the versioning of the FireFly API itself.

peterbroadhurst:
Going to close this for now - we can revisit later.

PR #967, opened 2022-08-16
Performance improvements for batch lookups

awrichar:
Ensure batch ID queries match the current indexes (to avoid sequential scan), and improve the caching in aggregator when processing a group of pins.


peterbroadhurst (approved):
Changes looks great 👍

PR #971, opened 2022-08-16
Fix group race condition, optimize transaction cache

awrichar:
Insert local group only after inserting groupinit message.
This ensures no other threads pick up on the group before the
message is sequenced.

Update transaction cache after initial insert.


peterbroadhurst (approved):
Looks great 👍

PR #972, opened 2022-08-17
Add cache for operations

awrichar:


internal/operations/operation_updater.go
```diff
@@ -207,26 +206,20 @@ func (ou *operationUpdater) doBatchUpdate(ctx context.Context, updates []*core.O
...
-			txIDs = append(txIDs, op.Transaction)
-		}
-	}
-	var transactions []*core.Transaction
-	if len(txIDs) > 0 {
-		txFilter := database.TransactionQueryFactory.NewFilter(ctx).In("id", txIDs)
-		transactions, _, err = ou.database.GetTransactions(ctx, ou.manager.namespace, txFilter)
-		if err != nil {
-			return err
+			transaction, err := ou.txHelper.GetTransactionByIDCached(ctx, op.Transaction)
```
peterbroadhurst, line 218: This could also help reduce the load, had forgotten we did this extra indirected lookup, just to be able to attache any associated Blockchain TXID to the FF TX 👍

peterbroadhurst: The line that made me think a bit is below (outside the comment-able section):

```go
		if err := ou.txHelper.AddBlockchainTX(ctx, tx, update.BlockchainTXID); err != nil {
```

The most important thing from a cache freshness perspective, is that this operation ends up as additive for any blockchain transaction ID. Right now we only have zero/1 for all our ops, so very unlikely to be able to test this. But I will try and dig into that function a bit if I don't naturally come to it in this review, to see if there's any cache issue inside it.

internal/operations/operation_updater.go
```diff
@@ -347,3 +340,18 @@ func (ou *operationUpdater) close() {
 		}
 	}
 }
+
+func (ou *operationUpdater) resolveOperation(ctx context.Context, ns string, id *fftypes.UUID, status core.OpStatus, errorMsg *string, output fftypes.JSONObject) (err error) {
```
peterbroadhurst, line 344: Aha, we just moved this one 👍  (to my earlier comment)

internal/operations/operation_updater.go
peterbroadhurst: Ok, so `AddBlockchainTX` just trusts the TX object that comes in is up to date - and adds to its list.

It does call `updateTransactionsCache` within that function - so I believe we're all good here.

internal/database/sqlcommon/operation_sql.go
```diff
@@ -193,17 +193,3 @@ func (s *SQLCommon) UpdateOperation(ctx context.Context, ns string, id *fftypes.
 
 	return s.commitTx(ctx, tx, autoCommit)
 }
-
-func (s *SQLCommon) ResolveOperation(ctx context.Context, ns string, id *fftypes.UUID, status core.OpStatus, errorMsg *string, output fftypes.JSONObject) (err error) {
```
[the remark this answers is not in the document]
peterbroadhurst: Ok - so it's this path: https://github.com/hyperledger/firefly/blob/main/internal/apiserver/route_spi_patch_op_by_id.go

I think that SPI function should be deleted now, so Operations become internal again. No need for that to be in this PR.

peterbroadhurst (approved):
Well this all looks great to me.

A few bits of narration as I went through, but no suggestions for anything additional.

My overall assessment on concurrency safety with the operations is that as operations are basically just updated with the latest change, there's very little risk there. Made a specific comment about the transaction mapping to blockchain transactions - again think that's fine.

So let's go ahead and see how this performs/behaves in a long run 🚀

PR #975, opened 2022-08-17
Adjust defaults for all cache items that don't expose a Size

awrichar:
Caches may be limited on size (if the items advertise their Size) or on
number of items. Since most cached items do not expose a Size, the limit
should be set on number of items alone.

Picked numbers pretty much at random, so I'm open to feedback if they
should be higher or lower, or if any of these items should get a Size method
to do size estimation.

I've tagged with "migration_consideration" since I have renamed a few
config keys, but I don't think they're widely used.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -36,7 +36,7 @@ nav_order: 2
 
 |Key|Description|Type|Default Value|
 |---|-----------|----|-------------|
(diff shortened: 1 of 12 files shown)


shorsher (approved):

nguyer (approved):

shorsher (approved):

shorsher:
Going to go ahead and fix the build failure now

PR #988, opened 2022-08-19
Restore camelCase key names

awrichar:
Since the Viper patch has been absorbed (#930), it's OK to use
camelCase in config files again.


shorsher (approved):
🐍 🐫

Issue #944, opened 2022-08-09
Add more positive & negative E2E tests for identity (multiparty mode and non-multiparty mode)


awrichar:

awrichar:
Partially blocked by #952

anna-jackson:
I've done a very small part of this issue under #996.

Some observations from some initial attempts at some of the other negative test cases
 * `parent/child relationships that are not allowed` - this results in a 400 error response to the API call registering the identity rather than a `message_rejected` event
 * `trying to register a child of someone else's org` - I was able to successfully register an identity as a child of org 1 using the org 2 client. I don't know whether this was due to me writing code that didn't do what I thought it did, whether I've misunderstood the test case, or whether there is actually an issue here.

awrichar:
The second bullet is interesting - it might be due to the way the CLI sets up the blockchain identities. Bottom line is if you don't have access to org 1's private key, you should not be able to create a child of it. But in the local environment, everyone might have access to all the blockchain keys...

PR #996, opened 2022-08-22
Add e2e test which attempts to register an invalid identity by reusing a key

anna-jackson:
Signed-off-by: Anna Jackson <anna.jackson@kaleido.io>

Adds one of the proposed negative test cases from https://github.com/hyperledger/firefly/issues/944
> Registering invalid identities- keys that have already been used


shorsher (approved):
🔑

awrichar (approved):
Thanks for adding!

PR #998, opened 2022-08-22
Add script for auditing blockchain event order

awrichar:
Can be used to verify the event ordering logic of a given blockchain. Simply queries all blockchain events to verify that the database order is consistent with the blockchain order.

### DIFF
--- a/auditevents.sh
+++ b/auditevents.sh
@@ -39,6 +39,7 @@ lastSequence=
 lastProtocolId=
 
 echo "Checking all blockchain events on ${HOST} for increasing sequence and protocol ID."
+printf "%-12s %s\n" "Sequence" "Protocol ID"
 while [[ "$received" -eq "$limit" ]]; do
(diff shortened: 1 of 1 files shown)


awrichar:
After uploading, I realized this is only guaranteed to be valid if:
- there are no custom contract listeners
- there have been no contract migrations

Might need to enhance it further to deal with these situations.

awrichar:
Pushed a change to only examine the FireFly contract events (ie when listener==null). Could also make this configurable if we want the option to audit events from a particular listener.

And on thinking further, contract migration shouldn't be a problem, since the start block of the new contract should always be later than the end block of the previous one.

peterbroadhurst (approved):
Script looks great.

I do think we'll want to use this with Token events and custom Smart Contract events - as tests that drive on-chain logic hard (without the batching that the on-chain/off-chain coordination does) will be most likely to expose issues.

The ideal test from my perspective, would probably ERC-20 transfers (only) with a number of Confirmations enabled - driving the chain as hard as possible.

awrichar:
Yep, wanted to get a baseline defined, but it should be easy enough to add additional params to pass in source/listener.

awrichar:
May be superseded by #1002

PR #1002, opened 2022-08-24
Add tool for auditing blockchain event order

awrichar:
This is an alternative to #998, written in Go rather than bash.

Not sure where we should put scripts if we keep adding them... do we just keep adding root level folders like this?


shorsher (approved):

shorsher:
I really like the go version. I'm not sure the best way to organize all of the scripts,

PR #1006, opened 2022-08-25
Verify integrity of group for private messages

awrichar:
For pinned messages, defer all verification to the aggregator. For unpinned,
perform verifications inline on the DX callback.


internal/events/aggregator.go
```diff
@@ -328,19 +353,24 @@ func (ag *aggregator) cacheBatch(cacheKey string, batch *core.BatchPersisted, ma
...
 	for _, pin := range pins {
-		manifest, ok := manifests[*pin.Batch]
-		if !ok {
-			_, manifest, err = ag.GetBatchForPin(ctx, pin)
+		found, ok := localCache[*pin.Batch] // avoid trying to fetch the same batch repeatedly (mainly for cache misses)
+		if ok {
+			manifest = found.manifest
+			batch = found.batch
+		} else {
+			batch, manifest, err = ag.GetBatchForPin(ctx, pin)
```
peterbroadhurst, line 369: Checked that this is already cached from a DB perspective.

internal/events/aggregator.go
```diff
@@ -559,6 +587,13 @@ func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *core.Mess
 		}
 	}
 
+	if valid && msg.Header.Group != nil {
```
peterbroadhurst, line 590: I was thinking hard about the implications of moving this check fully out of the DX callback at the point we _store_ the data+messages, until the point we do the aggregation to _confirm_ the messages.

The positive is very important, and clearly required: We handle the case that the group contains identities that need to be verified in order per the the blockchain, and the aggregation of the pin for those identities might happen _after_ the receipt.

The concern is that we're reducing the amount of verification we do before we store the data+messages.

I believe there is some part we can still verify in the DX callback, and we should do that:
- We know DX would have rejected the receipt of the message, unless it came from a `peer` that was already registered
- We know the `batch` when it was received contains a node UUID
- In the DX callback I believe then we should:
   - Look up the node in the batch payload by UUID
   - Verify that the peer of the node looked up by UUID, matches the DX Peer in the DX Callback
- Then for easier visibility into the history, on unconfirmed messages, I suggest:
   - Use the Node UUID (rather than `peer`) to be the thing we store against the batch
   - Add a field to `Message` such as `node` which contains a UUID:
     - Has `omitempty` on the REST API
     - Is unset for broadcast messages
     - Filled in to "me" on origination of our own message
     - Filled in from the batch within the DX callback

peterbroadhurst (changes requested):
I've made an architectural suggestion to tweak the approach in this PR - wanted to send this back for you to consider, before continuing with the detailed review of the rest of the code.

internal/events/aggregator.go
awrichar: Yep, this is a good point. In adding the new group checks, I noted that the group often isn't available during the DX callback, because it's created by the aggregator. I somehow convinced myself that the node might similarly not be available... but you're correct that message would never have gotten delivered by DX if the node hadn't been successfully created 💡

I think that will actually simplify the change a bit.

peterbroadhurst (changes requested):
These changes look great, and I see we've simplified the change with this round of changes.

However, we might have lost one thing that was originally an inspiration for this change, and should probably go in before we merge:

We verify the _sender_ is a valid member in a group, whenever we send something (and it's "us"), and whenever we receive something. However, there's still nowhere that we verify the other other members in the group when we create it. So we could still create a group, with members that could never send data (only receive it).

So I think we should verify all members are valid at the point we create the group.

awrichar:
Good point - I added validation of all node+org combos at the point of group creation.

Notable caveats:
- This validation is only prior to sending the group. I did not add validation when you receive a group from the network (but could if it seems valuable).
- There are some duplicate queries for orgs and nodes in the group init flow here... we query them in `getRecipients()`, `getGroupNodes()`, and now again in `groupInit()`. They're cached, but it feels like it may be worth another cleanup pass to see if we can reduce these queries.

peterbroadhurst (approved):
Thanks @awrichar - think this is a great set of changes.
Appreciate you also listing the potential other changes that could be made. I do not think any of those are critical for v1.1.

PR #1017, opened 2022-08-30
adds `fetchreference` to /events/:id

### LINKED ISSUE #1016: Add `fetchreference` param for GET `/events/:id` endpoint
We currently have a `fetchreferences` query param for GET `/events` that will fetch the information of the event referenced in the api. There should be an equivalent query parameter when getting an event by `id`.

shorsher:
When provided, firefly will include the record that the `ref` field references in the response body.

closes #1016


docs/swagger/swagger.yaml
```diff
@@ -5433,6 +5433,13 @@ paths:
         required: true
         schema:
           type: string
+      - description: When set, the API will return the record that this item references
+          in its 'ref' field
```
awrichar, line 5437: Pretty sure the field is "reference" spelled out (maybe this was copy-pasted from another inaccurate field description though?)

internal/coremsgs/en_api_translations.go
```diff
@@ -42,6 +42,7 @@ var (
 	APIParamsDatatypeVersion                = ffm("api.params.datatypeVersion", "The version of the datatype")
 	APIParamsEventID                        = ffm("api.params.eventID", "The event ID")
 	APIParamsFetchReferences                = ffm("api.params.fetchReferences", "When set, the API will return the record that this item references in its 'ref' field")
+	APIParamsFetchReference                 = ffm("api.params.fetchReference", "When set, the API will return the record that this item references in its 'ref' field")
```
awrichar, line 45: Kinda wish we'd used singular `fetchreference` across the board (instead of `fetchreferences`). Maybe we could honor both on the "get many" endpoint?

awrichar (approved):
Looks good. Left a few small suggestions.

PR #1021, opened 2022-08-30
Allow any plugin to terminate the system on a bad event

### LINKED ISSUE #957: Revisit plugin handling of event processing errors
Currently most plugins treat a non-parseable event as a "stop the world" situation. They abort their event loop, log an error, and stop processing messages. We should revisit if this is the best way to handle such problems.

awrichar:
Terminating the system is better than leaving it up in a crippled state.

Closes #957


internal/blockchain/ethereum/ethereum.go
```diff
@@ -448,9 +451,9 @@ func (e *Ethereum) eventLoop() {
 				continue
 			}
 
-			// Send the ack - only fails if shutting down
 			if err != nil {
 				l.Errorf("Event loop exiting: %s", err)
```
peterbroadhurst, line 455: I think it's probably worth something really visible here like:
```suggestion
				l.Errorf("Unable to recover event loop. TERMINATING SERVER: %s", err)
```

peterbroadhurst (approved):
Thanks @awrichar - do think this is a step to a much more understandable error situation, if anyone ever gets into it.

I have made one comment (which applies to all event loops) for you to consider before merging. I have marked approval.

Issue #621, opened 2022-03-22
Documentation: Custom identity walkthrough


peterbroadhurst:
Here's a shell starting point:

```sh
> go install github.com/hyperledger/firefly-cli/ff@latest                                                                                                                                                
> # Build FireFly from main
> make docker
...

> # Start FireFly with manifest to use the local docker build, rather than latest release
> ff init -m manifest.json ids 2
initializing new FireFly stack...
Stack 'ids' created!
...

> # Start up the stack
> ff start ids
reading stack config... done
this will take a few seconds longer since this is the first time you're running this stack...
...

> # Create a signing key for new identity
> ACCOUNT=$( curl -sH 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"1","method":"personal_newAccount","params":["myunlockpw"]}'  http://localhost:5100 | jq -r '.result' )
> echo $ACCOUNT
{"jsonrpc":"2.0","id":"1","result":"0x1a57a77bf68df300fc5e7bf4f5e851596a222f66"}

> # Unlock the new signing key on the node - must be repeated after every node restart
> curl -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"1","method":"personal_unlockAccount","params":["'$ACCOUNT'","myunlockpw",0]}'  http://localhost:5100
{"jsonrpc":"2.0","id":"1","result":true}

> # Query our parent org
> # Note there's a TODO to allow you to specify the DID as an input here, rather than needing the UUID
> ORG_UUID=$( curl -s 'http://localhost:5000/api/v1/status' | jq -r '.org.id' )
> echo $ORG_UUID

> # Register with FireFly - using convenience of confirm=true as doing from the cmdline
> curl -H 'Content-Type: application/json' -d '{"name":"myid1","key":"'$ACCOUNT'","parent":"'$ORG_UUID'"}' 'http://localhost:5000/api/v1/namespaces/default/identities?confirm=true'

> # Query the new identity
> curl -s 'http://localhost:5000/api/v1/namespaces/default/identities' | jq
```

KrishnaPG:
Came across this while looking for information on how to use the Identity and Auth / Permissions with Firefly (Specifically the DID and key-exchange parts for parties). It seems all the API are open for anyone to call without identity or auth/permissions.

The documentation of Firefly does not seems to have much information on how to leverage the DIDs or the associated keys. Any info or example would be great on what are the recommendations / best-practices for: 

[...]

PR #1022, opened 2022-08-30
[identity-tutorial] identity tutorial

eberger727:
Tutorial to guide user through creating a Custom Identity in FireFly.

References https://github.com/hyperledger/firefly/issues/621


docs/tutorials/create_custom_identity.md
```diff
@@ -0,0 +1,171 @@
...
+## Quick reference
+
+Out of the box, a FireFly Supernode contains both an `org` and a `node` identity. Your use case might demand more granular notions of identity (ex. customers, clients, etc.). Instead of creating a Supernode for each identity, you can create multiple custom identities within a FireFly Supernode.
+
+## Additional info
+
+- Reference: [Identities](../reference//identities.md)
+- Swagger: <a href="../swagger/swagger.html#/Default%20Namespace/postNewIdentity" data-proofer-ignore>POST /api/v1/identities</a>
+
+### Step 1: Create a signing key for new identity
```
peterbroadhurst, line 30: Think it's important to caveat that this is an instruction with the default CLI created environment, and creating signing keys for other more complex signing configuration is not covered in this doc.

docs/tutorials/create_custom_identity.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Step 2: Unlock the new signing key on the node
+
+> This must be repeated after every node restart
+
+`POST` `http://localhost:5100`
+
+`Payload`:
+
+```json
```
peterbroadhurst, line 63: Maybe change this to `js` (rather than `json`) so it doesn't highlight the comment in 🔴

docs/tutorials/create_custom_identity.md
peterbroadhurst: Also worth noting there is a CLI command for this, so you don't have to use an API

docs/tutorials/create_custom_identity.md
```diff
@@ -0,0 +1,171 @@
...
+{
+  "jsonrpc": "2.0",
+  "id": "1",
+  "result": "0xfa140b38b81b196c8883deacd344367e17598638" // This is the new signing key
+}
+```
+
+### Step 2: Unlock the new signing key on the node
```
peterbroadhurst, line 55: This step is also done for you by the CLI

docs/tutorials/create_custom_identity.md
```diff
@@ -0,0 +1,171 @@
...
+    "claim": "dde0fa34-7595-48aa-8fe3-911d56d0cd2f",
+    "verification": "c563aed8-5bb3-4a46-b615-f552988f5361",
+    "update": null
+  }
+}
+```
+
+### Step 5: Query the New Custom Identity
+
+`GET` `http://localhost:5000/api/v1/identities`
```
peterbroadhurst, line 147: Worth demonstrating the `fetchverifiers` on this example?

peterbroadhurst (changes requested):
Think this is a great addition. Couple of requests - main one is that we should guide people to use the CLI I think for the signing key creation instructions, vs. Postman.

The Postman instructions are specific to CLI created environments anyway (they wouldn't work for example with a more sophisticated environment, such as a firefly-signer connection to a public chain, or a remote Kaleido blockchain etc.)

eberger727:
Thanks for the review @peterbroadhurst 
Don't think I'll finish the changes before rc3 goes out, but will be ready by Tuesday

docs/tutorials/create_custom_identity.md
nguyer: Agreed with noting the caveats. Very important. However, the CLI command _does not_ actually invoke this API. The CLI command is used for creating the new private key / address, and installing the key on the signer - or creating the new identity via the Fabric CA if you're using Fabric.

peterbroadhurst (approved):
Thanks both for the updates to use the CLI.

PR #1027, opened 2022-08-31
Use advisory lock instead of exclusive lock on events table

awrichar:
Advisory locks are identified with a 64-bit integer, so I've put in place a
rudimentary mapping from namespace names to a (probably) unique int64.
This should ensure consistent ordering for the writes to the events table
within a particular namespace, without blocking reads or other namespaces.

### DIFF
--- a/internal/database/postgres/postgres.go
+++ b/internal/database/postgres/postgres.go
@@ -19,6 +19,7 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"math/big"
 
 	"database/sql"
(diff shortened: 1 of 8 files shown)


peterbroadhurst (approved):
This looks great to me @awrichar - having read the docs on this `“advisory” locks can be taken on numbers that have user-defined meanings`.

Hopefully this is not only namespace specific, but also less obtrusive to PSQL's background tasks than a big fat table lock.

PR #1028, opened 2022-08-31
Add status information for subscriptions & contract listeners

shorsher:
If the `fetchstatus` query param is supplied when querying `subscriptions` or `listeners` by ID, firefly will return additional status information, if it is available.

### Contract listeners:
FF environments connected to evm connectors will return listener checkpoint formation under the `status` key.

Example query: `http://localhost:5000/api/v1/namespaces/default/contracts/listeners/92e3bc0c-3c3f-455b-81d3-4124a0270b1d?fetchstatus`

Response:
```
{
    "id": "92e3bc0c-3c3f-455b-81d3-4124a0270b1d",
    "interface": {
        "id": "30775d45-61af-4f76-9d35-69b5d8ffbf42"
    },
    "namespace": "default",
    "name": "0182f410-a686-79ed-23a5-ac8c62acbf51",
    "backendId": "0182f410-a686-79ed-23a5-ac8c62acbf51",
    "location": {
        "address": "0x62d365e68fc55845ba904acef688feba7a598c9c"
    },
    "created": "2022-08-31T13:21:48.1687995Z",
    "event": {
        "name": "Changed",
        "description": "",
        "params": [
            {
                "name": "from",
                "schema": {
                    "type": "string",
                    "details": {
                        "type": "address",
                        "internalType": "address",
                        "indexed": true
                    }
                }
            },
            {
                "name": "value",
                "schema": {
                    "type": "integer",
                    "details": {
                        "type": "uint256",
                        "internalType": "uint256"
                    }
                }
            }
        ]
    },
    "signature": "Changed(address,uint256)",
    "topic": "bsc",
    "options": {
        "firstEvent": "oldest"
    },
    "status": {
        "checkpoint": {
            "block": 0,
            "transactionIndex": -1,
            "logIndex": -1
        }
    }
}
```

### Subscriptions
Subscriptions will provide offset information if available

Example Query: `http://localhost:5000/api/v1/namespaces/default/subscriptions/32cddc74-cc1a-4493-b376-45781e2d8d35?fetchstatus`

Response:
```
{
    "id": "32cddc74-cc1a-4493-b376-45781e2d8d35",
    "namespace": "default",
    "name": "simple-storage",
    "transport": "websockets",
    "filter": {
        "message": {},
        "transaction": {},
        "blockchainevent": {}
    },
    "options": {
        "firstEvent": "-1",
        "withData": false
    },
    "created": "2022-08-31T17:13:04.295134Z",
    "updated": null,
    "status": {
        "currentOffset": 31
    }
}
```

### DIFF
--- a/internal/contracts/manager.go
+++ b/internal/contracts/manager.go
@@ -580,7 +580,9 @@ func (cm *contractManager) GetContractListenerByNameOrIDWithStatus(ctx context.C
 	}
 	status, err := cm.blockchain.GetContractListenerStatus(ctx, listener.BackendID)
 	if err != nil {
-		return nil, err
+		status = core.ListenerStatusError{
+			StatusError: err.Error(),
+		}
 	}
 	enrichedListener = &core.ContractListenerWithStatus{
 		ContractListener: *listener,
--- a/internal/contracts/manager.go
+++ b/internal/contracts/manager.go
@@ -54,6 +54,7 @@ type Manager interface {
 	AddContractListener(ctx context.Context, listener *core.ContractListenerInput) (output *core.ContractListener, err error)
 	AddContractAPIListener(ctx context.Context, apiName, eventPath string, listener *core.ContractListener) (output *core.ContractListener, err error)
 	GetContractListenerByNameOrID(ctx context.Context, nameOrID string) (*core.ContractListener, error)
+	GetContractListenerByNameOrIDWithStatus(ctx context.Context, nameOrID string) (*core.ContractListenerWithStatus, error)
 	GetContractListeners(ctx context.Context, filter database.AndFilter) ([]*core.ContractListener, *database.FilterResult, error)
 	GetContractAPIListeners(ctx context.Context, apiName, eventPath string, filter database.AndFilter) ([]*core.ContractListener, *database.FilterResult, error)
 	DeleteContractListenerByNameOrID(ctx context.Context, nameOrID string) error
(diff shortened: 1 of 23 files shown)


internal/contracts/manager.go
```diff
@@ -572,6 +573,22 @@ func (cm *contractManager) GetContractListenerByNameOrID(ctx context.Context, na
...
 }
 
+func (cm *contractManager) GetContractListenerByNameOrIDWithStatus(ctx context.Context, nameOrID string) (enrichedListener *core.ContractListenerWithStatus, err error) {
+	listener, err := cm.GetContractListenerByNameOrID(ctx, nameOrID)
+	if err != nil {
+		return nil, err
+	}
+	status, err := cm.blockchain.GetContractListenerStatus(ctx, listener.BackendID)
+	if err != nil {
+		return nil, err
```
peterbroadhurst, line 583: Just an observation that if you use `fetchstatus` this means that if the connector is not available or returns a `404`, you will get an error. I wonder @shorsher if that's the behavior we want, or we should pop a `status.error` field in there instead.

shorsher: Yep, agreed. Thanks for the suggestion 🙏🏻

shorsher: I think I made this way too complicated, but really struggled with the right way to get `status.error` to show up correctly while also including the `checkpoint` as a `JSONAny` type 😅 


Before, `cm.blockchain.GetContractListenerStatus(ctx, listener.BackendID)` would return a `JSONAny` of:
```json
{
  "checkpoint": {
     "block": 0,
     "transactonIndex": 0,
     "logIndex": 0
   },
  "catchup": false
}
```

This was great, because I could make the `ContractListenerWithStatus` struct look like:
```golang
type ContractListenerWithStatus struct {
	ContractListener
	Status *fftypes.JSONAny `ffstruct:"ContractListenerWithStatus" json:"status,omitempty" ffexcludeinput:"true"`
}
```

and set what is returned from `GetContractListenerStatus()` to `Status`, and everything would be 👍🏻. I was not sure how to update `ContractListenerWithStatus` to also add an `error` key within `status`.

shorsher: thanks for the help here, should have thought of using `interface{}`. Code is way better than what I originally came up with.

peterbroadhurst (approved):
Thanks for making those extra changes 👍

PR #1031, opened 2022-08-31
update event docs

shorsher:
in a chain with #1028 

not currently published on my github pages due to hosting another doc changes.

### DIFF
--- a/docs/tutorials/custom_contracts/ethereum.md
+++ b/docs/tutorials/custom_contracts/ethereum.md
@@ -549,7 +549,7 @@ Here is an example of sending 100 wei with a transaction:
 
 Now that we've seen how to submit transactions and preform read-only queries to the blockchain, let's look at how to receive blockchain events so we know when things are happening in realtime.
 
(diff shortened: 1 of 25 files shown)


docs/tutorials/custom_contracts/ethereum.md
```diff
@@ -747,6 +811,36 @@ You can see in the event received over the WebSocket connection, the blockchain
...
+  "filter": {
+    "events": "blockchain_event_received",
+    "message": {},
+    "transaction": {},
+    "blockchainevent": {
+      "listener": "1bfa3b0f-3d90-403e-94a4-af978d8c5b14"
+    }
+  },
+  "options": {
+    "firstEvent": "-1",
```
nguyer, line 833: I don't doubt this is actually what the API said the first event was, but I'm wondering if that is what the API was supposed to say here. `-1` doesn't make much sense to me. `0` does. But not `-1`.

nguyer (approved):
Looks good to me

PR #1034, opened 2022-09-01
Routes to directly publish blob/JSON data to shared storage (IFPS etc.)

peterbroadhurst:
A small extension to external API routes, that let you initiate uploads through your public storage plugin directly.

Unlike with the full on-chain/off-chain orchestration capabilities, no modification is made to the data, and no batching happens.

So this is the feature you would use for example to publish the metadata+binary of an NFT.


shorsher (approved):
Looks great. I know this is currently `multiparty` only, but assuming this is the first step in allowing communication with `sharedstorage` in gateway namespaces as well?

peterbroadhurst:
Yes, absolutely 👍 
The next task is to allow `/data` APIs to work in non-multiparty namespaces, with DX and shared storage plugins enabled.

PR #1035, opened 2022-09-01
remove legacy namespace query for organizations

### LINKED ISSUE #1032: 1.1 migration /api/v1/status inconsistency
I've attempted to upgrade my firefly instance from 1.0 to 1.1-rc using the migration guide

When starting up 1.1 for the first time, an API call to `/api/v1/status` shows my org is registered, but my node is not [see below for full response].  

I was expecting both the org and the node to be unregistered at this point in the migration, as the guide instructs us the org and node will need to be reregistered manually in a later step

However, a call to `/api/v1/network/organizations` returns `[]` which is in line with expectations

`/api/v1/status`
```
{
  "namespace": {
    "name": "default",
    "networkName": "default",
    "description": "",
    "created": "2022-08-31T19:30:38.411907442Z"
  },
  "node": {
    "name": "zzak1jp9qi",
    "registered": false
  },
  "org": {
    "name": "zzak1jp9qi",
    "registered": true,
    "did": "did:firefly:org/zzak1jp9qi",
    "id": "ecf12f34-f5a2-47d2-b81a-8b2897329a17",
    "verifiers": [
      {
        "type": "ethereum_address",
        "value": "0xa45436691a6574ca6c873578730af4f0a6e210ec"
      }
    ]
  },
  "plugins": {
    "blockchain": [
      {
        "name": "blockchain0",
        "pluginType": "ethereum"
      }
    ],
    "database": [
      {
        "name": "database0",
        "pluginType": "sqlite3"
      }
    ],
    "dataExchange": [
      {
        "name": "dataexchange0",
        "pluginType": "ffdx"
      }
    ],
    "events": [
      {
        "pluginType": "system"
      },
      {
        "pluginType": "websockets"
      },
      {
        "pluginType": "webhooks"
      }
    ],
    "identity": [],
    "sharedStorage": [
      {
        "name": "sharedstorage0",
        "pluginType": "ipfs"
      }
    ],
    "tokens": [
      {
        "name": "erc20_erc721",
        "pluginType": "fftokens"
      }
    ]
  },
  "multiparty": {
    "enabled": true,
    "contract": {
      "active": {
        "index": 0,
        "location": {
          "address": "0x359ce02812c4e08b492e7c2543216f98bfff7550"
        },
        "info": {
          "subscription": "sb-0319aaa9-9338-4901-5d12-543d0848d3a0",
          "version": 1
        }
      }
    }
  }
}
```

shorsher:
closes #1032

### DIFF
--- a/internal/identity/identitymanager.go
+++ b/internal/identity/identitymanager.go
@@ -415,11 +415,6 @@ func (im *identityManager) cachedIdentityLookupByVerifierRef(ctx context.Context
 	if err != nil {
 		return nil, err
 	} else if verifier == nil {
-		if namespace != core.LegacySystemNamespace && im.multiparty != nil && im.multiparty.GetNetworkVersion() == 1 {
-			// For V1 networks, fall back to LegacySystemNamespace for looking up identities
-			// This assumes that the system namespace shares a database with this manager's namespace!
-			return im.cachedIdentityLookupByVerifierRef(ctx, core.LegacySystemNamespace, verifierRef)
-		}
 		return nil, err
 	}
 	identity, err := im.database.GetIdentityByID(ctx, namespace, verifier.Identity)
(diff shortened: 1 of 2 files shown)


peterbroadhurst (approved):

awrichar:
I think this has removed a key piece of migration that we'd planned to include in 1.1. After this change, you can _no longer_ operate a V1 network with organizations that were registered to ff_system (you must re-register them on each namespace). I think we should bring back the fallback logic but find a different way to address the inconsistency pointed out in #1032.

PR #1037, opened 2022-09-02
Allow gateway init with DX and Shared Storage plugins + /data APIs

peterbroadhurst:
This allows use of the #1034 data for Gateway scenarios, such as publishing NFT metadata to IPFS.

Note that DX is used for local storage of the blob data only.

I've tried to carefully work through what is enabled:
- Removing the restriction of what plugins can be configured
- Determining which plugins/managers are enabled based on the plugins configured
- Allowing the broadcast manager to initialize without a batch manager, or multiparty
- Tweak the APIs so all the on-chain/off-chain messaging APIs are only multiparty
- Enable the data APIs only when the data manager is available


nguyer (approved):
Aside from the E2E tests which need to be addressed, looks good to me.

PR #1038, opened 2022-09-02
[address-resolver-route] POST /verifiers/resolve

eberger727:
`POST /api/v1/verifiers/resolve`
Payload:

```json
{
  "type": "ethereum_address",
  "value": "hd-1-0-3"
}
```
Response

```json
{
  "type": "ethereum_address",
  "value": "0x7301ed0266c954309e62f76749d6c75a32c7c0b1"
}
```


internal/apiserver/route_post_verifiers_resolve.go
```diff
@@ -0,0 +1,42 @@
...
+	Method:          http.MethodPost,
+	PathParams:      nil,
+	QueryParams:     nil,
+	Description:     coremsgs.APIEndpointsPostVerifiersResolve,
+	JSONInputValue:  func() interface{} { return &core.VerifierRef{} },
+	JSONOutputValue: func() interface{} { return &core.VerifierRef{} },
+	JSONOutputCodes: []int{http.StatusOK},
+	Extensions: &coreExtensions{
+		CoreJSONHandler: func(r *ffapi.APIRequest, cr *coreRequest) (output interface{}, err error) {
+			return cr.or.ResolveSigningKey(cr.ctx, r.Input.(*core.VerifierRef))
```
eberger727, line 39: I am calling `ResolveSigningKey()` from `Orchestrator` to handle the request. Should I move `ResolveSigningKey()` to the blockchain plugin instead?

internal/orchestrator/data_query.go
```diff
@@ -88,6 +88,22 @@ func (or *orchestrator) GetMessageByIDWithData(ctx context.Context, id string) (
 	return or.fetchMessageData(ctx, msg)
 }
 
+func (or *orchestrator) ResolveSigningKey(ctx context.Context, inputKey *core.VerifierRef) (*core.VerifierRef, error) {
+	if inputKey.Type != or.blockchain().VerifierType() {
```
shorsher, line 92: One thing to consider is that gateway namespaces don't require a blockchain plugin, so `or.blockchain()` could possibly be `nil`.

eberger727: Checking for it here now: https://github.com/hyperledger/firefly/pull/1038/files#diff-e097a0bac30f72e5cc6978dc4e6336e1377310f01235622293baf87758df9302R153

internal/identity/identitymanager.go
```diff
@@ -146,6 +147,34 @@ func (im *identityManager) NormalizeSigningKey(ctx context.Context, inputKey str
...
+func (im *identityManager) ResolveInputSigningKey(ctx context.Context, inputKey *core.VerifierRef) (*core.VerifierRef, error) {
+	log.L(ctx).Debugf("Resolving input signing key: type='%s' value='%s'", inputKey.Type, inputKey.Value)
+
+	if im.blockchain == nil {
+		return nil, i18n.NewError(ctx, coremsgs.MsgBlockchainNotConfigured)
+	}
+
+	verifierType := inputKey.Type
+
+	if verifierType.String() == "" {
```
eberger727, line 159: If verifier type is not given, the default is used from the blockchain plugin

internal/identity/identitymanager.go
```diff
@@ -146,6 +147,34 @@ func (im *identityManager) NormalizeSigningKey(ctx context.Context, inputKey str
...
+		verifierType = im.blockchain.VerifierType()
+	}
+
+	if verifierType != im.blockchain.VerifierType() {
```
eberger727, line 163: If verifier type is invalid, an error is thrown. I think this is better than using the default blockchain plugin, as it will better inform the user

peterbroadhurst (approved):
Looks great, thanks @eberger727

PR #1051, opened 2022-09-12
Manifest with head builds

peterbroadhurst:
In draft, but raised early to get a full coverage of builds:
- This currently has `head` builds of ERC-20/ERC-721 and ERC-1155 docker containers
- We do not yet have an FFTM/EVMConnect head build, until these go in:
    - https://github.com/hyperledger/firefly-transaction-manager/pull/37 - needs a release
    - https://github.com/hyperledger/firefly-evmconnect/pull/33 - updated to release from above

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6,8 +6,8 @@
   },
   "evmconnect": {
     "image": "ghcr.io/hyperledger/firefly-evmconnect",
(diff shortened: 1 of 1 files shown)


peterbroadhurst:
replacing with v1.1.0 target manifest updates

PR #1062, opened 2022-09-16
Fix casing on docs tutorial pages

nguyer:
Standardizing on lowercase names for markdown files here. Previously we had some of both which caused issues on case insensitive file systems like macOS.


peterbroadhurst (approved):
👍

PR #1068, opened 2022-09-22
make contract listener `location` optional

### LINKED ISSUE #1066: Allow contract listeners to be created without `location`
FireFly should remove the requirement for `location` when creating a contract listener. A common use case is subscribing to events that match a specific signature, regardless of the contract of origin. This is especially common when FireFly needs to interact with contracts that have been instantiated by a factory, a single contract listener would be able to capture all events from the created contracts. 

This is already supported by `evmconnect`, so the changes should just be within the contracts manager & blockchain plugins.

shorsher:
closes #1066

### DIFF
--- a/db/migrations/postgres/000100_allow_null_listener_location.up.sql
+++ b/db/migrations/postgres/000100_allow_null_listener_location.up.sql
@@ -1,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE contractlisteners ALTER COLUMN location DROP NOT NULL;
+
+COMMIT;
--- a/internal/blockchain/ethereum/eventstream.go
+++ b/internal/blockchain/ethereum/eventstream.go
@@ -184,12 +184,16 @@ func (s *streamManager) createSubscription(ctx context.Context, location *Locati
 		fromBlock = "latest"
 	}
 	sub := subscription{
-		Name:             subName,
-		Stream:           stream,
-		FromBlock:        fromBlock,
-		EthCompatAddress: location.Address,
-		EthCompatEvent:   abi,
+		Name:           subName,
+		Stream:         stream,
+		FromBlock:      fromBlock,
+		EthCompatEvent: abi,
 	}
+
+	if location != nil {
+		sub.EthCompatAddress = location.Address
+	}
+
 	res, err := s.client.R().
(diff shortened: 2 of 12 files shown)


db/migrations/postgres/000100_allow_null_listener_location.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
```
awrichar, line 2: Probably need a SQLite migration (which would be a full table copy since you can't drop a constraint) _or_ if that's too heavy, need to edit the creation of the original SQLite table creation to not include NOT NULL in the first place (and note migrations won't be supported).

shorsher: thank you, not sure how I forgot that 😅

internal/blockchain/ethereum/eventstream.go
```diff
@@ -184,12 +184,16 @@ func (s *streamManager) createSubscription(ctx context.Context, location *Locati
...
-		EthCompatAddress: location.Address,
-		EthCompatEvent:   abi,
+		Name:           subName,
+		Stream:         stream,
+		FromBlock:      fromBlock,
+		EthCompatEvent: abi,
 	}
+
+	if location != nil {
+		sub.EthCompatAddress = location.Address
```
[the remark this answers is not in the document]
shorsher: Absolutely 😄

internal/blockchain/fabric/eventstream.go
```diff
@@ -160,16 +160,20 @@ func (s *streamManager) createSubscription(ctx context.Context, location *Locati
...
+		Stream: stream,
 		Filter: eventFilter{
-			ChaincodeID: location.Chaincode,
 			EventFilter: event,
 		},
 		FromBlock: fromBlock,
 	}
+
+	if location != nil {
+		sub.Channel = location.Channel
```
awrichar, line 173: Might want to confirm that fabconnect functions as expected when this is unset

awrichar (approved):

shorsher:
Need to do a quick change to how Fabric handles `Location`, so moving this to draft.

PR #1070, opened 2022-09-22
use `cm.namespace` when generating contract URL's 

### LINKED ISSUE #1067: swagger/openapi url response is missing namespace in POST response
The openapi/swagger url's returned when creating a custom contract api seem to omit the namespace now, which breaks the link.

example response:
```
{
    "id": "0ee89842-3db2-43f5-8469-9f894d616332",
    "namespace": "default",
    "interface": {
        "id": "cf7c0b1f-7b90-4442-9794-58afcc8b5ba1"
    },
    "location": {
        "address": "0xd22d700e9f8b3ecac72126d1bf6fefc513787954"
    },
    "name": "simple-storage",
    "message": "882a64f2-7f71-4a9a-9eb9-90ca0f56ce2c",
    "urls": {
        "openapi": "http://127.0.0.1:5000/api/v1/namespaces//apis/simple-storage/api/swagger.json",
        "ui": "http://127.0.0.1:5000/api/v1/namespaces//apis/simple-storage/api"
    }
}
```

Note, the response from `GET` `/apis` works great.

shorsher:
~~Fixes error where namespace was not set when resolving contract API on create. This resulted in contract listener url's being broken in the `POST` response.~~

When generating contract URL's, always use the `ContractManager` namespace.

closes #1067

### DIFF
--- a/internal/contracts/manager.go
+++ b/internal/contracts/manager.go
@@ -257,7 +257,7 @@ func (cm *contractManager) resolveInvokeContractRequest(ctx context.Context, req
 func (cm *contractManager) addContractURLs(httpServerURL string, api *core.ContractAPI) {
 	if api != nil {
 		// These URLs must match the actual routes in apiserver.createMuxRouter()!
-		baseURL := fmt.Sprintf("%s/namespaces/%s/apis/%s", httpServerURL, api.Namespace, api.Name)
+		baseURL := fmt.Sprintf("%s/namespaces/%s/apis/%s", httpServerURL, cm.namespace, api.Name)
 		api.URLs.OpenAPI = baseURL + "/api/swagger.json"
 		api.URLs.UI = baseURL + "/api"
 	}


internal/definitions/sender_contracts.go
```diff
@@ -55,17 +55,17 @@ func (bm *definitionSender) DefineContractAPI(ctx context.Context, httpServerURL
 	api.ID = fftypes.NewUUID()
 
 	if bm.multiparty {
-		if err := bm.contracts.ResolveContractAPI(ctx, httpServerURL, api); err != nil {
```
awrichar, line 58: I think we still want to do the resolution before broadcasting the definition... but perhaps we need to set the namespace, then clear it before the broadcast, then set it again before we return...

shorsher: The other solution I was proposing was to just use the `ContractManager` namespace here: https://github.com/hyperledger/firefly/blob/1f31807dd98c22f45f00528cebb26071d5b25457/internal/contracts/manager.go#L257

awrichar: Yea, that seems like a fine solution

awrichar (approved):

Issue #1025, opened 2022-08-31
Webhook subscriptions not receiving all events


ivis-tsukioka:
I set up a subscription of webhook for FireFly to send events to our app. 
But our app have received a part of  event's notice from FireFly.
I think that webhook subscription is permanence.
Is it right?

I send request body(below json) for "[POST] /api/v1/namespaces/default/subscriptions" to set up a subscription of webhook.
Did I get a wrong setting?

Now, after receiving event's notice, I set up a new a subscription of webhook to continuously receive event.

request body
```json
{
    "options": {
        "url": "http://<IP address>:8082/emit",
    	"fastack": false,
    	"readAhead":100,
    	"method": "POST",
    	"withData": false,
        "firstEvent": "oldest",
        "headers" : {
            "Authorization": "<Bearer token>"
        }
    },

    "namespace": "default",
    "name": "emitSub",

    "transport": "webhooks",
    "filter": {
        "topic": "string",
        "events": "blockchain_event_received",
        "blockchainevent": {
            "message": {},
            "transaction": {},
            "listener": "<listener_id>"
        }
    }
}
```

nguyer:
Hi @ivis-tsukioka. If you have questions about how subscriptions and webhooks work, I would suggest asking in the Hyperledger Discord server in the #firefly channel https://discord.gg/nnQw2aGhX6 You will likely get a much better response there from the community.

If you suspect there is a bug in FireFly, please feel free to re-open this issue, and clearly indicate:
- The exact steps to reproduce what you saw
- The actual result
- The expected result

Thanks!

nguyer:
After further investigation, I do think there is likely a bug here, so I'm reopening this one

nguyer:
I did some testing on this and was able to reproduce what @ivis-tsukioka was seeing.

- I have a custom smart contract that emits 10 events when a certain function executes
- I have a subscription set up to send webhooks to a simple web server (written in Go) that will log the request and always send a 200.
- When I invoke the function I get one, or maybe two events (haven't figured out what makes the difference) and then nothing else. I see all 10 events in the FireFly Explorer, so FireFly got them.
- If I invoke the function again, I get nothing so it seems like the subscription is "stuck", but I'm pretty confident my web server should be "acking" the request with a 200 response each time.
- If I delete and recreate the subscription, up to the first two events are re-delivered and it gets "stuck" again.

nguyer:
[...]

PR #1072, opened 2022-09-22
Fix acknowledgement for webhooks in non-reply, non-fastack cases

peterbroadhurst:
Raised from #1025 

- Adds better logging
- Ensures we do a `DeliveryResponse` in the case of a non-`reply`, non-`fastack` webhook subscription
- Disabled `fastack` when `reply` is configured, as the reply would not be sent otherwise
- (Re)instate the lost `events` section in the config docs
- Correct the docs to reflect that if you want to retry on non-`2xx` error codes, you need to set `events.webhooks.retry`

### DIFF
--- a/internal/events/webhooks/webhooks.go
+++ b/internal/events/webhooks/webhooks.go
@@ -63,12 +63,13 @@ type whResponse struct {
 func (wh *WebHooks) Name() string { return "webhooks" }
 
 func (wh *WebHooks) Init(ctx context.Context, config config.Section) (err error) {
+	connID := fftypes.ShortID()
 	*wh = WebHooks{
-		ctx:          ctx,
+		ctx:          log.WithLogField(ctx, "webhook", wh.connID),
 		capabilities: &events.Capabilities{},
 		callbacks:    make(map[string]events.Callbacks),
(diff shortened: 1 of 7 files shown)


internal/events/webhooks/webhooks.go
```diff
@@ -223,8 +226,10 @@ func (wh *WebHooks) attemptRequest(sub *core.Subscription, event *core.EventDeli
 		}
 	}
 
+	log.L(wh.ctx).Debugf("Webhook-> %s %s event %s on subscription %s", req.method, req.url, event.ID, sub.ID)
 	resp, err := req.r.Execute(req.method, req.url)
 	if err != nil {
+		log.L(wh.ctx).Errorf("Webhook<-! %s %s event %s on subscription %s failed: %s", req.method, req.url, event.ID, sub.ID, err)
```
nguyer, line 232:
```suggestion
		log.L(wh.ctx).Errorf("Webhook<- %s %s event %s on subscription %s failed: %s", req.method, req.url, event.ID, sub.ID, err)
```

Removing the `!` would make this consistent with the other log messages for webhooks

nguyer (approved):
Looks good! Thanks!

PR #1073, opened 2022-09-22
V1.0.x fix: Allow update of node using parent org identity

peterbroadhurst:
Testing patching Node identity for Cert cycling, I found that in v1.0 we were missing the API from the Swagger, but it did exist

`PATCH` `namespaces/{ns}/identities/{iid}`

However, attempt to run it for a node resulted in the following - due to a discrepancy between the checking for the author of the message between identity creation, and identity update:
`[2022-09-22T22:39:53.629Z]  WARN node_0: Invalid identity update message 2ac79c71-3983-45a4-8942-7f1bc124225b - wrong author: did:firefly:org/org_0 dbtx=reAmSQrS role=aggregator`

This PR is a V1.0 stream fix to correct this. A separate PR will be needed for the V1.1 stream - in #1074

### DIFF
--- a/internal/definitions/definition_handler_identity_update.go
+++ b/internal/definitions/definition_handler_identity_update.go
@@ -48,8 +48,17 @@ func (dh *definitionHandlers) handleIdentityUpdateBroadcast(ctx context.Context,
 		return HandlerResult{Action: ActionReject}, nil
 	}
 
+	parent, retryable, err := dh.identity.VerifyIdentityChain(ctx, identity)
+	if err != nil && retryable {
+		return HandlerResult{Action: ActionRetry}, err
+	} else if err != nil {
+		log.L(ctx).Infof("Unable to process identity update (parked) %s: %s", msg.Header.ID, err)
+		return HandlerResult{Action: ActionWait}, nil
+	}
+
 	// Check the author matches
-	if identity.DID != msg.Header.Author {
+	expectedSigner := dh.getExpectedSigner(identity, parent)
+	if expectedSigner.DID != msg.Header.Author {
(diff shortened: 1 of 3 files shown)


peterbroadhurst:
Screenshot showing a successful patch to an identity after applying this 
image

internal/definitions/definition_handler_identity_update.go
```diff
@@ -48,8 +48,17 @@ func (dh *definitionHandlers) handleIdentityUpdateBroadcast(ctx context.Context,
 		return HandlerResult{Action: ActionReject}, nil
 	}
 
+	parent, retryable, err := dh.identity.VerifyIdentityChain(ctx, identity)
+	if err != nil && retryable {
+		return HandlerResult{Action: ActionRetry}, err
+	} else if err != nil {
+		log.L(ctx).Infof("Unable to process identity update (parked) %s: %s", msg.Header.ID, err)
```
awrichar, line 55: Why is this one parked? This would be for things like an unknown or unconfirmed parent identity. All other failures here seem to result in outright rejection of the update.

awrichar (approved):
Approved, but with one question about error handling

internal/definitions/definition_handler_identity_update.go
peterbroadhurst: In this case, we couldn't resolve something up the parent stack. The code in creation (which this was copied from) did a wait at this point, on the basis there could be something in the parent stack still being resolved.

I believe it's technical also possible when replaying a chain for that to be true here. Or at least I've not done enough analysis of the topics used on the messages in that scenario to rule it out. So I left the logic equivalent in create and update paths.

nguyer (approved):

Issue #1056, opened 2022-09-14
extend firefly-core with plugins without requiring a fork


hosie:
It seems that the only way to package a plugin into firefly-core is to make code changes in the `firefly` repo itself.  Which means that the plugin developer effectively needs to maintain a fork and track origin.

A couple of thoughts on how we could make this more elegant
 - rather than ( or at least, as well as)  having a hard coded list of [transport providers](https://github.com/hyperledger/firefly/blob/main/internal/events/eifactory/factory.go#L30-L34)  we could introduce a way to dynamically register implementations of the `event.Plugin` interface.  i.e. similar to the registration of policy engines in [firefly-transaction-manager](https://github.com/hyperledger/firefly-transaction-manager/blob/main/pkg/policyengines/registry.go#L45)
 - move the [bootstrapping code](https://github.com/hyperledger/firefly/blob/main/cmd/firefly.go#L101) into `pkg` so that a plugin developer can build their own golang executable to extend `firefly-core` with their plugin and only provide a very thin `cmd` package that simply invokes the `firefly-core` bootstrap ( after registering the plugins as my first bullet 1)
 
 Full disclosure: I am not a golang expert so there may be a much more elegant way to achieve this that I am not seeing.

peterbroadhurst:
Think both of these are a great step forwards. They match how Firefly Transaction Manager evolved, and we already took part of pattern and back-ported to a small piece of Core (via common) with the API authentication plugin: https://github.com/hyperledger/firefly-common/blob/8463e27fe490a87061a6b75080b45a0b9a535f32/pkg/auth/authfactory/factory.go#L39-L46

nguyer:
I think this is a great idea as well. There is actually an instance of the first pattern in `firefly-common` with the auth plugin.

https://github.com/hyperledger/firefly-common/blob/main/pkg/auth/authfactory/factory.go#L39-L46

We wanted to have the same auth mechanism available for all FireFly microservices, so the plugin interface, and the basic auth implementation went into `firefly-common`. But other services could call `RegisterPlugins()` to "install" their own authorizer function.

It's probably a bit different than how we want the general case to work here, but just thought I'd point it out in case it's useful.

PR #1075, opened 2022-09-26
Plugin registry (issue #1056)

hosie:


hosie:
This is the first wave of changes for issue #1056 .
This PR
 - refactors the bootstrap code to make it easier to extend with a new executable
 - add dynamic registration to the event plugin factory
 
 future PRs will add dynamic registration to the other plugin factories

internal/events/webhooks/webhooks.go
```diff
@@ -60,7 +60,8 @@ type whResponse struct {
 	Body    *fftypes.JSONAny   `json:"body"`
 }
 
-func (wh *WebHooks) Name() string { return "webhooks" }
+func Name() string                { return "webhooks" }
```
peterbroadhurst, line 63: Struggling to understand why we've stopped this being on the interface - will come back to this after rest of review.

pkg/events/eifactory/factory.go
```diff
@@ -18,40 +18,49 @@ package eifactory
...
+var pluginsByName = map[string]func() events.Plugin{
+	websockets.Name(): func() events.Plugin { return &websocketsPlugin },
+	webhooks.Name():   func() events.Plugin { return &webhooksPlugin },
+	system.Name():     func() events.Plugin { return &systemEventsPlugin },
 }
 
-var pluginsByName = make(map[string]events.Plugin)
+// Allows other code to register additional implementations of Event plugins
+// that can also be initialized with this factory
+func RegisterPlugins(plugins map[string]func() events.Plugin) {
```
peterbroadhurst, line 44: I think this should be a singleton, where the `Name()` function of the plugin is used to determine the name

peterbroadhurst: Ah, ok - so the interface here is to register a new factory for the plugin.
Because we need to be able to generate lots of instances of it.

In which case, I think we should define an `EventPluginFactory` interface, and have `Name()` as a function on that interface, as well as `NewInstance()` or similar.

pkg/firefly/firefly.go
```diff
@@ -0,0 +1,173 @@
...
+		return _utManager
+	}
+	return namespace.NewNamespaceManager(true)
+}
+
+func Run(cfgFile string) error {
+
+	// Read the configuration
+	ResetConfig()
+	err := config.ReadConfig(configSuffix, cfgFile)
```
peterbroadhurst, line 76: Sorry for having a hazy memory of the interface here on Viper, but is this actually the file, or is this a filename prefix that is used for searching?

I ask because firstly to confirm there's no change in behavior here, but hen also I think this has been a point of confusion in the code before - so it might be good to be explicit in the comments as an additional improvement here.

peterbroadhurst (commented):
The move of the code into the FireFly package looks great.

Couple of comments/questions as I try to understand the proposal for the new registration based interface, which I assume would run to all other plugins.

hosie:
Thank you @peterbroadhurst this is great.

I really like the idea of the factory interface.  That might also be a good opportunity to align the code with the config spelling wrt `name` vs `type`.  e.g. when registering an instance of a `factory`, you would provide its `type`.  When someone wants to use your factory to create an instance of the `plugin`, then they provide a `name`.  

e.g.
```go
type Factory interface {
	Type() string
	NewInstance(name string) Plugin
}
```
which would be registered via
```go
func RegisterPluginFactory(factory Factory)
```

Maybe the need for this becomes more obvious when we look at the other plugins.  

e.g. for the given config
```yaml
plugins:
  database:
  - name: database0
    type: sqlite3
    sqlite3:
      url: /etc/firefly/db?_busy_timeout=5000
      migrations:
        auto: true
  - name: database1
    type: postgres
    postgres:
      url: postgresql://mypostgres/mydb
      migrations:
        auto: true        
  - name: database2
    type: postgres
    postgres:
      url: postgresql://myotherpostgres/mydb
      migrations:
        auto: false
```
There  needs to have been 2 calls to `RegisterPluginFactory`.  One with a factory whose `Type` is `sqlite3` and the other with `Type` `postgres`.  The namespace manager would then call `NewInstance` on the `sqlite` factory (with name `database0` and then call `NewInstance` twice on the `postgres` factory with names `database1` and `database2`.

However, given your comment
> which I assume would run to all other plugins.

it is worth mentioning that event plugins are special.  They don't have a name, only a type. e.g.
```
 transports:
   enabled: ["websockets","webhooks","system"]
```
And also, the rest of the code base assumes there is only ever one instance of each type.

i.e 

when we call `eifactory.GetPlugin` with type `websockets` here https://github.com/hyperledger/firefly/blob/main/internal/apiserver/route_get_websockets.go#L41 we must get back exactly the same golang object that was returned here https://github.com/hyperledger/firefly/blob/main/internal/namespace/manager.go#L1095 (and initialised a few lines later).

So there are a couple of options
 a) have a single pattern that applies to all plugins
 b) treat event plugins as special.

Given that, for all plugins, we need to define the behaviour in the case where `GetPlugin` is called twice with the same `name` and `type`, and _it returns the existing instance with that name or else creates a new instance _ seems like a sensible behaviour then it seems like `a` is a good option.  

I would propose that a module in firefly core provides that behaviour rather than asserting that the plug in providers do.  e.g. maybe rename `difactory` and `eifactory` etc.. to `diregistry` `eiregistry` and they would hold a list of factories ( each with a unique `type`) and a list of Plugin objects ( with a unique `type/name` tuple) . They registry would implement a `GetPlugin` function that takes `name` and `type` and only calls `NewInstance` on the factory for that `type` if it does not already have an instance with that `name`.

For event plugins, then we would just always use the same `name` ( e.g. `nil` or `""` or `default`).

peterbroadhurst:
The thing that seems to be stopping the Events interface being different, is the fact that we have `name` passed into `NewInstance(name string) Plugin`.

e.g. somehow the instance of the plugin itself seems to need to be aware of its name, or even to be responsible for storing it.

Would it be easier to make the storing of `name` the responsibility of the Manager that instantiates the plugin instead?

hosie:
Yes good point.  The namespace manager currently does that for all other plugins ~but not for events~. Actually, as I look closer, I see that the namespace manger does pass the instantiated instances of the event plugins down to the orchestrator for each namespace.  So the 3 cases that were causing the problem by assuming `eifactory.GetPlugin` does *not* create new instances are the exception
 -  https://github.com/hyperledger/firefly/blob/main/internal/apiserver/route_get_websockets.go#L41
 - https://github.com/hyperledger/firefly/blob/main/internal/apiserver/server.go#L355
 - https://github.com/hyperledger/firefly/blob/main/internal/events/event_manager.go#L162

So, if we can figure out a contract between apiserver and either namespace manager or orchestrator to get hold of the existing instance of the websocket event plugin , then everything stays nice and clean.  Factory just worries about creating new instances and namespace manager worries about managing those instances.

... or we accept that events are different. That they are singletons and the plugins do not register a factory, they each just register a single instance of the Plugin object.  That certainly seems easiest at this stage.  I guess the question is whether we want to allow multiple instances of event plugins in the future (e.g. maybe connecting to different brokers for different namespaces).

awrichar:
Since namespace manager is responsible for parsing the config and instantiating the events plugins, I think it makes sense to also expose a "getter" for the websockets plugin. Since the API server always has access to namespace manager, it can ask for the websockets plugin at the two points it's required.

hosie:
need to spend some time sorting out the merge

PR #1088, opened 2022-10-11
Allow listening for go debug on all interfaces, not just localhost

matthew1001:
It's difficult using the go pprof debug utility when running in a container since the network interface it binds to defaults to localhost. This PR makes 2 changes:

1. Make the interface address configurable
2. Change the default interface to "*"

It still requires changes to the docker run command or docker compose overrides file to get a route through to the debug listener, but doesn't require you to rebuild firefly.

### DIFF
--- a/internal/coreconfig/coreconfig.go
+++ b/internal/coreconfig/coreconfig.go
@@ -209,6 +209,8 @@ var (
 	PluginsIdentityList = ffc("plugins.identity")
 	// DebugPort a HTTP port on which to enable the go debugger
 	DebugPort = ffc("debug.port")
+	// DebugAddress the HTTP interface for the debugger to listen on
+	DebugAddress = ffc("debug.address")
 	// EventTransportsDefault the default event transport for new subscriptions
 	EventTransportsDefault = ffc("event.transports.default")
 	// EventTransportsEnabled which event interface plugins are enabled
@@ -359,6 +361,7 @@ func setDefaults() {
 	viper.SetDefault(string(CacheOperationsTTL), "5m")
 	viper.SetDefault(string(HistogramsMaxChartRows), 100)
 	viper.SetDefault(string(DebugPort), -1)
+	viper.SetDefault(string(DebugAddress), "")
 	viper.SetDefault(string(DownloadWorkerCount), 10)
 	viper.SetDefault(string(DownloadRetryMaxAttempts), 100)
 	viper.SetDefault(string(DownloadRetryInitDelay), "100ms")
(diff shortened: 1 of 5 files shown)


internal/coreconfig/coreconfig.go
```diff
@@ -359,6 +361,7 @@ func setDefaults() {
 	viper.SetDefault(string(CacheOperationsTTL), "5m")
 	viper.SetDefault(string(HistogramsMaxChartRows), 100)
 	viper.SetDefault(string(DebugPort), -1)
+	viper.SetDefault(string(DebugAddress), "")
```
peterbroadhurst, line 364: The suggest the defaults stays as `localhost`, as exposing the debug server externally is not something I believe users would want to do outside of a dockerized/k8s environment where they have carefully considered the security implications.

A follow-up PR with `firefly-cli` to have a flag to allocate ports to the debug ports of various services (`ethconnect`/`evmconnect`/`core` etc.) might be valuable for dev environments.

internal/coreconfig/coreconfig.go
```diff
@@ -209,6 +209,8 @@ var (
 	PluginsIdentityList = ffc("plugins.identity")
 	// DebugPort a HTTP port on which to enable the go debugger
 	DebugPort = ffc("debug.port")
+	// DebugAddress the HTTP interface for the debugger to listen on
+	DebugAddress = ffc("debug.address")
```
peterbroadhurst, line 213: Think you'll need to add a documentation entry for this one too.

peterbroadhurst (changes requested):
Would like to avoid changing the default on this one if that's ok.

internal/coreconfig/coreconfig.go
matthew1001: Yeah I almost left the default as localhost - I'll revert it back. Happy to look at the firefly-cli/firefly-* PRs after this one.

internal/coreconfig/coreconfig.go
matthew1001: Good point, let me do that

peterbroadhurst: Here's where to add the config doc:
https://github.com/hyperledger/firefly/blob/892ebae4d2202a4abc59aa056aaffbe8849cd456/internal/coremsgs/en_config_descriptions.go#L219

Then `make reference` before you check in.

matthew1001:
Both comments should be addressed in the new commit. I notice that we have a config test so I updated that, although it feels like it's of limited use as it's just testing the defaults haven't been changed. Didn't seem to do any harm to add a couple of tests to it though.

internal/coreconfig/coreconfig_test.go
```diff
@@ -29,4 +29,6 @@ func TestInitConfigOK(t *testing.T) {
 	Reset()
 
 	assert.Equal(t, 25, config.GetInt(APIDefaultFilterLimit))
+	assert.Equal(t, "localhost", config.GetString(DebugAddress))
```
peterbroadhurst, line 32: Yeah, this test is for the config scaffolding code, rather than for individual fields - but all good 👍

peterbroadhurst (approved):

PR #1089, opened 2022-10-14
Set firstEvent default using string

peterbroadhurst:
`viper.GetString` does not return a value, if the default was set to a type that wraps string.

So the fact we were setting the default for `firstEvent` as `core.SubOptsFirstEventOldest`, meant the default did not work.

This is a significant bug, as it changes the default in v1.1 to listen from the current block (the EthConnect/FabConnect/EVMConnect default for an unset value), rather than the beginning of the blockchain (as we did in v1.0).


nguyer (approved):
Looks great. Thanks for sorting this one out!

PR #1095, opened 2022-11-07
Add endpoint for deploying contracts

nguyer:


internal/blockchain/ethereum/ethereum.go
```diff
@@ -679,6 +683,45 @@ func (e *Ethereum) SubmitNetworkAction(ctx context.Context, nsOpID string, signi
...
+	}
+	if signingKey != "" {
+		body["from"] = signingKey
+	}
+	body, err := e.applyOptions(ctx, body, options)
+	if err != nil {
+		return err
+	}
+
+	client := e.fftmClient
```
peterbroadhurst, line 709: Note that this is redundant now I believe

peterbroadhurst (approved):
Looks great.

I know we've discussed a follow-up, to detect if EthConnect returns an error due to not supporting this spelling of the `ContractDeploy` function.

Two additional work items I noticed in the review:
1. Directly related to change: Adding the new event types to this table: https://github.com/hyperledger/firefly/blob/1bb1675b41ef80875da870ae6b738d675c136364/docs/reference/types/_includes/event_description.md#reference-
2. Unrelated cruft added to by this PR: Remove the defunct `fftm` configuration, from early in the v1.1 cycle when we had an option for a 2nd URL as not everything had moved to the new connector architecture: https://github.com/hyperledger/firefly/blob/975d4b7855d3b898ce0b8c02e3f47299c21d7757/internal/blockchain/ethereum/ethereum.go#L155-L157

PR #1096, opened 2022-11-08
Idempotent apis

peterbroadhurst:
This feature allows API requests to FireFly Core to be made idempotent.

That is that the caller of the API can specify its own unique identifier (an arbitrary string up to 256 characters) that uniquely identifies the request. So if there is a network connectivity failure, or an abrupt termination of either runtime, the application can safely attempt to resubmit the REST API call and be returned a `409 Conflict` HTTP code.

Examples of how an app might construct such an idempotencyKey include:
- Unique business identifiers from the request that comes into its API up-stream - passing idempotency along the chain
- A hash of the business unique data that relates to the request - maybe all the input data of a blockchain transaction for example, if that payload is guaranteed to be unique.
    > Be careful of cases where the business data might _not_ be unique - like a transfer of 10 coins from A to B... that
    > could happen multiple times, and each would be a separate business transaction. Where as transfer with invoice
    > number abcd1234 of 10 coins from A to B might be perfectly unique.
- A unique identifier of a business transaction generated within the application and stored in its database before submission
    > Here be careful you haven't just moved the problem up one layer. How does that unique ID get generated? Is that
    > itself idempotent?

FireFly already uses an idempotent interface downstream to key plugins:
   - EVMConnect blockchain operations - fully idempotent, using the operation ID
   - Token operations - inherit idempotence from the blockchain connector they communicate with

> Note that the interface between FireFly and EthConnect for blockchain operations is not currently fully idempotent.
> This PR does not make the changes needed for that, and I note that EVMConnect is the recommended choice for
> new FireFly projects.

### Tasks

- [x] Add `idempotencyKey` field to messages
   - Persisted directly on `Message` object
   - Local only - not propagated in `persistBatch` when received remotely
   - Note complexity was in the batching logic, as it's important if the message writer batching has one out of a set of messages that fails the idempotency check, it doesn't fail the whole batch of message/data inserts. To do this without reducing performance for the good case took a bit of extra work
- [x] Add `idempotencyKey` to `invoke` of blockchain smart contracts
    - Persisted on `Transaction` object
- [x] Add `idempotencyKey` to `deploy` of blockchain smart contracts
    - Persisted on `Transaction` object
- [x] Add `idempotencyKey` to `transfer` (/`mint`/`burn`/`approval`) of tokens
    - Persisted on `Transaction` object
- [x] Add `idempotencyKey` to `publish` APIs
- [x] Update e2e tests for Message to test idempotent submission
- [x] Update e2e tests for Tokens to test idempotent submission
- [x] Update e2e tests for Custom contracts to test idempotent submission

> I did look at in this PR, whether I could pass the `idempotencyKey` back to applications on events if they submitted
> them on a request. However, this proved to need more investigation because the `.transaction` field of events is
> only enriched on `transaction_submitted` events. To switch it to be cache-back-queried on every event, and to bloat
> the JSON with the full transaction for all events, felt like a change that needed discussion independent from adding
> the `idempotencyKey` capability to the submission APIs.


internal/data/message_writer.go
```diff
@@ -154,29 +165,104 @@ func (mw *messageWriter) writerLoop(index int) {
...
+		l <- err
+	}
+}
+
+func (mw *messageWriter) checkIdempotencyDuplicate(ctx context.Context, m *core.Message) error {
+	if m.IdempotencyKey != "" {
+		fb := database.MessageQueryFactory.NewFilter(ctx)
+		existing, _, err := mw.database.GetMessages(ctx, m.Header.Namespace, fb.Eq("idempotencykey", (string)(m.IdempotencyKey)))
+		if err != nil {
+			// Don't overwrite the original error for this - return -1 to the caller, who will return the previous error
```
awrichar, line 215: `-1`?

internal/data/message_writer.go
```diff
@@ -154,29 +165,104 @@ func (mw *messageWriter) writerLoop(index int) {
...
+		if idempotencyErr != nil {
+			// We have an idempotency duplicate - we need to remove it from the batch so we can retry the rest
+			dupRemoved = true
+			if listener := batch.listeners[*m.Header.ID]; listener != nil {
+				// Notify the listener for this message,
+				listener <- idempotencyErr
 			}
-			batch = nil
+			delete(batch.listeners, *m.Header.ID)
+			// Remove all the data associated with this message from the batch
```
awrichar, line 242: Is there an edge case where two messages in the same batch could reference the same piece of data, and only one gets removed?

awrichar (approved):
Looks very good. One edge case noted for consideration.

PR #1098, opened 2022-11-08
Add forward compatibility with Go 1.19

nguyer:


peterbroadhurst (approved):
Nice to know `ioutil` functions are fully blessed `io` functions now.

I did confirm that happened a good while ago (1.16), so we shouldn't be causing problems for anyone who has a tactical need to re-instate a 1.17 build.

PR #1111, opened 2022-12-05
Add 'fetchstatus=true' query param to /operations/{opid}

### LINKED ISSUE #1108: Move source of truth for EVMConnect/FFTM blockchain transaction status to from Core to FFTM
In https://github.com/hyperledger/firefly/issues/989 we had an observation as follows:

> We are now spending a lot of time processing receipts, and updating operations, with how chatty EVMConnect is. Large payloads, more often. Optimizing this would be best achieved by switching the webhooks from EVMconnect to just be status changes, and making the large payload be queried through on demand (as we made changes like this on listeners for example).

There is another challenge with the fact that we do best-effort WebSocket updates of operations from EVMConnect -> FF Core:
- If a websocket event is missed that changes a tranasction from Pending -> Succeeded, we will never reconcile that state

I propose a straight forward change that is aligned with a number of recent tweaks (specifically in #1028 ):
- We introduce a `status` section to `operation` 
- We only send a WebSocket updates from FFTM (EVMConnect) to Core when they result in a state transition:
    - `Pending`->`Succeeded`/`Failed`
- We only include a small `output` body in these WebSocket transactions
    - Transaction hash
    - Blockchain identifier of the form `0000012345/0000032/0000001`
- We add a `fetchstatus` option to the `GET` `/operations/{operationId}` calls\
    - Makes an in-line `GET` to FFTM/EVMConnect for the latest full body of the `/transactions/{txId}` FFTM API
    - This is returned under a `status` field in the response to the `/operations/{operationId}` call
    - The UI enables `fetchstatus` by default
- In the case that the status of the backend during such a `GET` is resolved to `Succeeded`/`Failed`, but the status of the operation is still `Pending` - we dispatch an update of the FF Core status, as if we'd just received the websocket event

matthew1001:
Closes #1108 

See related UI PR https://github.com/hyperledger/firefly-ui/pull/208 which uses this query parameter to populate a `Detail` section of the operation slide out.

Signed-off-by: Matthew Whitehead <matthew.whitehead@kaleido.io>


internal/blockchain/ethereum/ethereum.go
```diff
@@ -947,3 +947,66 @@ func (e *Ethereum) GetAndConvertDeprecatedContractConfig(ctx context.Context) (l
...
+	if operation.Status == core.OpStatusPending {
+		// handleReceipt() expects a single object with the fields from both operation.Output
+		// and the blockchain connector's transaction detail, so we merge the two
+		// before passing to handleReceipt()
+		mergedObj, err := mergeObjects(statusResponse, operation.Output)
+		if err != nil {
+			// Log a warning and return the output on its own
+			log.L(ctx).Warnf("Failed to add detail to transaction status: %v", err)
+		}
+		e.handleReceipt(ctx, mergedObj)
```
peterbroadhurst, line 1009: Would like to understand this a bit better.
Is this something that's historical from another part of the code?

... digging a bit myself to find out

peterbroadhurst: No, the `handleReceipt` function seems to directly take the output from a WebSocket notification as far as I can tell:
https://github.com/hyperledger/firefly/blob/8f380025acc10713a437a0400973df5ac393b59f/internal/blockchain/ethereum/ethereum.go#L348-L371

Those notifications are of a different form to just the `/transaction/{id}` data. They contain a `headers` section (which is inherited from the EthConnect API that came before).
It's a very simple wrapper as so:
https://github.com/hyperledger/firefly-transaction-manager/blob/affb781dc00216703b2b3945cff556b8797e0f93/pkg/fftm/policyloop.go#L288-L305

peterbroadhurst (changes requested):
Sorry my comments are a bit long her @matthew1001 - if you'd like to discuss let me know.

Think the net is:
- I don't think there's any need for merging operations and status in any case
- We should only do the special "turn status query into event" recovery process, in the one specific case that we know it's required
- We should use `JSONObject` for consistency in our model with other structures like this, rather than `interface{}`.
- While I think it might be valid to defer the coding on the Fabric side, I don't see any actual difference in how it works there - so the comments should be a TODO I think.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -947,3 +947,66 @@ func (e *Ethereum) GetAndConvertDeprecatedContractConfig(ctx context.Context) (l
...
+
+func (e *Ethereum) GetTransactionStatus(ctx context.Context, operation *core.Operation) (interface{}, error) {
+	txnID := (&core.PreparedOperation{ID: operation.ID, Namespace: operation.Namespace}).NamespacedIDString()
+	transactionRequestPath := fmt.Sprintf("/transactions/%s", txnID)
+	client := e.client
+	if client == nil {
+		client = e.client
+	}
+	var resErr ethError
+	var statusResponse interface{}
```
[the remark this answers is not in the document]
matthew1001: Yeah that's a good point. I've updated it to use `fftypes.JSONObject`

internal/blockchain/ethereum/ethereum.go
matthew1001: I've done some refactoring under the new commit https://github.com/hyperledger/firefly/pull/1111/commits/547432920ab2e39aad1a281f9f917a73d0c05e24 which I think addresses the other comments. Let me know if that looks more like what you were expecting re. what we pass to `handleReceipt()`

Might be worth a chat about https://github.com/hyperledger/firefly/pull/1111#discussion_r1039628753 to discuss in a bit more detail?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -947,3 +931,44 @@ func (e *Ethereum) GetAndConvertDeprecatedContractConfig(ctx context.Context) (l
...
+	if txStatus != "" {
+		// If the status has changed, mock up blockchain receipt as if we'd received it
+		// as a web socket notification
+		if operation.Status == core.OpStatusPending && txStatus != ethTxStatusPending {
+			receipt := &common.BlockchainReceiptNotification{
+				Headers: common.BlockchainReceiptHeaders{
+					ReceiptID: headers.GetString("id"),
+					ReplyType: headers.GetString("type")},
+				TxHash:  statusResponse.GetString("transactionId"),
+				Message: statusResponse.GetString("errorMessage")}
```
peterbroadhurst, line 965: Should I see `ProtocolID` here?

internal/blockchain/fabric/fabric.go
```diff
@@ -921,3 +921,8 @@ func (f *Fabric) GetAndConvertDeprecatedContractConfig(ctx context.Context) (loc
 	})
 	return location, fromBlock, err
 }
+
+func (f *Fabric) GetTransactionStatus(ctx context.Context, operation *core.Operation) (interface{}, error) {
```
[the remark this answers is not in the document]
peterbroadhurst: Just to close out on this one @matthew1001 - are we going to leave this as a TODO in this PR?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -947,3 +931,52 @@ func (e *Ethereum) GetAndConvertDeprecatedContractConfig(ctx context.Context) (l
...
+		// If the status has changed, mock up blockchain receipt as if we'd received it
+		// as a web socket notification
+		if operation.Status == core.OpStatusPending && txStatus != ethTxStatusPending {
+			receipt := &common.BlockchainReceiptNotification{
+				Headers: common.BlockchainReceiptHeaders{
+					ReceiptID: headers.GetString("id"),
+					ReplyType: headers.GetString("type")},
+				TxHash:  statusResponse.GetString("transactionId"),
+				Message: statusResponse.GetString("errorMessage"),
+				ProtocolID: common.ProtocolIDForReceipt(
```
peterbroadhurst, line 968: I'm not sure we want a split responsibility on the generation of this.
It seems on the normal notification code path we just take this from the payload supplied by FFTM (which I think is the only connector implementation that correctly populates it today), whereas here we construct it

internal/blockchain/ethereum/ethereum.go
```diff
@@ -947,3 +947,66 @@ func (e *Ethereum) GetAndConvertDeprecatedContractConfig(ctx context.Context) (l
...
+	res, err := client.R().
+		SetContext(ctx).
+		SetError(&resErr).
+		SetResult(&statusResponse).
+		Get(transactionRequestPath)
+	if err != nil || !res.IsSuccess() {
+		return nil, wrapError(ctx, &resErr, res, err)
+	}
+
+	if operation.Status == core.OpStatusPending {
```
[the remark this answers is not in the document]
matthew1001: Done.

internal/blockchain/fabric/fabric.go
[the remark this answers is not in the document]
matthew1001: Querying transaction status when `fetchstatus` is set has now been implemented under the latest commit. The edge-case enhancement of marking an operation as complete if a WS notification is missed is going to be left as a `TODO`

internal/blockchain/ethereum/ethereum.go
matthew1001: Done.

peterbroadhurst (approved):
👍

PR #1122, opened 2022-12-19
Add support for passing custom HTTP headers through

nguyer:
Depends on https://github.com/hyperledger/firefly-common/pull/41

Adds support for setting a list of HTTP headers to put on the request Context. If an FFResty client is _also_ to pass headers through, it will set all the same headers on any request that it makes.

Example config:

```yaml
api:
  address: 0.0.0.0
  port: 5000
  publicURL: http://127.0.0.1:5000
  passthroughHeaders:
    - X-SuperCoolCustomHeader
plugins:
  blockchain:
    - ethereum:
        ethconnect:
          url: http://127.0.0.1:5102
          passthroughHeadersEnabled: true
```

This would configure FireFly to pass a header named `X-SuperCoolCustomHeader` through from an HTTP request all the way to the Ethereum blockchain connector.


peterbroadhurst (approved):
👍

PR #1123, opened 2022-12-28
Add ability to bind a contract interface to a token pool

awrichar:
Part of [FIR-16](https://github.com/hyperledger/firefly-fir/pull/16).
Depends on https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/106.

_Discussion about the overall architecture/usefulness of this proposed feature should be directed to [FIR-16](https://github.com/hyperledger/firefly-fir/pull/16), while discussion here should focus on specific technical details of the current proposed approach._

This adds the ability for FireFly to provide ABI details to a token connector, making it easier for the token connector to know the shape of the underlying contract it is dealing with (vs. having to probe/guess at the methods available).

**Notes**
* This feature is only useful to users who have significant knowledge of the underlying token contract that sits on the other side of their token connector.
* It's likely to be most useful when creating a pool against a pre-deployed contract - ie when passing `config.address`. Worth mentioning that FireFly doesn't really "know" you can create pools against pre-deployed contracts (as `config` is opaque), but it's up to the user to understand the combination of features that are useful here.
* Theoretically you could create a factory contract that deploys custom ERC20/ERC721 instances, and then you'd need to pass `interface` on every pool creation even while relying on the factory sitting on the other side of the token connector to do the deploy. Seems like a less useful combo, but calling it out here for completeness.


internal/tokens/fftokens/fftokens.go
```diff
@@ -151,6 +153,17 @@ type activatePool struct {
 	RequestID   string             `json:"requestId,omitempty"`
 }
 
+type tokenInterface struct {
+	Format  core.TokenInterfaceFormat `json:"format"`
+	ABI     []*abi.Entry              `json:"abi,omitempty"`
+	Methods []*fftypes.FFIMethod      `json:"methods,omitempty"`
```
peterbroadhurst, line 159: I think `ffi` would be less confusing for the name of this parameter if possible.

As on the other end of the token connector where it talks to EVMConnect/EthConnect etc. the `methods` array is going to represent an array of ABI-style methods. Also where `abi` is used above, I think it's _actually_ an array of ABI methods (not the whole ABI).

internal/tokens/fftokens/fftokens.go
```diff
@@ -643,13 +662,68 @@ func (ft *FFTokens) ActivateTokenPool(ctx context.Context, nsOpID string, pool *
 	return false, nil
 }
 
-func (ft *FFTokens) MintTokens(ctx context.Context, nsOpID string, poolLocator string, mint *core.TokenTransfer) error {
+func (ft *FFTokens) prepareABI(ctx context.Context, methods []*fftypes.FFIMethod) ([]*abi.Entry, error) {
+	abiMethods := make([]*abi.Entry, len(methods))
```
peterbroadhurst, line 666: This line confirms for me the comment earlier - that the `abi` field of the payload passed to the Token connector is _not_ the full ABI, but rather the array of ABI-formatted Methods.

... all to say, I feel quite strongly we need to rename `methods` to `ffi` in the payload to the token connector (or rename _both_ to `abiMethods`/`ffiMethods`)

peterbroadhurst (changes requested):
One change requested on the FF <-> FFTokens API interface that I got stuck on, but looks great!

internal/tokens/fftokens/fftokens.go
awrichar: Or should it be called `methods` in both cases (and the `format` param tells you how to decode it)?

awrichar: Since it's not _really_ an `abi` or `ffi`, it's always a list of methods encoded in `abi` or `ffi` format.

peterbroadhurst (approved):
👍  - leaving merge coordination with you due to the pre-reqs

PR #1132, opened 2023-01-06
Add DELETE method for data API

nguyer:
Depends on: https://github.com/hyperledger/firefly-dataexchange-https/pull/73


db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
```diff
@@ -0,0 +1,25 @@
...
+  SELECT DISTINCT data.namespace, data.id, data.blob_hash, blobs.payload_ref, blobs.created, blobs.peer, blobs.size
+  FROM data
+  LEFT JOIN blobs ON blobs.hash = data.blob_hash
+  WHERE data.blob_hash IS NOT NULL
+;
+DROP INDEX blobs_hash;
+DROP TABLE blobs;
+ALTER TABLE temp_blobs RENAME TO blobs; 
+CREATE INDEX blobs_hash_data_id ON blobs(hash, data_id);
+CREATE INDEX blobs_payload_ref ON blobs(payload_ref);
```
peterbroadhurst, line 23: Note for self: Confirm why we need this reverse lookup index

db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
```diff
@@ -0,0 +1,25 @@
...
+INSERT INTO temp_blobs (namespace, data_id, hash, payload_ref, created, peer, size)
+  SELECT DISTINCT data.namespace, data.id, data.blob_hash, blobs.payload_ref, blobs.created, blobs.peer, blobs.size
+  FROM data
+  LEFT JOIN blobs ON blobs.hash = data.blob_hash
+  WHERE data.blob_hash IS NOT NULL
+;
+DROP INDEX blobs_hash;
+DROP TABLE blobs;
+ALTER TABLE temp_blobs RENAME TO blobs; 
+CREATE INDEX blobs_hash_data_id ON blobs(hash, data_id);
```
peterbroadhurst, line 22: Should this just be `namespace, data_id`? Not sure the `hash` adds value, and particularly not first as `data_id` is the thing we expect to be unique in most practical cases.

internal/apiserver/route_delete_data.go
```diff
@@ -0,0 +1,45 @@
...
+	Name:   "deleteData",
+	Path:   "data/{dataid}",
+	Method: http.MethodDelete,
+	PathParams: []*ffapi.PathParam{
+		{Name: "dataid", Description: coremsgs.APIParamsDataID},
+	},
+	QueryParams:     nil,
+	Description:     coremsgs.APIEndpointsGetDataByID,
+	JSONInputValue:  nil,
+	JSONOutputValue: func() interface{} { return &core.Data{} },
```
peterbroadhurst, line 37: Should be `nil` for `204`

db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
peterbroadhurst: ok - this is for a compatibility check during deletion, that we do not delete a payload ref in DX if it's referred to by another data ID.

internal/events/dx_callbacks.go
```diff
@@ -213,11 +213,13 @@ func (em *eventManager) privateBlobReceived(dx dataexchange.Plugin, event dataex
 	// Dispatch to the blob receiver for efficient batch DB operations
 	em.blobReceiver.blobReceived(em.ctx, &blobNotification{
 		blob: &core.Blob{
+			Namespace:  br.Namespace,
 			Peer:       br.PeerID,
 			PayloadRef: br.PayloadRef,
 			Hash:       &br.Hash,
 			Size:       br.Size,
 			Created:    fftypes.Now(),
+			DataID:     fftypes.MustParseUUID(br.DataID),
```
peterbroadhurst, line 222: Bit worried about the `Must` here. Does this introduce a case where a DX plugin with a malformed `dataId` in the event notification could panic the server?

peterbroadhurst (changes requested):
Looks great @nguyer!

A couple of minor changes requested.

db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
nguyer: Correct 🙂

db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
nguyer: I thought we had discussed this and decided that we should always be looking up blobs by both their hash and data ID to prevent spoofing of different data?

internal/events/dx_callbacks.go
nguyer: Good point. I'll make this a bit safer.

db/migrations/postgres/000105_add_namespace_and_data_id_to_blobs.up.sql
peterbroadhurst: Yes, but I don't think that changes my suggestion on the index that's most efficient.

I believe the lookup is:
`{ ns: "mynamespace", data_id: "id12345", hash: "abce1234" }`

I assert the most efficient index to support this lookup is:
`namespace`, `data_id`

- Leaving off `namespace` feels wrong
- Adding in a 3rd dimension of `hash` seems unnecessary (as the scan length should be trivially small as we're covering a very minor edge case there)

internal/data/data_manager.go
```diff
@@ -539,3 +540,42 @@ func (dm *dataManager) WriteNewMessage(ctx context.Context, newMsg *NewMessage)
...
+	data, err := dm.database.GetDataByID(ctx, dm.namespace.Name, id, false)
+	if err != nil {
+		return err
+	}
+	if data == nil {
+		return i18n.NewError(ctx, coremsgs.Msg404NoResult)
+	}
+	if data.Blob != nil && data.Blob.Hash != nil {
+
+		blob, err := dm.database.GetBlob(ctx, dm.namespace.Name, data.ID, data.Blob.Hash)
```
peterbroadhurst, line 559: I believe we determined there could be multiple blobs at this point @nguyer 

e.g. I was expecting to see:

```go
blobs, err := dm.database.GetBlobs(ctx, db.namespace.Name, fb.And(fb.Eq("data_id",data.ID), fb.Eq(data.Blob.Hash))
```

I'm not sure it even makes sense to have a `GetBlob` database-level action, because the only unique key for a record is the `Sequence` key (which we correctly use in delete).

peterbroadhurst: The specific case I'm thinking of, is when you receive the same data, with the same ID/Hash from multiple location in your business next, in multiple different messages.

pkg/dataexchange/plugin.go
```diff
@@ -89,6 +89,9 @@ type Plugin interface {
 	// DownloadBlob streams a received blob out of storage
 	DownloadBlob(ctx context.Context, payloadRef string) (content io.ReadCloser, err error)
 
+	// DeleteBlob streams a received blob out of storage
```
awrichar, line 92: Comment is wrong

docs/swagger/swagger.yaml
```diff
@@ -5055,6 +5055,32 @@ paths:
       tags:
       - Default Namespace
   /data/{dataid}:
+    delete:
+      description: Gets a data item by its ID, including metadata about this item
```
awrichar, line 5059: This looks wrong

internal/apiserver/route_delete_data.go
```diff
@@ -0,0 +1,44 @@
...
+
+var deleteData = &ffapi.Route{
+	Name:   "deleteData",
+	Path:   "data/{dataid}",
+	Method: http.MethodDelete,
+	PathParams: []*ffapi.PathParam{
+		{Name: "dataid", Description: coremsgs.APIParamsDataID},
+	},
+	QueryParams:     nil,
+	Description:     coremsgs.APIEndpointsGetDataByID,
```
awrichar, line 34: This is the wrong description

pkg/database/plugin.go
```diff
@@ -334,8 +337,8 @@ type iBlobCollection interface {
 	// InsertBlobs performs a batch insert of blobs assured to be new records - fails if they already exist, so caller can fall back to upsert individually
 	InsertBlobs(ctx context.Context, blobs []*core.Blob) (err error)
 
-	// GetBlobMatchingHash - lookup first blob batching a hash
-	GetBlobMatchingHash(ctx context.Context, hash *fftypes.Bytes32) (message *core.Blob, err error)
+	// GetBlob - lookup the blob for a data ID
+	GetBlob(ctx context.Context, namespace string, dataID *fftypes.UUID, hash *fftypes.Bytes32) (message *core.Blob, err error)
 
 	// GetBlobs - get blobs
 	GetBlobs(ctx context.Context, filter ffapi.Filter) (message []*core.Blob, res *ffapi.FilterResult, err error)
```
awrichar, line 344: The fact that we still have a cross-namespace `GetBlobs` is anomalous. If it's still needed for compatibility, maybe a comment here is worthwhile. In general cross-namespace lookups are unreliable, as different namespaces may actually use different databases.

internal/data/data_manager.go
```diff
@@ -539,3 +540,44 @@ func (dm *dataManager) WriteNewMessage(ctx context.Context, newMsg *NewMessage)
...
+	data, err := dm.database.GetDataByID(ctx, dm.namespace.Name, id, false)
+	if err != nil {
+		return err
+	}
+	if data == nil {
+		return i18n.NewError(ctx, coremsgs.Msg404NoResult)
+	}
+	if data.Blob != nil && data.Blob.Hash != nil {
+		fb := database.BlobQueryFactory.NewFilter(context.Background())
+		blobs, _, err := dm.database.GetBlobs(ctx, fb.And(fb.Eq("data_id", data.ID), fb.Eq("hash", data.Blob.Hash)))
```
awrichar, line 559: Are we intentionally querying across namespaces here? I would expect it to be scoped to the namespace managed by this instance of `dataManager`.

awrichar (commented):
Overall looks very good. Noted a few spelling items and some questions around namespace scoping in the new landscape.

internal/data/data_manager.go
peterbroadhurst: Thanks for working that all through 👍

peterbroadhurst (approved):
🗄️ 
🔥 
🗑️

PR #1133, opened 2023-01-08
Fix coverage gap in FFIErrors, and make DB interface consistent

peterbroadhurst:
There was no coverage of the DB interface for FFI Errors, and oddly there was a difference in the query interface to all other DB packages. Unsure why that was, so I made it consistent.


shorsher (approved):
💯

PR #1141, opened 2023-01-12
Add option to invoke the addressResolver on every call

peterbroadhurst:
The address resolver is a very useful plug point for FireFly with EVM based chains that have sophisticated signing infrastructures.

We have a requirement for the interface to be called on every API call that has a `key` input on the Hyperledger FireFly API, regardless of whether the input string conforms to an `0x` address already, and regardless of how many times it is called.

This PR introduces a single new config option `alwaysResolve` to do this.

I added a second commit on this PR, after doing quite a detailed review of every place we were calling the address
resolver. I found the terminology was still confusing (know we've had a few changes) and make a tweak.

I also made the following changes:
- Removed two places within the `ethereum` connector it was using the `NormalizeSigningKey` function, to just format an ethereum address. That is now fixed
- I removed the additional cache on address resolution that was in the `Identity Manager`, because this would have got int the way of the removal of the cache in the blockchain layer. Also, _both_ Ethereum and Fabric already had caches for the expensive logic that wasn't just string formatting, so it seemed completely redundant.
- Fixed an intermittent coverage gap in Webhooks, fixing also the fact we were logging an error when there wasn't one, and removing a return of an error from `doDelivery` that was always `nil`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,7 +33,6 @@ linters:
   disable:
     - structcheck
   enable:
-    - deadcode
     - depguard
     - dogsled
     - errcheck
@@ -56,4 +55,3 @@ linters:
     - typecheck
     - unconvert
     - unused
-    - varcheck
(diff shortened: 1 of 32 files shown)


nguyer:
I ran the tests on this branch locally and have verified that there is no longer a gap in coverage, despite what codecov is reporting.

nguyer (approved):
Looks good. Makes sense to me.

PR #1144, opened 2023-01-16
Verify the default key in all cases

peterbroadhurst:
In the specific case of the `multipartyRootVerifier` we were not calling the `ResolveInputSigningKey` function on the Blockchain plugin, meaning a corner case where #1141 was not causing an invoke of the plug point.


nguyer (approved):
Looks good. Just waiting for tests to pass then it can be merged.

PR #1152, opened 2023-01-17
public vs. permissioned page

aznrayizzle:
added back public vs. permissioned page


awrichar (approved):

awrichar:
Looks like you need to add signoff for DCO.

This looks fine to me, but might be worth a check with @peterbroadhurst to confirm if this is close enough to the current committed architecture. This doc looks like it was written while FFTM was still evolving, so could need some updating.

Issue #1155, opened 2023-01-20
Webhook-subscription to a listener for [blockchain_event] panics/crashes the respective FF-node, when processing the event


Rabax55:
# Bug Report

** FireFly-version ***
  "Version": "1.1.2",
  "Commit": "8b462fff760e1c7cc1d11c3ddb34638c094b0f2e",
  "Date": "2022-11-17T21:16:50Z",
  "License": "Apache-2.0"

** Current behavior: **
A webhook-subscription to a listener for [blockchain_event] panics/crashes the respective FF-node, when processing the event

** Expected behavior: **
The webhook-subscription should relay the [blockchain_event] to it's webhook-endpoint 

** Steps to reproduce: **
1. Create a new FF-stack - i called it "test_webhook" 
   -> 'ff init test_webhook 2 --prompt-names' (testcase 2 members name/node being 'org_00/n_org_00' and 'org_01/n_org_01')

2. Deploy a smart contract w/ a function emitting an event 
   -> for the test the notorious SimpleStorage.sol w/ event 'Changed' was used...

3. Define the smart contract ABI and Register contract API to the stack
   -> done via FF-Sandbox on node n_org_00
   -> Name of the contract API: 'SimpleStorage_100'

4. Register a contract listener  
   -> done via FF-Sandbox on node n_org_01
   -> listener name: 'SimpleStorage_100' (same for topic: and name:), 
      Event: 'Changed',
      firstEvent: 'oldest'

5. Create a subscription for the listener 'SimpleStorage_100', on node n_org_01    
   -> done via FF-System-API, like so:
      POST http://127.0.0.1:5001/api/v1/namespaces/default/subscriptions - w/ body:
      {
        "filter": {
          "blockchainevent": {
            "listener": "ce44e800-0938-469a-a7f7-ccf3e8d38974",
            "name": "Changed" 
          },
          "events": ".*",
          "message": {
            "author": ".*",
            "group": ".*",
            "tag": ".*"
          },
          "topic": ".*",
          "transaction": {
            "type": ".*"
          }
        },
        "name": "Org_01_TestWebhookFirefly",
        "namespace": "default",
        "options": {
          "fastack": true,
          "firstEvent": "oldest",
          "json": true,
          "readAhead": 65535,
          "reply": true,        
          "replytag": ".*",
          "replytx": ".*",
          "url": "http://<webhook-endpoint@>ngrok.io/<route>?<query-parm-1&<query-parm-2",
          "withData": false        
        },
[...]

PR #1156, opened 2023-01-24
Fix panic on webhooks without messages

nguyer:
Resolves https://github.com/hyperledger/firefly/issues/1155

This PR changes the behavior of webhook callbacks in several ways:

- If `reply` is specified on the subscription it no longer panics if there is no Message on the event (such as is the case with custom contract events)
- If `withData` is specified subscription, but no Message is on the event, the event payload itself will still be sent

This means that subscriptions can be setup with either the `reply` and `withData` fields set (or both) and still use the same subscription for custom contract events as well.

### DIFF
--- a/internal/events/webhooks/webhooks.go
+++ b/internal/events/webhooks/webhooks.go
@@ -211,18 +211,19 @@ func (wh *WebHooks) attemptRequest(sub *core.Subscription, event *core.EventDeli
 
 	if req.method == http.MethodPost || req.method == http.MethodPatch || req.method == http.MethodPut {
 		switch {
-		case !withData:
-			// We are just sending the event itself
-			req.r.SetBody(event)
 		case req.body != nil:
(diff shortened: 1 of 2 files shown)


awrichar (approved):

internal/events/webhooks/webhooks.go
```diff
@@ -211,18 +211,19 @@ func (wh *WebHooks) attemptRequest(sub *core.Subscription, event *core.EventDeli
...
 			// We might have been told to extract a body from the first data record
 			req.r.SetBody(req.body)
 		case len(allData) > 1:
 			// We've got an array of data to POST
 			req.r.SetBody(allData)
-		default:
-			// Otherwise just send the first object directly
+		case len(allData) == 1:
+			// Just send the first object directly
 			req.r.SetBody(firstData)
```
awrichar, line 222: In retrospect, Codecov is saying that this line is now missing coverage...

PR #1157, opened 2023-01-25
Add "attachment" field to message header

awrichar:
This field indicates another type of FireFly object that is to be correlated with the message. This replaces the proliferation of message "types" that were actually a duality, like "transfer_broadcast" and "transfer_private". These previous types are deprecated, but all changes are backwards compatible.

As is often the case, naming is the primary concern here. Is "attachment" the right word to use? It's fairly close to accurate IMO, although it runs the risk of being confused with the data payload of the message. Something around "correlation" crossed my mind as well, but we've used "correlator" on events and I didn't want to reuse.

### DIFF
--- a/docs/reference/types/message.md
+++ b/docs/reference/types/message.md
@@ -74,6 +74,7 @@ nav_order: 15
 | `cid` | The correlation ID of the message. Set this when a message is a response to another message | [`UUID`](simpletypes#uuid) |
 | `type` | The type of the message | `FFEnum`:<br/>`"definition"`<br/>`"broadcast"`<br/>`"private"`<br/>`"groupinit"`<br/>`"transfer_broadcast"`<br/>`"transfer_private"`<br/>`"approval_broadcast"`<br/>`"approval_private"` |
 | `txtype` | The type of transaction used to order/deliver this message | `FFEnum`:<br/>`"none"`<br/>`"unpinned"`<br/>`"batch_pin"`<br/>`"network_action"`<br/>`"token_pool"`<br/>`"token_transfer"`<br/>`"contract_deploy"`<br/>`"contract_invoke"`<br/>`"token_approval"`<br/>`"data_publish"` |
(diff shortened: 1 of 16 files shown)


peterbroadhurst:
I think to evaluate this, it would be good to net out the different types of attachments.

I believe all of the mentioned examples have a `Transaction` associated with them, but we're torn on this because we've got a batch transaction that is the thing that pins the batch, that this message is assembled into. I wonder if some thinking on that is worthwhile... whether we have anywhere on the API today that relies on the Message->Batch->Transaction relationship explicitly, that would preclude a Message->Transaction direct relationship to co-exist.

awrichar:
To spell out the technical purpose of this field (which today is served by the "special" message types) - the event aggregator, upon receiving one of these messages, must know to scan a different database table to do a reverse lookup for this message by hash. It must find that "other" record before considering the message confirmed.

Right now there are exactly 2 such lookups - the token transfers table and the token approvals table.

The line of thinking around transactions is interesting; I'll give it some thought as to whether there's a different direction there for these lookups...

awrichar:
Alright, I put some more thought into this. I was initially thinking it might intersect with [FIR-17](https://github.com/hyperledger/firefly-fir/pull/17), but I don't think it will. So the scope here is entirely confined to a cleanup of the existing two token operations that can be correlated with a message (token transfers and token approvals), with no real plans to add any others.

The objects involved look like this:
FireFly token transfer with message

Note that currently there is a mapping between a `message` and a `token transfer` here (in the form of a `messageHash` field on the transfer table). I think the comment above suggested looking at whether that is the best mapping, or whether something with the two `transaction` objects might be better. Options there:
1. If we were to add a transaction ID to `message`, I think the first choice would be propagating the ID of the `batch_pin` transaction. If we were to add a _second_ "related transaction ID", then we could also consider adding the ID of the `token_transfer` transaction. After all, both transactions need to complete to consider the message confirmed.
2. Mapping in the other direction does not make sense - that is, having the `token transfer` somehow point to the `batch_pin` transaction. The transfer is only attached to a single message, and the batch may have other messages in it. We also consider the `token transfer` to be confirmed regardless of whether the message is received or not.

PR #1162, opened 2023-01-26
Add "transactions" field to messages

awrichar:
Tracks the ID of the batch transaction and any other related transaction.

Really just trying this on for size right now to see if it makes sense as a direction. It's related to discussions on both https://github.com/hyperledger/firefly/pull/1157 and https://github.com/hyperledger/firefly-fir/pull/17.

This would give an easier way to look up the `batch_pin` transaction associated with each message (without having to join via the batch), and a way to look up the additional `token_transfer` or `token_approval` transaction associated with messages that were attached to a transfer or approval. It would also potentially support FIR-17 by providing a place to stash a pre-assigned transaction ID on the message, in the case that the transaction is created prior to batch assembly.

### DIFF
--- a/pkg/core/message.go
+++ b/pkg/core/message.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -86,6 +86,11 @@ type MessageHeader struct {
 	DataHash  *fftypes.Bytes32      `ffstruct:"MessageHeader" json:"datahash,omitempty" ffexcludeinput:"true"`
 }
 
+type MessageTransactions struct {
+	Batch   *fftypes.UUID `ffstruct:"MessageTransactions" json:"batch,omitempty" ffexcludeinput:"true"`
+	Related *fftypes.UUID `ffstruct:"MessageTransactions" json:"related,omitempty" ffexcludeinput:"true"`
+}
+
 // Message is the envelope by which coordinated data exchange can happen between parties in the network
 // Data is passed by reference in these messages, and a chain of hashes covering the data and the
 // details of the message, provides a verification against tampering.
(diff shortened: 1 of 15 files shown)


pkg/core/message.go
```diff
@@ -86,6 +86,11 @@ type MessageHeader struct {
 	DataHash  *fftypes.Bytes32      `ffstruct:"MessageHeader" json:"datahash,omitempty" ffexcludeinput:"true"`
 }
 
+type MessageTransactions struct {
+	Batch   *fftypes.UUID `ffstruct:"MessageTransactions" json:"batch,omitempty" ffexcludeinput:"true"`
+	Related *fftypes.UUID `ffstruct:"MessageTransactions" json:"related,omitempty" ffexcludeinput:"true"`
```
awrichar, line 91: I'm not in love with the naming here, but I did find that I had to use two distinct fields rather than a simple list of IDs (because we bulk-update the messages when the batch and batch transaction are assigned).

pkg/core/message.go
```diff
@@ -110,9 +116,10 @@ type Message struct {
 // Fields such as the state/confirmed do NOT transfer, as these are calculated individually by each member.
 func (m *Message) BatchMessage() *Message {
 	return &Message{
-		Header: m.Header,
-		Hash:   m.Hash,
-		Data:   m.Data,
+		Header:       m.Header,
+		Hash:         m.Hash,
+		Data:         m.Data,
+		Transactions: m.Transactions,
```
awrichar, line 122: I'm not sure this meets the constraints above for what should be transferred/immutable in the batch. It's not part of the hash, so there's really no verification on it. It's a convenient way to ensure the far side receives `transactions.related`... if they wanted to recompute it, it would have to be done by the event aggregator, by looking up the transfer/approval and copying the TX ID off of it and updating the message with it.

pkg/core/message.go
awrichar: Also worth noting, we do not _have_ to store both. We could opt for one or the other, as the paths that compute them are totally independent - just wanted to bottom out what it would look like in each case.

pkg/core/message.go
awrichar: Although... I guess I _could_ put the "related" transaction ID (the one for the token transfer or approval) into the message header. It would require some code shuffling, but it is computed in the same code path as the message sealing. The batch transaction ID would have to be outside the header, as it's computed much later, after sealing.

I guess maybe that would look something like:
```
{
  "header": {
    "txtype": "batch_pin",
    "txparent": {
      "type": "token_transfer",
      "id": "<id for transfer transaction>"
    }
  },
  "txid": "<id for batch pin transaction"
}
```

It's slightly weird in how it's organized, but I don't hate it. It would also solve #1157 in a different way, by adding a new field in the message header that tells you to correlate the message with a transfer or approval.

awrichar: I went ahead and explored this alternate route as #1163... so the saga continues.

PR #1163, opened 2023-01-26
Add "txid" and "txparent" to messages

awrichar:
Propagate the `batch_pin` transaction ID onto the message when the batch is assembled (sender) or when it is persisted (receiver), as a new `txid` field.

For token transfers and approvals that come with a message, store the `token_transfer` or `token_approval` transaction details on the message up-front, as a new `txparent` field.

Illustration of changes to the message object:
message with transfer

This does mean the allocation of the `token_transfer` or `token_approval` transaction ID happens earlier and outside a database transaction, but I don't see a problem with this.

This probably supersedes both #1157 and #1162, as I think it's a cleaner route toward solving that problem _and_ teeing up [FIR-17](https://github.com/hyperledger/firefly-fir/pull/17).

### DIFF
--- a/internal/events/aggregator.go
+++ b/internal/events/aggregator.go
@@ -534,14 +534,14 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 
 func needsTokenTransfer(msg *core.Message) bool {
 	return (msg.Header.TxParent != nil && msg.Header.TxParent.Type == core.TransactionTypeTokenTransfer) ||
-		msg.Header.Type == core.MessageTypeTransferBroadcast ||
-		msg.Header.Type == core.MessageTypeTransferPrivate
+		msg.Header.Type == core.MessageTypeDeprecatedTransferBroadcast ||
+		msg.Header.Type == core.MessageTypeDeprecatedTransferPrivate
 }
 
 func needsTokenApproval(msg *core.Message) bool {
 	return (msg.Header.TxParent != nil && msg.Header.TxParent.Type == core.TransactionTypeTokenApproval) ||
-		msg.Header.Type == core.MessageTypeApprovalBroadcast ||
-		msg.Header.Type == core.MessageTypeApprovalPrivate
+		msg.Header.Type == core.MessageTypeDeprecatedApprovalBroadcast ||
+		msg.Header.Type == core.MessageTypeDeprecatedApprovalPrivate
 }
 
 func (ag *aggregator) attemptMessageDispatch(ctx context.Context, msg *core.Message, data core.DataArray, tx *fftypes.UUID, state *batchState, pin *core.Pin) (newState core.MessageState, dispatched bool, err error) {
--- a/internal/events/aggregator.go
+++ b/internal/events/aggregator.go
@@ -532,6 +532,18 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 	return nil
 }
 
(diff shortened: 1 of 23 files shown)


pkg/core/message.go
```diff
@@ -110,9 +112,10 @@ type Message struct {
 // Fields such as the state/confirmed do NOT transfer, as these are calculated individually by each member.
 func (m *Message) BatchMessage() *Message {
 	return &Message{
-		Header: m.Header,
-		Hash:   m.Hash,
-		Data:   m.Data,
+		Header:   m.Header,
+		Hash:     m.Hash,
+		Data:     m.Data,
+		TxParent: m.TxParent, // TODO: this should really move inside Header
```
awrichar, line 118: Important TODO here - I think this should be possible, but requires some re-shuffling of logic when building the token transfer/approval, so that the transaction ID is computed before the message is sealed.

internal/events/aggregator.go
```diff
@@ -532,6 +532,18 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 	return nil
 }
 
+func needsTokenTransfer(msg *core.Message) bool {
+	return (msg.TxParent != nil && msg.TxParent.Type == core.TransactionTypeTokenTransfer) ||
+		msg.Header.Type == core.MessageTypeTransferBroadcast ||
```
awrichar, line 537: These "special" message types can be deprecated if we go this route, as `TxParent.Type` would be the new/authoritative way to identify an associated transfer/approval.

pkg/core/message.go
```diff
@@ -42,14 +42,14 @@ var (
...
-	// MessageTypeApprovalPrivate is a private message to accompany/annotate a token approval
-	MessageTypeApprovalPrivate = fftypes.FFEnumValue("messagetype", "approval_private")
+	// MessageTypeDeprecatedTransferBroadcast is deprecated - use MessageTypeBroadcast (and refer to TxParent.Type)
+	MessageTypeDeprecatedTransferBroadcast = fftypes.FFEnumValue("messagetype", "transfer_broadcast")
+	// MessageTypeDeprecatedTransferPrivate is deprecated - use MessageTypePrivate (and refer to TxParent.Type)
+	MessageTypeDeprecatedTransferPrivate = fftypes.FFEnumValue("messagetype", "transfer_private")
+	// MessageTypeDeprecatedApprovalBroadcast is deprecated - use MessageTypeBroadcast (and refer to TxParent.Type)
+	MessageTypeDeprecatedApprovalBroadcast = fftypes.FFEnumValue("messagetype", "approval_broadcast")
+	// MessageTypeDeprecatedApprovalPrivate is a deprecated - use MessageTypePrivate (and refer to TxParent.Type)
+	MessageTypeDeprecatedApprovalPrivate = fftypes.FFEnumValue("messagetype", "approval_private")
```
nguyer, line 52: Sorry I wasn't able to ascertain it from elsewhere in the diff, but what was the reasoning behind deprecating the existing message types? Is this a data migration issue?

nguyer (approved):

pkg/core/message.go
awrichar: Shouldn't be an issue. There are now two ways supported to mark a message as "this is a broadcast that will be accompanied by a token transfer":
* Send a message with type "transfer_broadcast"
* Send a message with type "broadcast" and txparent.type "token_transfer"

The latter is now preferred as the cleaner and more extensible route for the future, but the former is still supported.

Issue #1167, opened 2023-01-30
When using an address resolver, any HTTP error status codes are converted to `500`.


hosie:
When using an address resolver, any HTTP error status codes are converted to `500`.

Steps to reproduce
 - configure an address resolver for the blockchain plugin to point at an endpoint that will return a non 200 status code
e.g....
```yaml
  blockchain:
  - name: blockchain0
    type: ethereum
    ethereum:
      addressResolver: 
        urlTemplate: "https://httpbin.org/status/403"
      ethconnect:
        url: http://evmconnect_0:5008
        topic: "0"  
```
 - deploy a contract and create an API
 - invoke the api, passing any string as the `key` property in the JSON payload.
 e.g.
 ```
 curl -i -X 'POST' \
  'http://127.0.0.1:5000/api/v1/namespaces/default/apis/simple_storage/invoke/get' \
  -H 'accept: application/json' \
  -H 'Request-Timeout: 2m0s' \
  -H 'Content-Type: application/json' \
  -d '{
  "input": {},
  "options": {},
  "key": "foo"
}'
 ```

You will see that the response code is always `500` but the body of the response mentions the response code from the address resolver...
```
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Vary: Origin
Date: Mon, 30 Jan 2023 18:53:41 GMT
Content-Length: 76

{"error":"FF10340: Failed to resolve signing key string 'foo' [403]: null"}
```

Expected behaviour is for the HTTP status code from firefly to be the same as that from the address resolver.

PR #1170, opened 2023-01-31
propagate status code from resolver

hosie:
fixes https://github.com/hyperledger/firefly/issues/1167

### DIFF
--- a/internal/coremsgs/en_error_messages.go
+++ b/internal/coremsgs/en_error_messages.go
@@ -190,7 +190,6 @@ var (
 	MsgGoTemplateCompileFailed            = ffe("FF10337", "Go template compilation for '%s' failed: %s", 500)
 	MsgGoTemplateExecuteFailed            = ffe("FF10338", "Go template execution for '%s' failed: %s", 500)
 	MsgAddressResolveFailed               = ffe("FF10339", "Failed to resolve signing key string '%s': %s", 500)
-	MsgAddressResolveBadStatus            = ffe("FF10340", "Failed to resolve signing key string '%s' [%d]: %s", 500)
 	MsgAddressResolveBadResData           = ffe("FF10341", "Failed to resolve signing key string '%s' - invalid address returned '%s': %s", 500)
 	MsgDXNotInitialized                   = ffe("FF10342", "Data exchange is initializing")
 	MsgGroupRequired                      = ffe("FF10344", "Group must be set", 400)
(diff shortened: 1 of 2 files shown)


internal/coremsgs/en_error_messages.go
```diff
@@ -279,3 +278,8 @@ var (
 	MsgDeprecatedResetWithAutoReload      = ffe("FF10438", "The deprecated reset API cannot be used when dynamic config reload is enabled", 409)
 	MsgConfigArrayVsRawConfigMismatch     = ffe("FF10439", "Error processing configuration - mismatch between raw and processed array lengths")
 )
+
+// function versions of error factories that allow status code to be parameterized
+var MsgAddressResolveBadStatus = func(status int) i18n.ErrorMessageKey {
+	return ffe("FF10340", "Failed to resolve signing key string '%s' [%d]: %s", status)
+}
```
nguyer, line 285: Clever, but unfortunately I think this will cause problems. The `ffe` function doesn't build a specific instance of an error. It registers a specific error/translation code within the runtime. There is code that specifically checks to make sure the same code isn't already registered (helpful when making code changes and accidentally re-using an existing code).

I believe the code as written above will cause this function to panic:
https://github.com/hyperledger/firefly-common/blob/main/pkg/i18n/messages.go#L93-L95

Off the top of my head, I'm not sure of viable alternatives other than listing out the specific error codes we want to return, and then mapping to those. Or changing the way the `FFE/FFM` function works itself.

nguyer (changes requested):

internal/coremsgs/en_error_messages.go
hosie: good spot.  I just tested with a resolver URL of 
```
      addressResolver: 
        urlTemplate: "https://httpbin.org/status/{{.Key}}"
```
so that I could force different status codes and, as you predicted...
```
2023/01/31 18:50:25 http: panic serving 172.27.0.1:60446: Message ID FF10340 re-used
goroutine 754 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1825 +0xb8
panic({0xd3d740, 0x400045d0e0})
	/usr/local/go/src/runtime/panic.go:844 +0x260
github.com/hyperledger/firefly-common/pkg/i18n.FFM({0x2e08?, 0xac?, {0x0?, 0x0?}}, {0xe8781d, 0x7}, {0xebd096, 0x32})
	/go/pkg/mod/github.com/hyperledger/firefly-common@v1.2.1/pkg/i18n/messages.go:94 +0x284

```

hosie:
Closing this while we find a better solution

PR #1182, opened 2023-02-08
Allow a message to contain two data items with the same hash

awrichar:
It's possible to have two data items with the same ID but different hashes. It should be possible to send both data items in a single message - I don't see any reason for this secondary check on the hash.

As a larger issue, it seems problematic that we can have two different data items with the same hash. The hash semantics should probably be revisited, to determine if other fields like the ID should be incorporated into the hash.


nguyer (approved):
Looks good to me :shipit:

PR #1184, opened 2023-02-13
Fix path in Fabric test-network tutorial

nguyer:
While re-running this tutorial I noticed that one of the paths to a key seemed incorrect. I have successfully run through it using the updated path in this PR.


matthew1001:
I reviewed the changes but I'm not a required approver so someone else will have to approve it before merging.

awrichar (approved):

alcemirsantos:
This fix is not live at the tutorial yet.   I came here to do exactly the same PR.

PR #1203, opened 2023-03-06
[tokens] Only Use broadcastName for Multi-Party Namespaces

hfuss:
We saw if `broadcastName` was set on a tokens plugin, token pools would fail to create within gateway namespaces:
```
[2023-03-01T20:25:10.568Z] ERROR Failed to activate token pool 'f2de67bb-5adf-4b98-ac06-7d25a19805ad': FF10272: Unknown tokens plugin 'erc20-erc721' ns=cok pid=1 role=event-manager
[2023-03-01T20:25:10.568Z] DEBUG fftokens updating operation cok:fb435cab-a2a5-4d1a-af28-f0bf84840c0b status=Failed error=FF10272: Unknown tokens plugin 'erc20-erc721' ns=cok pid=1
[2023-03-01T20:25:10.568Z] DEBUG SQL-> begin dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.568Z] DEBUG SQL<- begin dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.569Z] DEBUG SQL-> update operations dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.569Z] DEBUG SQL<- update operations affected=1 dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.569Z] DEBUG SQL-> lock cok dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.570Z] DEBUG SQL<- lock cok dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.570Z] DEBUG SQL-> insert events dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.570Z] DEBUG SQL<- insert events sequences=[9] dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.570Z] DEBUG SQL-> commit dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.573Z] DEBUG SQL<- commit dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.573Z]  INFO Emitted token_pool_op_failed event 47472d4f-70f7-4f3e-827d-a3e70c76cf72 for cok:fb435cab-a2a5-4d1a-af28-f0bf84840c0b (correlator=f2de67bb-5adf-4b98-ac06-7d25a19805ad,topic=f2de67bb-5adf-4b98-ac06-7d25a19805ad) dbtx=N1fXw8pi ns=cok pid=1
[2023-03-01T20:25:10.573Z]  INFO Inflight request 'f2de67bb-5adf-4b98-ac06-7d25a19805ad' resolved with timeout after 2068.43ms ns=cok pid=1 role=sync-async-bridge
[2023-03-01T20:25:10.573Z]  INFO <-- POST /api/v1/namespaces/cok/tokens/pools [400] (2071.90ms): FF10272: Unknown tokens plugin 'erc20-erc721' httpreq=hzG60xJW pid=1 req=Rd7exmVy
```

This is bc the definition handler doesn't know how to map the remote name back to the local name within gateway namespaces: https://github.com/hyperledger/firefly/blob/3b728534d50c0cee6ed27890114b5a4c44fcec5c/internal/definitions/handler_tokenpool.go#L36-L42

This fixes that by only using the `broadcastName` for multi-party namespaces when sending the token pool definition.

### DIFF
--- a/internal/definitions/handler_tokenpool.go
+++ b/internal/definitions/handler_tokenpool.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
--- a/internal/definitions/handler_tokenpool.go
+++ b/internal/definitions/handler_tokenpool.go
@@ -52,7 +52,7 @@ func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, stat
 
 	pool.Namespace = dh.namespace.Name
 	if err := pool.Validate(ctx); err != nil {
-		return HandlerResult{Action: ActionReject, CustomCorrelator: correlator}, i18n.NewError(ctx, coremsgs.MsgDefRejectedValidateFail, "token pool", pool.ID, err)
+		return HandlerResult{Action: ActionReject, CustomCorrelator: correlator}, i18n.WrapError(ctx, err, coremsgs.MsgDefRejectedValidateFail, "token pool", pool.ID)
 	}
 
 	// Check if pool has already been confirmed on chain (and confirm the message if so)
--- a/internal/definitions/sender_tokenpool.go
+++ b/internal/definitions/sender_tokenpool.go
@@ -56,8 +56,7 @@ func (bm *definitionSender) DefineTokenPool(ctx context.Context, pool *core.Toke
 			if innerErr := errors.Unwrap(err); innerErr != nil {
 				return hr, innerErr
 			}
-			return hr, err
 		}
-		return hr, nil
+		return hr, err
 	})
 }
--- a/internal/definitions/sender_tokenpool.go
+++ b/internal/definitions/sender_tokenpool.go
@@ -18,7 +18,6 @@ package definitions
 
 import (
 	"context"
-
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
 	"github.com/hyperledger/firefly/internal/coremsgs"
@@ -26,15 +25,16 @@ import (
 )
 
 func (bm *definitionSender) DefineTokenPool(ctx context.Context, pool *core.TokenPoolAnnouncement, waitConfirm bool) error {
-	// Map token connector name -> broadcast name
-	if broadcastName, exists := bm.tokenBroadcastNames[pool.Pool.Connector]; exists {
-		pool.Pool.Connector = broadcastName
-	} else {
-		log.L(ctx).Infof("Could not find broadcast name for token connector: %s", pool.Pool.Connector)
-		return i18n.NewError(ctx, coremsgs.MsgInvalidConnectorName, broadcastName, "token")
-	}
 
 	if bm.multiparty {
+		// Map token connector name -> broadcast name
+		if broadcastName, exists := bm.tokenBroadcastNames[pool.Pool.Connector]; exists {
+			pool.Pool.Connector = broadcastName
+		} else {
+			log.L(ctx).Infof("Could not find broadcast name for token connector: %s", pool.Pool.Connector)
+			return i18n.NewError(ctx, coremsgs.MsgInvalidConnectorName, broadcastName, "token")
+		}
+
 		if err := pool.Pool.Validate(ctx); err != nil {
 			return err
 		}
--- a/internal/definitions/sender_tokenpool.go
+++ b/internal/definitions/sender_tokenpool.go
@@ -18,6 +18,7 @@ package definitions
 
 import (
 	"context"
+	"errors"
 
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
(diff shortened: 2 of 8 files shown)


awrichar:
Looks like some build/test is failing. Also note there is an E2E `TokensRemoteNameTestSuite` that runs in multiparty mode only - maybe we could run that in gateway mode too to catch this.

internal/definitions/sender_tokenpool_test.go
```diff
@@ -56,10 +56,10 @@ func TestBroadcastTokenPoolInvalid(t *testing.T) {
 	mdm.AssertExpectations(t)
 }
 
-func TestBroadcastTokenPoolInvalidNonMultiparty(t *testing.T) {
+func TestBroadcastTokenPoolInvalidMultiparty(t *testing.T) {
```
awrichar, line 59: Is this test now identical to the previous one?

hfuss: heh not quite - wasn't working in gateway mode bc the error checking changed. So I'll actually fix this test as part of adding new unit / E2E tests.

internal/definitions/sender_tokenpool.go
```diff
@@ -26,17 +26,18 @@ import (
...
+		if broadcastName, exists := bm.tokenBroadcastNames[pool.Pool.Connector]; exists {
+			pool.Pool.Connector = broadcastName
+		} else {
+			log.L(ctx).Infof("Could not find broadcast name for token connector: %s", pool.Pool.Connector)
+			return i18n.NewError(ctx, coremsgs.MsgInvalidConnectorName, broadcastName, "token")
+		}
+
 		if err := pool.Pool.Validate(ctx); err != nil {
-			return err
+			return i18n.NewError(ctx, coremsgs.MsgDefRejectedValidateFail, "token pool", pool.Pool.ID, err)
```
hfuss, line 40: I'm wrapping the validate error in a definition rejected to match what was done here: https://github.com/hyperledger/firefly/blob/3b728534d50c0cee6ed27890114b5a4c44fcec5c/internal/definitions/handler_tokenpool.go#L54-L56

Because gateway namespaces were returning `FF10403` whereas multi-party namespaces were returning `FF00140`.

Not sure if this is correct though since we're defining the token pool here... not handling a definition. Happy to undo this if need be.

awrichar: The validation on the "handler" side is the most important, as that will allow us to reject bad data (regardless of where it came from).

The validation on the "sender" side is just a convenience, to help users avoid putting in bad data. But I would not consider that a definition rejection.

When doing the "short-circuit" route for non-multiparty mode, we've tended to prefer keeping just the "handler" side validation since it doesn't make sense to validate twice. But that does mean we return "definition rejected" sometimes...

I guess I would say we should _not_ be wrapping errors here in "definition rejected". If anything, we should be doing the opposite, perhaps by enhancing the way that `fakeBatch` works in order to "unwrap" rejection errors in the gateway mode case.

internal/definitions/handler_tokenpool.go
```diff
@@ -52,7 +52,7 @@ func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, stat
 
 	pool.Namespace = dh.namespace.Name
 	if err := pool.Validate(ctx); err != nil {
-		return HandlerResult{Action: ActionReject, CustomCorrelator: correlator}, i18n.NewError(ctx, coremsgs.MsgDefRejectedValidateFail, "token pool", pool.ID, err)
+		return HandlerResult{Action: ActionReject, CustomCorrelator: correlator}, i18n.WrapError(ctx, err, coremsgs.MsgDefRejectedValidateFail, "token pool", pool.ID)
```
awrichar, line 55: I don't think this should change (unless it changes everywhere)... `MsgDefRejectedValidateFail` has 3 substitution params in the string (for the type, the ID, and the validation error)

awrichar: (I'm also OK with changing it everywhere if that seems to make the most sense... just need to adjust the error string and everywhere else it's used)

internal/definitions/sender_tokenpool.go
```diff
@@ -49,6 +51,13 @@ func (bm *definitionSender) DefineTokenPool(ctx context.Context, pool *core.Toke
 	}
 
 	return fakeBatch(ctx, func(ctx context.Context, state *core.BatchState) (HandlerResult, error) {
-		return bm.handler.handleTokenPoolDefinition(ctx, state, pool.Pool)
+		hr, err := bm.handler.handleTokenPoolDefinition(ctx, state, pool.Pool)
+		if err != nil {
+			if innerErr := errors.Unwrap(err); innerErr != nil {
+				return hr, innerErr
+			}
+			return hr, err
```
awrichar, line 59: nit: you could omit this line and just `return hr, err` on 61

awrichar (approved):
Looks good; thanks for going around the loop on this!

hfuss:
Code coverage is 100% now: https://app.codecov.io/github/hyperledger/firefly/commit/a4e8b779a6c2fcf117645d8f7207b719818a994c

PR #1201, opened 2023-03-03
Run Solidity tests on GitHub PRs

awrichar:

PR #1213, opened 2023-03-07
Allow pinning a message to any blockchain contract invoke

awrichar:
Implements [FIR-17](https://github.com/hyperledger/firefly-fir/pull/17) (see there for full description of functionality)

In a chain with ~~#1201 and~~ #1202

Required CLI changes: https://github.com/hyperledger/firefly-cli/pull/244
Required UI changes: https://github.com/hyperledger/firefly-ui/pull/214

Still outstanding:
- [x] Unit test coverage
- [x] Documentation


smart_contracts/fabric/custompin-sample/batchpin/batchpin.go
```diff
@@ -0,0 +1,63 @@
+package batchpin
```
awrichar, line 1: I ended up copying this file between the FireFly chaincode project and this one. I wanted to have a single shared copy of the file somewhere in this repo, but couldn't figure out a way to make that work with the Fabric chaincode packaging.

smart_contracts/fabric/custompin-sample/chaincode/contract.go
```diff
@@ -0,0 +1,29 @@
...
+func (s *SmartContract) MyCustomPin(ctx contractapi.TransactionContextInterface, data string) error {
+	event, err := batchpin.BuildEventFromString(ctx, data)
+	if err != nil {
+		return err
+	}
+	bytes, err := json.Marshal(event)
+	if err != nil {
+		return fmt.Errorf("failed to marshal event: %s", err)
+	}
+	return ctx.GetStub().SetEvent("BatchPin", bytes)
```
awrichar, line 28: I considered defining another way to embed a BatchPin event inside another arbitrary event (instead of requiring BatchPin to be the top level event). That is not implemented yet, but could be added (here or in the future).

internal/batch/batch_processor.go
```diff
@@ -403,7 +412,7 @@ func (bp *batchProcessor) initPayload(id *fftypes.UUID, flushWork []*batchWork)
 	return payload
 }
 
-func (bp *batchProcessor) getNextNonce(ctx context.Context, state *dispatchState, nonceKeyHash *fftypes.Bytes32, contextHash *fftypes.Bytes32) (int64, error) {
+func (bm *batchManager) getNextNonce(ctx context.Context, state *dispatchState, nonceKeyHash *fftypes.Bytes32, contextHash *fftypes.Bytes32) (int64, error) {
```
awrichar, line 415: I had to reparent a few methods from batchProcessor to batchManager in order to wire up the LoadContexts functionality externally... didn't move the methods to the other file in the interest of keeping the diff smaller, but I could move them if it seems confusing now.

internal/blockchain/common/common.go
```diff
@@ -147,13 +148,21 @@ func (cb *callbacks) BatchPinOrNetworkAction(ctx context.Context, nsOrAction str
...
-			log.L(ctx).Errorf("No handler found for blockchain batch pin on remote namespace '%s'", nsOrAction)
+			log.L(ctx).Errorf("No handler found for blockchain batch pin on network namespace '%s'", networkNamespace)
 			return nil
 		}
 		return cb.batchPinComplete(ctx, namespaces, batch, signingKey)
 	}
+	batch.TransactionType = core.TransactionTypeBatchPin
+	if strings.HasPrefix(params.NsOrAction, blockchain.FireFlyActionPrefix) {
+		typeName := params.NsOrAction[len(blockchain.FireFlyActionPrefix):]
+		if typeName == "contract_invoke_pin" {
```
awrichar, line 162: The "action" field from the event is used for many different things at this point. Here, if the action is "firefly:contract_invoke_pin", we interpret that as the transaction type (otherwise we fall back to the default transaction type of "batch_pin").

internal/blockchain/ethereum/ethereum.go
```diff
@@ -726,6 +756,17 @@ func (e *Ethereum) InvokeContract(ctx context.Context, nsOpID string, signingKey
 	if err != nil {
 		return err
 	}
+	if batch != nil {
+		if err := e.checkDataSupport(ctx, abi); err != nil {
+			return err
+		}
+		method, batchPin := e.buildBatchPinInput(ctx, 2, "", batch)
```
awrichar, line 763: Note that we don't know the shape of the batch pin contract the user is ultimately going to call from their contract on-chain. We assume "version 2" of the ABI arguments (which has 4 args, where "version 1" expected 5 args) - but if we ever add a different version of this pinBatch method, there may need to be a way to choose upfront how to have the args encoded.

internal/multiparty/manager.go
```diff
@@ -228,7 +228,34 @@ func (mm *multipartyManager) SubmitNetworkAction(ctx context.Context, signingKey
...
+	}), nil
+}
+
 func (mm *multipartyManager) SubmitBatchPin(ctx context.Context, batch *core.BatchPersisted, contexts []*fftypes.Bytes32, payloadRef string) error {
+	if batch.TX.Type == core.TransactionTypeContractInvokePin {
+		preparedOp, err := mm.prepareInvokeOperation(ctx, batch, contexts, payloadRef)
+		if err != nil {
+			return err
+		} else if preparedOp != nil {
+			_, err = mm.operations.RunOperation(ctx, preparedOp)
```
awrichar, line 253: Biggest concern with the new functionality: since this is going to trigger a custom blockchain operation, there are a hundred ways it could fail. We're in an infinite retry loop here, and the assumption has always been that the blockchain piece cannot fail (which is pretty reasonable for the normal, very simple `pinBatch` contract).

If we have a fundamentally bad invoke here (wrong method, wrong params, wrong permissions, etc) then this processor will block forever with infinite retries. It will block any further batches of the same type from this author from being assembled.

In the case of private messages, the message has also claimed a nonce at this point, which means if the message never successfully gets pinned to the chain, that private group topic is also blocked.

internal/txcommon/contract_inputs.go
```diff
@@ -0,0 +1,56 @@
...
+func RetrieveBlockchainInvokeInputs(ctx context.Context, op *core.Operation) (*core.ContractCallRequest, error) {
+	var req core.ContractCallRequest
+	s := op.Input.String()
+	if err := json.Unmarshal([]byte(s), &req); err != nil {
+		return nil, i18n.WrapError(ctx, err, i18n.MsgJSONObjectParseFailed, s)
+	}
+	return &req, nil
+}
+
+func OpBlockchainInvoke(op *core.Operation, req *core.ContractCallRequest, batch *BatchPinData) *core.PreparedOperation {
```
awrichar, line 48: It's an anomaly to expose this builder to the world (normally they're "owned" by a single manager), but I didn't see a good way around it now that both Contract Manager and Multiparty Manager need the ability to trigger an invoke operation.

.github/workflows/integration.yml
```diff
@@ -19,27 +19,26 @@ jobs:
             TestEthereumMultipartyTokensRemoteNameE2ESuite,
             TestFabricGatewayE2ESuite,
             TestFabricMultipartyE2ESuite,
+            TestFabricMultipartyGlobalListenerE2ESuite,
```
awrichar, line 22: Because the Fabric version of this test had to be its own suite, I've only added it to the nightly integration (vs. every PR). That also means the first time it will run is the night _after_ this PR gets merged.

Open to thoughts, but we have a lot of jobs on each PR already, and I didn't necessarily want to add another one unless we feel like the coverage is needed.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -726,6 +756,19 @@ func (e *Ethereum) InvokeContract(ctx context.Context, nsOpID string, signingKey
 	if err != nil {
 		return err
 	}
+	if batch != nil {
+		err := e.checkDataSupport(ctx, abi)
+		if err == nil {
+			method, batchPin := e.buildBatchPinInput(ctx, 2, "", batch)
```
awrichar, line 762: Note that we don't know the shape of the batch pin contract the user is ultimately going to call from their contract on-chain. We assume "version 2" of the `pinBatch` arguments (which has 4 args, where "version 1" expected 5 args) - but if we ever add a different version of this `pinBatch` method, there may need to be a way to choose upfront how to have the args encoded.

internal/batch/batch_processor.go
```diff
@@ -182,11 +191,15 @@ func (bp *batchProcessor) addWork(newWork *batchWork) (full, overflow bool) {
 	if !added {
 		newQueue = append(newQueue, newWork)
 	}
+
+	// Special handling for transactions that allow only one message per batch
+	batchOfOne := newWork.msg.Header.TxType == core.TransactionTypeContractInvokePin
```
peterbroadhurst, line 196: Note so I remember to come back to this:

I don't from this line understand how we flush the existing batch out. This makes it `full` when we find there's a `batchOfOne`... but what about when there were already some messages being assembled into a batch when this happened.

internal/batch/batch_processor.go
peterbroadhurst: I think moving it would be better if that's ok @awrichar

internal/batch/batch_processor.go
```diff
@@ -433,7 +445,7 @@ func (bp *batchProcessor) getNextNonce(ctx context.Context, state *DispatchState
 	return nonceState.latest, nil
 }
 
-func (bp *batchProcessor) maskContext(ctx context.Context, state *DispatchState, msg *core.Message, topic string) (msgPinString string, contextOrPin *fftypes.Bytes32, err error) {
+func (bm *batchManager) maskContext(ctx context.Context, state *dispatchState, msg *core.Message, topic string) (msgPinString string, contextOrPin *fftypes.Bytes32, err error) {
```
peterbroadhurst, line 448: same per above for moving this function to the batch manager go file.

internal/batch/batch_processor.go
peterbroadhurst: @awrichar - I didn't find ^^^ if you could help me with where it landed.

internal/blockchain/fabric/fabric.go
```diff
@@ -119,6 +119,10 @@ type Location struct {
 	Chaincode string `json:"chaincode"`
 }
 
+type ContractOptions struct {
+	GlobalListener bool `json:"globalListener"`
```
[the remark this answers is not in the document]
peterbroadhurst: I think the confusion might be, it's still only on one channel. `global` for me, might feel like we're listening on all channels.

Maybe `customPinSupport`, or `multiChaincodeListener`?

internal/multiparty/manager.go
peterbroadhurst: Agree with this concern. I believe this current "retry indefinitely" approach is safe from an existing functionality, and data integrity perspective. So I think merging this approach into main in this PR, then revisiting the design in a future PR ahead of a 1.3 release is the right way to go (rather than blocking this PR from merging)

smart_contracts/ethereum/solidity_firefly/contracts/Firefly.sol
```diff
@@ -2,27 +2,39 @@
...
+        string memory payloadRef;
+        bytes32[] memory contexts;
+        (uuids, batchHash, payloadRef, contexts) = abi.decode(
+            data,
+            (bytes32, bytes32, string, bytes32[])
+        );
+        emit BatchPin(
+            tx.origin,
+            block.timestamp,
+            "firefly:contract_invoke_pin",
```
peterbroadhurst, line 20: Like the clarity of this eyecatcher.
Good to keep it under 32b length string (this is 28), as that's 2-words (len+data) in Eth storage.

peterbroadhurst (changes requested):
This looks great! Thanks for the commentary through the PR - very helpful.

I had one material question on the batch logic (how we close the previous batch). Otherwise all very minor comments.

internal/batch/batch_processor.go
awrichar: Yea, good catch. This was working at some point but then fell off the wagon with one of my (many) refactors of this method.

awrichar: Actually, in coming back to this - each batch processor fundamentally deals with only one type of transaction. So we don't really have to keep peeking at messages... we can just look at `bp.conf.txType` to determine if we're a _real_ batch processor or a one-item-per-batch kind of batch processor.

awrichar: OK... I think it actually would work as-is, but I pushed some logic that hopefully makes this clearer.

Essentially for a batch processor that is assigned to `TransactionTypeContractInvokePin`, the work queue will only have 1 item (full but not overflowing) or 2 items (full and overflowed). From there I think the existing logic will take care of flushing it properly. Correct me if I'm missing anything though.

internal/blockchain/fabric/fabric.go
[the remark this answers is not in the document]
awrichar: Leaning toward `customPinSupport` as maybe the most concise description. If anything, it's overly specific, but it does express the "why" for the intended use of this config.

awrichar:
I believe I've addressed all comments, and I added a tutorial to the documentation. It could probably be fleshed out even further, but it covers the high points.

docs/tutorials/custom_contracts/pinning.md
peterbroadhurst, line 1: This tutorial is great 👍

internal/batch/batch_processor.go
peterbroadhurst: Great! I spent some time re-familiarizing myself with the logic for `overflow` (where later we retain the overflowed item for the next batch) - that was definitely what I was missing when I made the comment 👍

peterbroadhurst (approved):
📍 
📜 
⛓️

PR #1216, opened 2023-03-10
Ensure cached transaction gets updated with new blockchain IDs

awrichar:
There are multiple paths that record the blockchainID for a transaction. Along this path, we were updating the column in the database but not the cache - resulting in this field being intermittently out of sync.

### DIFF
--- a/internal/txcommon/txcommon.go
+++ b/internal/txcommon/txcommon.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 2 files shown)


hfuss (approved):
🧙🏻

internal/txcommon/txcommon.go
```diff
@@ -138,12 +138,13 @@ func (t *transactionHelper) PersistTransaction(ctx context.Context, id *fftypes.
 			return false, nil
 		}
 
-		newBlockchainIDs, changed := tx.BlockchainIDs.AddToSortedSet(blockchainTXID)
+		var changed bool
+		tx.BlockchainIDs, changed = tx.BlockchainIDs.AddToSortedSet(blockchainTXID)
```
awrichar, line 142: Note it's significant to update `tx` because it's cached down on line 162.

nguyer (approved):

PR #1224, opened 2023-03-17
Do not reject messages from an unrecognized identity

### LINKED ISSUE #1221: Late joiners to a multiparty network fail to process previous definitions
Say I build a multiparty network with N members, and they each secure their identity, and then later broadcast some definitions (token pools, interfaces, etc).

Then member N+1 joins the network and begins processing all messages to catch up. Because this member will be processing everything as quickly as possible, and because this processing happens on parallel threads, and because identities and other definitions are on separate topics... it's quite likely that some definition messages may be parsed before the identities of all prior nodes have been parsed and confirmed.

When this happens, the new node will simply reject those definitions:

https://github.com/hyperledger/firefly/blob/35319a22f7a468381105e7f79a4363fe5778879a/internal/events/aggregator.go#L437

We need a better way to handle the "unknown author" situation - possibly some new "parked" state for these definitions, so that we can go back and process them in order if the author becomes resolved.

awrichar:
Because batch downloads happen asynchronously on download manager, it's possible
that the details for an identity claim may be downloaded after other things
created by that identity.

Instead of flatly rejecting all messages from an unknown identity, leave them
in pending state so that they can be unstuck by a later rewind.

Fixes #1221


peterbroadhurst (approved):
👍 - looks great. Thanks for working through all the thinking needed on this one, to determine the fix is actually very small.

PR #1226, opened 2023-03-20
Add "intent" to address resolver plugin interface and API

peterbroadhurst:
The address resolver interface is extremely useful as a checking/enforcement point for signing actions being accepted into the system. However, currently there is no way to distinguish between simply a call to `/api/v1/resolve` to request a resolution of an input string to a signing key from the plugin, vs. the other actions (messages, custom invoke, token ops) that actually accept a signing request into the system.

This PR adds a simple addition field of `.Intent` that is `lookup` for `/api/v1/resolve` and `sign` for all other actions, and is passed into the existing Go API to the blockchain connector, and Go templates on the remote API hook (ethereum plugin only).

### DIFF
--- a/internal/blockchain/fabric/fabric.go
+++ b/internal/blockchain/fabric/fabric.go
@@ -499,7 +499,11 @@ func (f *Fabric) eventLoop() {
 	}
 }
 
-func (f *Fabric) ResolveInputSigningKey(ctx context.Context, signingKeyInput string) (string, error) {
+func (f *Fabric) ResolveSigningKey(ctx context.Context, signingKeyInput string, intent blockchain.ResolveKeyIntent) (string, error) {
+	// Note: "intent" is not currently used for Fabric, as the identity resolution is not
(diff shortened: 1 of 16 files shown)


internal/blockchain/fabric/fabric.go
```diff
@@ -499,7 +499,11 @@ func (f *Fabric) eventLoop() {
 	}
 }
 
-func (f *Fabric) ResolveInputSigningKey(ctx context.Context, signingKeyInput string) (string, error) {
+func (f *Fabric) ResolveSigningKey(ctx context.Context, signingKeyInput string, intent blockchain.ResolveKeyIntent) (string, error) {
+	// Note: "intent" is not currently used for Fabric, as the identity resolution is not
+	//       currently pluggable to external identity resolution systems (as is the case for
+	//       ethereum blockchain connectors).
+
```
nguyer, line 506: Thanks for documenting this!

nguyer (approved):

PR #1231, opened 2023-03-21
Allow FireFly core to start, even if an individual Namespace cannot initialize

peterbroadhurst:
Currently if you specify an invalid batch pin contract address for a namespace, either at startup, or when configuration dynamically reloads, then the whole FireFly core server fails with this error:

```
2023-03-21T11:14:31.573Z]  INFO Resolving FireFly contract at index 0 ns=default pid=41459
[2023-03-21T11:14:31.573Z] DEBUG ==> POST http://127.0.0.1:5102/ breq=Ul7tPf2a pid=41459 proto=ethereum
[2023-03-21T11:14:31.591Z] ERROR <== POST http://127.0.0.1:5102/ [404] (17.56ms) breq=Ul7tPf2a pid=41459 proto=ethereum
Error: FF10111: Error from ethereum connector: FF00164: No result found
```

This same error would occur if there were a temporary error contacting the blockchain through the blockchain connector.

This PR proposes a tweak to this such that this particular namespace will continue to retry its initialization, even if this one namespace cannot be initialized.

To do this the namespace initialization becomes asynchronous, and I had to make a change to the just-in-time creation of the `ff_system` namespace as a result of that (because we can't do the checking on that until after the V1 namespace has initialized, which now becomes async).

While initializing namespaces cannot be used on the API, and are not returned from `/api/v1/namespaces` by default.

However, there is a query to specify you would like to see them:
```
curl -s 'http://localhost:5000/api/v1/namespaces?includeinitializing' | jq
[
  {
    "name": "default",
    "networkName": "default",
    "description": "Default predefined namespace",
    "created": "2023-03-21T15:49:28.333291Z",
    "initializing": true,
    "initializationError": "FF10111: Error from ethereum connector: FF00164: No result found"
  }
```

> Note this PR does **not** change the behavior of FireFly in the case that a **plugin** fails to initialize.


internal/namespace/configreload.go
```diff
@@ -106,13 +106,6 @@ func (nm *namespaceManager) configReloaded(ctx context.Context) {
 		return
 	}
 
-	// Only initialize the updated namespaces (which includes all that depend on above plugins)
-	if err = nm.initNamespaces(ctx, updatedNamespaces); err != nil {
-		log.L(ctx).Errorf("Failed to initialize namespaces after config reload: %s", err)
-		nm.cancelCtx() // stop the world
-		return
-	}
-
```
nguyer, line 115: Trying to work out where the newly added namespaces go through the new retry loop? Sorry if I'm just not seeing it, but wanted to make sure we didn't miss a code path here

nguyer: Ah. yep I see it now. Sorry I just missed that the first time reading through.

pkg/core/namespace.go
```diff
@@ -35,6 +35,12 @@ type Namespace struct {
 	Contracts   *MultipartyContracts `ffstruct:"Namespace" json:"-"`
 }
 
+type NamespaceWithInitStatus struct {
+	*Namespace
+	Initializing        bool   `ffstruct:"NamespaceWithInitStatus" json:"initializing,omitempty"`
```
nguyer, line 40: I'm trying to think if there is already precedence in the API for something like this. Intuitively I would have thought the field would have been `"ready"` or `"initialized"` and its default value would have been `false` until it was set to `true`. But I think I see why it's this way, because it's more consistent with the query parameter. Internally, elsewhere such as on the Namespace Manager we have `started = true/false`. I'm not sure if consistency there is something valuable or not. I'll leave it up to you.

nguyer (approved):
Looks good! One question that's up to you to decide either way

pkg/core/namespace.go
peterbroadhurst: Yeah - was that reason, but also so that the new cruft of a boolean only turns up on the JSON in the case of something that is **not** ready.

peterbroadhurst:
Trying go 1.19 for the swagger - I thought we were still on 1.18 🤔

PR #1234, opened 2023-03-22
Validate and restore listeners on startup, and allow deletion if not found in connector

peterbroadhurst:
Had a situation where we wanted to restore fresh listeners for custom contracts on a FireFly core service, and had assumed FireFly core would re-create the listeners. So deleted them on the EVMConnect runtime directly and restarted FF.

Found related issues, that meant we had to go into the FF DB and remove the contract listeners directly (and have the app recreate them):
1. FireFly did not recreate the listener. This seems a little inconsistent with both the `BatchPin` listener, and the listeners of the token connectors...
   - Proposing in this PR that we automatically recreate the listeners where possible (only Ethereum in this PR, due to existing limitations around the Fabric connector integration where it does not have existing query support to check existence)
2. The API could not be used to delete the listener, because when EVMConnect returned a `404` FireFly failed the `DELETE` API.
    - This seemed like a straight bug to me. Due to the return of status being `interface{}` and the complexity of a `nil` check there, we're adding an extra `bool` to whether it's found or not (Fabric always returns true)
3. Found in testing this change, that the check on existence of a contract listener with the same parameters was not working, passing in a `[]byte` for the `location` rather than `string`


internal/contracts/manager.go
```diff
@@ -106,7 +107,12 @@ func NewContractManager(ctx context.Context, ns string, di database.Plugin, bi b
 		core.OpTypeBlockchainContractDeploy,
 	})
 
-	return cm, nil
+	// Validate all our listeners exist on startup - consistent with the multi-party manager.
+	// This means if EVMConnect is cleared, simply a restart of the namespace on core will
+	// cause recreation of all the listeners (noting that listeners that were specified to start
+	// from latest, will start from the new latest rather than replaying from the block they
+	// started from before they were deleted).
```
nguyer, line 114: I wonder if, rather than storing "latest" we should resolve what the latest block number is at the time, then store that? This is out of scope for this PR, but may be worth thinking about for better determinism.

internal/contracts/manager.go
```diff
@@ -196,6 +202,31 @@ func (cm *contractManager) GetFFIs(ctx context.Context, filter ffapi.AndFilter)
...
+		}
+		if len(listeners) == 0 {
+			log.L(ctx).Infof("Listener restore complete. Verified=%d", verifyCount)
+			return nil
+		}
+		for _, l := range listeners {
+			if err := cm.checkContractListenerExists(ctx, l); err != nil {
+				return err
+			}
+		}
```
nguyer, line 224: I think we need to increment `verifyCount` right here. Currently it doesn't look like it's ever updated.

nguyer (approved):
One minor thing which I think will result in a misleading log message, but other than that, looks good!

PR #1236, opened 2023-03-26
Add an indxed column and simple query API for distinct sub-paths

### LINKED ISSUE #1227: No way to list sub-paths for data
The `data` collection has a `blob.name` field, that is indexed.

The introduction of this field, was to support a convention where this is used as a fully-qualified file path like:
`path/to/my/file.png`

The FireFly REST API has good filtering support for listing files-in-a-folder using a starts-with semantic of `path/to/my/`.

However, there's no way to see what "sub paths" exist, to be able to do find unique sub-path prefixes like `path/to/*/**`.

This is problematic for building file-browser style UI interfaces with pagination.

peterbroadhurst:
Fixes #1227


peterbroadhurst:
Let me find the semantics for setting that name, which as you say are unchanged in this PR...

peterbroadhurst:
Here we go: https://github.com/hyperledger/firefly/pull/381 (pointing to FIR 3)

peterbroadhurst:
The Swagger contains the details. I'm happy to work on something related to instructions for using the `metadata` JSON field or `autometa` fields in a PR if you think there's a gap in the docs / instructions there. I would like to decouple it from this PR though if that's ok.

peterbroadhurst:
These are the correct fields I believe - although I do see that these fields don't have descriptions - not sure if that's a Swagger UI bug, or otherwise:
image

nguyer:
Yep, I'm happy for those enhancements to be a separate PR. I don't think it's urgent, so whenever we get to it. It just seems non-obvious how a user would use the new functionality here though, so as long as that gap gets closed at some point, I'm happy to merge this.

nguyer (approved):

PR #1242, opened 2023-03-28
Re-establish active subscriptions after dynamic Namespace reload

peterbroadhurst:
On a dynamic reload of an _existing_ namespace (for example if there's a tweak to the config) the old namespace is stopped, and all subscription event polling associated with that namespace is stopped,

However, for WebSockets events there is currently no trigger to restart the listeners.

In this `connect-in` case of Events, the Orchestrator relies upon a `RegisterConnection` function call from the Events plugin, when the `start` command comes over the WebSocket to from the application.

In this case the application never disconnects/reconnects... so there's no new `start` payload.

So this PR proposes a new `NamespaceRestarted` function call from Orchestrator _to_ the Events plugin, which then becomes responsible for making new `RegisterConnection` calls for all active subscriptions.

To make this safe to happen on every startup, asynchronously to connections coming in, a timestamp is passed to the `NamespaceRestarted` so it can only re-register connections that came in before the restart time.


nguyer (approved):

nguyer:
Looks like maybe there was a slight coverage drop? Or codecov could just be acting up again, not sure.

peterbroadhurst:
Putting this back into draft, as in dealing with the last bit of coverage (thank you coverage ❤️ ) I found it's because of the difference between the sync and async nature of this function changed by this PR, and look at the comment:
https://github.com/hyperledger/firefly/blob/a30677ba7970d4f9156a0d10bc709a5da3a02814/internal/namespace/manager.go#L384-L406

This is significant. (f an event stream (direct, or indirect via a tokens connector) has an event for a namespace... but the namespace hasn't started yet, I think core will just discard the event.

So I do not know if it is safe to merge this PR, while multiple namespaces are sharing event streams (in either the blockchain directly, or tokens connectors indirectly).

Because batches of events will be being supplied to Core, some of which might be for namespaces that are started, and others might be for namespaces that are still starting (or failed to start).

If Core `ack`s the events (which I believe is the current behavior) after logging that no namespace matched, then messages will have been lost.

There might be a patch, that this situation could be detected - because the namespaces would exist in stopped status. So core could instead of logging, instead block receipt of the events.

That would allow this PR to merge, but we wouldn't have full namespace isolation, until we split out the event streams completely.

peterbroadhurst:
I've pulled in a fix in this workstream, to the concurrent map writes on handlers that are now possible.

peterbroadhurst:
Ok - another step made on this PR, to ensure that:
1. Adding locking around handler registration to avoid concurrent map write errors
2. Namespaces register their handlers against the plugins _synchronously_ on startup/reload with the plugins
3. The bound handlers callbacks return errors if events are dispatched for that namespace, and the namespace is not ready

Between these changes, this protects against any situations where the async nature of Namespace startup, could result in missed events for a namespace that exists in the configuration.

nguyer (approved):
The extra checks and locks look good

peterbroadhurst:
Performed an E2E test in a dev environment:
- Configured env with `--block-period 10`
- Enabled `config.autoReload` in my `firefly_core_0.yml`
- Restarted Core
- Sent pinned broadcast message
- Commented out whole namespace in config - picked up by core without restart
- Waited for receipt/event to arrive from blockchain connector and new event drive rejection logic
- Uncommented namespace
- Confirmed redelivery of event happened, and message was confirmed 👍

Issue #1247, opened 2023-03-30
Initializing a group using a non-registered key fails


awrichar:
It's possible to send private messages where the on-chain transaction is signed by any arbitrary key (such as a user-specific key assigned via an HD Wallet, for instance). The "not found" error for registered verifiers is specifically bypassed for private messages:
https://github.com/hyperledger/firefly/blob/df2581ef959e6c7cce7ef299bc959411d14e201e/internal/events/aggregator.go#L437

However, the very first private message in a given group will automatically trigger a "group init" message if the privacy group doesn't yet exist. This message will be signed by the same key as the private message, but will not bypass the check above (since it's of the special type "groupinit"). Therefore, if the first message on a given group happens to come from a non-registered signing key, the "group init" message will be stuck and never confirmed.

It's also worth noting that this code path changed recently - messages from unrecognized signers would previously be rejected immediately, but as of #1224, they will simply be held in a sent/pending state and not confirmed.

awrichar:
> the "group init" message is unconfirmed, but the group still exists

I dug a little further on this, and it's even a little more problematic than I thought. We optimistically insert the group on the sending node at the same time we send the "group init" message. So there are actually 2 early places where a group is inserted (at sending time and immediately after receiving) before the "group init" message is ever confirmed. I guess we never put a lot of thought into the situation where a "group init" message could be malformed, and what that implies for the group it advertises.

On the receiving side, we could potentially perform the on-chain consistency check earlier - ie fully validate the message _before_ inserting the group. The sender would still consider the group valid because they would have inserted it when they sent the definition, but that might be OK (the other half of this fix involves making sure the sender sends a valid group message so that others don't end up ignoring it).

awrichar:
> On the receiving side, we could potentially perform the on-chain consistency check earlier - ie fully validate the message _before_ inserting the group

Opened #1251 to propose this.

peterbroadhurst:
> use the identity's registered key for this message, instead of an arbitrarily different key

I'm not sure this is the right direction. I would encourage us to go the other direction, and use the key of the first-message to be the group-init key and accept that key as valid.

e.g. just like the private messages themselves, any key can be used to sign the private message that contains the init.

Two benefits of this approach:
- No change to the anonymity approach between group init and the messages
- The key resolution / security check during API submission for the message send only needs to resolve the anonymous key

PR #1248, opened 2023-03-30
Always send "group init" message with a registered signing key

awrichar:
Private messages can be sent and signed by any arbitrary key if desired, but "group init" messages (which can be auto-triggered by a private message) must be signed by a registered key.

Possible/partial fix for #1247


peterbroadhurst (changes requested):
Just rejecting while this discussion is outstanding, to avoid accidental merge:

https://github.com/hyperledger/firefly/issues/1247#issuecomment-1505805421

awrichar:
Closed in favor of #1282

PR #1251, opened 2023-03-31
Check message on-chain consistency before initializing context

awrichar:
This prevents inserting a group from a malformed "group init" message.

Possible/partial fix for https://github.com/hyperledger/firefly/issues/1247

### DIFF
--- a/internal/events/aggregator.go
+++ b/internal/events/aggregator.go
@@ -451,8 +451,8 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 
 	unmaskedContexts := make([]*fftypes.Bytes32, 0)
 	nextPins := make([]*nextPinState, 0)
-	var newState core.MessageState
-	var dispatched bool
+	action := core.ActionWait
+	var correlator *fftypes.UUID
 
 	var cro data.CacheReadOption
 	if pin.Masked {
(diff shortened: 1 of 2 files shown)


internal/events/aggregator.go
```diff
@@ -469,6 +469,12 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 	case !dataAvailable:
 		l.Errorf("Message '%s' in batch '%s' is missing data", msgEntry.ID, manifest.ID)
 	default:
+		// Check the pin signer is valid for the message
+		action, err = ag.checkOnchainConsistency(ctx, msg, pin)
```
awrichar, line 473: This was the main goal of the PR - to move this check earlier instead of having it as part of `attemptMessageDispatch`.

However, since this call returns a `core.MessageAction`, it also made sense to refactor common handling of retry/wait actions up into this method as well, so consequently I split up `attemptMessageDispatch` a little bit to make that happen.

awrichar: Side note: `core.MessageAction` is definitely imperfect and can be slightly confusing. `ActionRetry` means "I also should have returned a non-nil err and you should treat that as retryable" (an err is ignored for any other action). `ActionConfirm` kind of means "looks good to me, go ahead to confirm or continue with the next step if there is one". There could be some further thought on how to handle success vs. the 3 distinct problem scenarios (wait/retry/reject).

internal/events/aggregator.go
```diff
@@ -509,27 +515,28 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
...
-		state.markMessageDispatched(manifest.ID, msg, msgBaseIndex, newState)
-	} else {
+	if action == core.ActionRetry {
+		return err
+	} else if action == core.ActionWait {
+		// We need to prevent dispatch of any subsequent messages on the same topic in the batch
 		for _, unmaskedContext := range unmaskedContexts {
 			state.SetContextBlockedBy(ctx, *unmaskedContext, pin.Sequence)
 		}
+		return nil
```
awrichar, line 528: Note - there a few earlier branches of this method which return `nil` _without_ marking the context blocked. Still trying to decide if that was intentional for some reason, or if they should all fall down to here instead.

peterbroadhurst (approved):

Issue #1250, opened 2023-03-30
Intermittent unit test failures in configreload_test


awrichar:
https://github.com/hyperledger/firefly/actions/runs/4567930148/jobs/8062336444?pr=1249

```
--- FAIL: TestConfigDownToNothingOk (0.02s)
    manager_test.go:138: PASS:	Name()
    manager_test.go:138: PASS:	InitConfig(string)
    manager_test.go:138: PASS:	Init(string,string,string,string,string)
    manager_test.go:138: PASS:	Start()
    manager_test.go:140: PASS:	Name()
    manager_test.go:140: PASS:	InitConfig(string)
    manager_test.go:140: PASS:	Init(string,string)
    manager_test.go:140: PASS:	SetHandler(string,string)
    manager_test.go:140: FAIL:	GetNamespace(string,string)
        		at: [/home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:419 /home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:774]
    manager_test.go:140: FAIL:	GetNamespace(string,string)
        		at: [/home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:420 /home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:774]
    manager_test.go:140: PASS:	GetNamespace(string,string)
    manager_test.go:140: FAIL:	UpsertNamespace(string,mock.AnythingOfTypeArgument,bool)
        		at: [/home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:422 /home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:774]
    manager_test.go:140: FAIL: 5 out of 8 expectation(s) were met.
        	The code you are testing needs to make 3 more call(s).
        	at: [/home/runner/work/firefly/firefly/internal/namespace/manager_test.go:140 /home/runner/work/firefly/firefly/internal/namespace/manager_test.go:281 /home/runner/work/firefly/firefly/internal/namespace/configreload_test.go:804]
    manager_test.go:141: PASS:	Name()
    manager_test.go:141: PASS:	InitConfig(string)
    manager_test.go:141: PASS:	Init(string,string,string)
    manager_test.go:141: PASS:	Start()
    manager_test.go:142: PASS:	Name()
    manager_test.go:142: PASS:	InitConfig(string)
    manager_test.go:142: PASS:	Init(string,string)
    manager_test.go:143: PASS:	Name()
[...]

PR #1259, opened 2023-04-04
Fix coverage gaps from new operation update logic

awrichar:
Address missing coverage from #1257


awrichar:
Unit tests encountered https://github.com/hyperledger/firefly/issues/1250 again...

Issue #1220, opened 2023-03-15
Advanced Object Deletion


nguyer:
## Background

While doing some testing locally, I ended up stuck in a less than ideal state. Here's a summary of what I did:

- Created a "gateway mode" stack connected to a public testnet
- Deployed an ERC-721 token contract that I created with the OpenZeppelin contract wizard
- Created a token pool for that contract
- Attempted to mint a token with a URI
- FireFly told me (based on the response from the token connector) that setting a URI is not supported with my contract
- This was due to the fact that I did not specify the contract interface when creating the token pool
- I went back and created the interface
- I then went to create a new token pool with a new name, while specifying the interface ID
- The FireFly transaction for creating the token pool becomes stuck "Pending", while the Token Create Pool Operation is successful
- Upon inspecting the logs I found a `200` response from the token connector to create the pool, then FireFly core logs:  ```[2023-03-15T16:53:11.943Z] DEBUG Token pool ID=79a6bb2f-11bd-45e6-99a7-f3045ec911ac Locator='address=0xfd403b063ecf959867973a4cf383c48b9a06e19e&schema=ERC721NoData&type=nonfungible' already confirmed ns=default pid=1 role=event-manager```

This now leaves me stuck state. I have no way of deleting the token pool, but because I created it incorrectly, I also have no way of using it. My options at this point are to:

- Deploy a new contract - not viable if this was a pre-existing contract that already has transactions that need to be index
- Completely delete my stack and start over - not viable if I have other production data in my DB
  - Worse, I don't think this would even work if I was in a multi party system, because as soon as my FireFly node comes back online it would see the token pool broadcast in the chain as it re-indexes everything

In reality, this problem extends to more than just token pools, but this situation was exceptionally painful, so it got me thinking about the problem. The same situation could occur with contact interfaces, contract APIs, etc, though those things have versions so it may not be as big of a deal there.

## What needs to change

- We need a way to delete things which have not been broadcasted, which should be a simple curd operation.
- We also need a way to delete a token pool which _has_ been broadcasted

awrichar:
Full list of definition types that this would encompass:
* datatype
* contract interface
* contract API
* identity
* token pool

And a summary of some specific enhancements that were discussed offline among maintainers:
1. Split the concept of "creating a definition" from "publishing a definition" and introduce a new set of APIs for the latter (such as `POST /tokens/pools/{id}/publish`)
[...]

PR #1261, opened 2023-04-04
Separate "define" and "publish" for token pools, and allow deleting unpublished pools

awrichar:
Add a new `publish` boolean query parameter to the `/tokens/pools` creation route, which defaults to "false" and will create the token pool definition locally only. If "true", the definition will be published to the multiparty network (note that this is a **change** to the default behavior, which is currently to publish all token pools in the context of a multiparty network).

Add a new `/tokens/pools/{id}/publish` route for publishing a previously-local definition to the multiparty network as a separate action.

Allow deleting token pools (along with associated transfers, approvals, and balances), but only if they have not been published.

This represents most of the behavior proposed in https://github.com/hyperledger/firefly/issues/1220 (in the context of token pools only).

We should also confirm that "publish" is the [correct word to use here](https://github.com/hyperledger/firefly/issues/1220#issuecomment-1496504327).


awrichar:
I came across another edge case as I was reviewing this again. Pasting my notes from that commit:

> If a published token pool matches a local one, link them together
> 
> There is currently no way to manage two copies of the same token pool (ie same pool
> locator) within a single namespace. Therefore, if a published pool collides with
> the pool locator of a local pool, the most sensible thing to do is to convert the
> local pool to link it up with the published definition (by setting its network name).
> 
> Some caveats:
> - this pool will have a different UUID between nodes (but the same network name)
> - this pool cannot be deleted once it is linked up with a published pool

awrichar:
In order to remove another potential conflict between local pools and published pools, I also had to remove the uniqueness constraint on `poolLocator` (previously within a given namespace, you could only have one pool for a particular `poolLocator`). Now you can have many different pools that are all actually the same underlying contract, and therefore the same `poolLocator`.

awrichar:
This will depend on https://github.com/hyperledger/firefly-tokens-erc1155/pull/119 and https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/129 to function fully (without these enhancements, a published token pool whose poolLocator matches an existing local token pool would be rejected).

awrichar:
Still seem to be hitting https://github.com/hyperledger/firefly/issues/1250 a lot in unit test runs.

internal/definitions/sender_tokenpool.go
```diff
@@ -24,34 +24,91 @@ import (
...
+			pool.NetworkName = networkName
 		}
 
-		if err := pool.Pool.Validate(ctx); err != nil {
-			return err
+		sender = ds.getTokenPoolSender(ctx, pool)
+		if sender.err != nil {
+			return sender.err
 		}
+		if !waitConfirm {
```
peterbroadhurst, line 48: I'm trying to understand/remember why we change the order of persistence based on whether we engaging the sync/async bridge or not. Bit worried these divergent codepaths might leave the door open to different behavior.

internal/definitions/sender_tokenpool.go
```diff
@@ -24,34 +24,91 @@ import (
...
+func (ds *definitionSender) getTokenPoolSender(ctx context.Context, pool *core.TokenPool) *sendWrapper {
+	// Map token connector name -> broadcast name
+	if broadcastName, exists := ds.tokenBroadcastNames[pool.Connector]; exists {
+		pool.Connector = broadcastName
+	} else {
+		log.L(ctx).Infof("Could not find broadcast name for token connector: %s", pool.Connector)
+		return wrapSendError(i18n.NewError(ctx, coremsgs.MsgInvalidConnectorName, broadcastName, "token"))
+	}
+
+	if err := pool.Validate(ctx); err != nil {
```
peterbroadhurst, line 75: Note to self to go look at Validate, and check if we do a pre-publish check on the uniqueness of the `networkName`, as it seems like we haven't committed to a DB update at this point (at least in the `waitConfirm=true` path).

peterbroadhurst (changes requested):
Overall looks absolutely great.
A few slightly waffly bits of commentary from me at this point @awrichar, unclear if I just need a bit more explanation or it makes sense to make a change for any of them.

internal/definitions/sender_tokenpool.go
awrichar: Yea, our pre-publish logic is wildly inconsistent for most definitions. Maybe should be cleaning it up as I go, or should at least come back round on all of them.

awrichar:
Another note to myself - do we need to delete event listeners (for transfers etc)?

internal/definitions/sender_tokenpool.go
awrichar: I'm trying to page back in as well... I think because this method is only called from the API server, there are two possible outcomes:
1. I requested async, we have to persist the token pool here in order to add the message ID
2. I requested sync, the token pool is going to get persisted anyway before I respond (ie by the event loop), so we skip this upfront save

I don't _think_ there's any problem here, but I acknowledge it's probably an over-optimization in (2). Perhaps cleaner to always Prepare+Upsert upfront?

awrichar: I guess in the case that the message in (2) is lost or stuck, the token pool <-> message linkage might not be persisted to the database, and that would be problematic at least for debugging.

awrichar: After further pondering, I've decided we shouldn't Upsert with the message ID at all. We need to wait until the message is received/confirmed. In the (rare) case of a race, it should be the winning/first message that gets persisted onto the token pool object. If we persist the ID of an unconfirmed message, it becomes more complex to handle all the possible branches when a message is later confirmed.

internal/definitions/sender_tokenpool.go
awrichar: OK as noted in https://github.com/hyperledger/firefly/pull/1261#discussion_r1195556197 I'll remove the pre-flight upsert from the `waitConfirm=false` path, and add a new pre-flight check on network name uniqueness for all paths.

awrichar:
I've added a new commit which fixes some of the inconsistencies noted. Specifically:
- there is a uniqueness check on `networkName` before every publish
- there is a multi-faceted uniqueness check on `id`, `name`, and `networkName` after every publish, with the following behaviors
  - `id` conflict: If this is a published confirmation of an unpublished pool, overwrite it. Otherwise, reject it.
  - `name` conflict: Attempt to generate a new unique name.
  - `networkName` conflict: Reject it.

However, I'm still not completely sure this is correct. I think there's a small possibility of someone "maliciously" publish a pool that could overwrite your local unpublished pool, and that feels wrong.

The problem is that in the current state of things in multiparty mode, we do _not_ write any definition info to the database upfront (ie we only store the definition message) - the actual object is only created after the definition is confirmed. With this new setup, we have a situation where an _already existing_ item is getting _replaced_ by a published item, and it's difficult to quantify when that's allowed.

awrichar:
OK, added one more commit to hopefully patch this hole. A local/unpublished pool can only be overwritten by a published pool if the publish message came from the local root org. I refactored the code to be as readable as possible, although there are still a lot of branches.

There are still some extreme edge cases where you could cause a bit of trouble by intentionally causing UUID conflicts between published and unpublished pools... but none that end in stuck messages or overwriting of data, so I feel OK about it.

internal/definitions/handler_tokenpool.go
```diff
@@ -42,33 +43,51 @@ func (dh *definitionHandler) handleTokenPoolBroadcast(ctx context.Context, state
...
 
 	pool.Message = msg.Header.ID
-	return dh.handleTokenPoolDefinition(ctx, state, pool)
+	pool.Name = pool.NetworkName
+	pool.Published = true
+	return dh.handleTokenPoolDefinition(ctx, state, pool, msg.Header.Author)
 }
 
-func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, state *core.BatchState, pool *core.TokenPool) (HandlerResult, error) {
+func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, state *core.BatchState, pool *core.TokenPool, author string) (HandlerResult, error) {
```
awrichar, line 51: This method has become quite complicated as it now needs to handle 3 different paths:
1. This is a local definition being created
2. This was a local definition on this node, and has now been published
3. This is a definition being created and immediately published by this node
4. This is a definition published by another node

I'm not entirely sure it handles all situations in the clearest or most optimal manner, but I think it does handle them consistently.

internal/definitions/handler_tokenpool.go
```diff
@@ -41,34 +42,58 @@ func (dh *definitionHandler) handleTokenPoolBroadcast(ctx context.Context, state
...
+
 	pool.Message = msg.Header.ID
-	return dh.handleTokenPoolDefinition(ctx, state, pool)
+	pool.Name = pool.NetworkName
+	pool.Published = true
+	return dh.handleTokenPoolDefinition(ctx, state, pool, isAuthor)
 }
 
-func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, state *core.BatchState, pool *core.TokenPool) (HandlerResult, error) {
+func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, state *core.BatchState, pool *core.TokenPool, isAuthor bool) (HandlerResult, error) {
```
awrichar, line 57: This method has become quite complicated as it now needs to handle 3 different paths:

1. This is a local definition being created
2. This was a local definition on this node, and has now been published
3. This is a definition being created and immediately published by this node
4. This is a definition published by another node

I'm not 100% sure it handles all situations in the clearest or most optimal manner, but I think it does handle them consistently.

peterbroadhurst (approved):
Thanks @awrichar for addressing all my comments, and providing the extra context in the PR comments as well.

Added approval - will leave the decision on the DB migration to you

db/migrations/postgres/000109_fix_tokentransfer_index.up.sql
```diff
@@ -2,6 +2,6 @@ BEGIN;
 DROP INDEX tokentransfer_protocolid;
 DROP INDEX tokenapproval_protocolid;
 
-CREATE UNIQUE INDEX tokentransfer_protocolid ON tokentransfer(namespace, connector, protocol_id);
-CREATE UNIQUE INDEX tokenapproval_protocolid ON tokenapproval(namespace, connector, protocol_id);
+CREATE UNIQUE INDEX tokentransfer_protocolid ON tokentransfer(namespace, pool_id, protocol_id);
+CREATE UNIQUE INDEX tokenapproval_protocolid ON tokenapproval(namespace, pool_id, protocol_id);
```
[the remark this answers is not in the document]
awrichar: Well I ended up with two other brand new database migrations anyway, so I might as well roll these lines into 110 or 111 instead of modifying this one.

awrichar:
Added two more commits to actually delete token listeners via a new `/deactivatepool` API added to token connectors. Will depend on these changes in the connectors:
https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/136
https://github.com/hyperledger/firefly-tokens-erc1155/pull/125

peterbroadhurst (approved):

PR #1262, opened 2023-04-05
Add route for deleting token pools

awrichar:
This allows deleting any token pool, regardless of gateway or multiparty mode, along with all associated token transfers, balances, and approvals.

Note that in multiparty mode, if you delete a token pool that was broadcast to the network, you will be voluntarily out of sync with other nodes.

This represents the simplest implementation of (4) from https://github.com/hyperledger/firefly/issues/1220#issuecomment-1496422542, in the context of token pools only.

Note that this does not purge event listeners from the token connector (events will just be ignored). Probably need to find a way to delete them though.


awrichar:
Merged into #1261

PR #1271, opened 2023-04-14
Pass `key` on queries to blockchain, after resolving with an intent to query

peterbroadhurst:
Currently on a `query` to a custom blockchain transaction:
- FireFly resolves the `key` from the input, with an intent to `sign` a blockchain transaction (rather than do a query)
- Discards the `key` without passing it to EVMConnect/FabConnect

This is problematic in EVM based chains, where even you can simulate a transaction with `eth_call` from any signing address (without signing), and the result of the execution of that query might be different based on the input signing key.

So this PR proposes:
- Introducing a new `query` intent for resolving the key
- Passing the `signingKey` to the connector
- Updating Ethereum and Fabric implementations to pass this key to the connector

> Also includes a manifest change to pull in https://github.com/hyperledger/firefly-evmconnect/pull/74


internal/contracts/manager.go
```diff
@@ -349,7 +349,12 @@ func (cm *contractManager) DeployContract(ctx context.Context, req *core.Contrac
 }
 
 func (cm *contractManager) InvokeContract(ctx context.Context, req *core.ContractCallRequest, waitConfirm bool) (res interface{}, err error) {
-	req.Key, err = cm.identity.ResolveInputSigningKey(ctx, req.Key, identity.KeyNormalizationBlockchainPlugin)
+	keyResolver := cm.identity.ResolveInputSigningKey
```
peterbroadhurst, line 352: Note I went for separate methods on the identity interface (even though it makes this calling code more complex), so as to very clearly distinguish the query case and try to minimize the risk of any accidental future use of the `query` intent on other code paths.

peterbroadhurst:
Need to address this in the E2E - :
```
firefly_e2e_firefly_core_1  | [2023-04-14T18:59:30.834Z]  INFO Resolving FireFly contract at index 0 ns=default pid=1
firefly_e2e_firefly_core_1  | [2023-04-14T18:59:30.834Z] DEBUG ==> POST http://evmconnect_1:5008/ breq=BvWCXpyw pid=1 proto=ethereum
firefly_e2e_firefly_core_1  | [2023-04-14T18:59:30.837Z] ERROR <== POST http://evmconnect_1:5008/ [500] (3.84ms) breq=BvWCXpyw pid=1 proto=ethereum
firefly_e2e_firefly_core_1  | [2023-04-14T18:59:30.837Z] ERROR namespace default attempt 3: FF10111: Error from ethereum connector: FF23020: Invalid 'to' address '': bad address - must be 20 bytes (len=0) pid=1
```

nguyer (approved):
Looks good to me. Thinking back, I can't remember why we didn't implement this at the time. I'm wondering if ethconnect fully supported it? (Back when it was just ethconnect)

PR #1278, opened 2023-04-18
Do not retry after a rejected message

awrichar:
Fixes regression from #1251

### DIFF
--- a/internal/events/aggregator.go
+++ b/internal/events/aggregator.go
@@ -535,8 +535,7 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 		np.IncrementNextPin(ctx, ag.namespace)
 	}
 	state.markMessageDispatched(manifest.ID, msg, msgBaseIndex, newState)
-
-	return err
+	return nil
 }
 
 func needsTokenTransfer(msg *core.Message) bool {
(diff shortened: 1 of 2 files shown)


internal/events/aggregator.go
```diff
@@ -535,8 +535,7 @@ func (ag *aggregator) processMessage(ctx context.Context, manifest *core.BatchMa
 		np.IncrementNextPin(ctx, ag.namespace)
 	}
 	state.markMessageDispatched(manifest.ID, msg, msgBaseIndex, newState)
-
-	return err
+	return nil
```
peterbroadhurst, line 538: Is this a smell that we're not checking the `err = ag.checkOnchainConsistency` error above?
Or else, what path was it that dropped an error down to here that we did **not** want to return?

awrichar: The contract of this method is that it should only return retryable errors, which should happen on line 522. No new errors can happen between that line and here.

Rejection sometimes returns a reason in the error field, which is generally just logged but shouldn't trigger retry.

We probably could spend another few days cleaning up all these paths though.

awrichar: I think the specific path that might return reject with err is via ag.definitions.HandleDefinitionBroadcast.

internal/events/aggregator.go
```diff
@@ -607,6 +606,10 @@ func (ag *aggregator) readyForDispatch(ctx context.Context, msg *core.Message, d
 		var handlerResult definitions.HandlerResult
 		handlerResult, err = ag.definitions.HandleDefinitionBroadcast(ctx, &state.BatchState, msg, data, tx)
 		log.L(ctx).Infof("Result of definition broadcast '%s' [%s]: %s", msg.Header.Tag, msg.Header.ID, handlerResult.Action)
+		if handlerResult.Action == core.ActionReject {
+			log.L(ctx).Infof("Definition broadcast '%s' rejected: %s", msg.Header.ID, err)
+			err = nil
```
peterbroadhurst, line 611: Thanks for adding this 👍

internal/events/aggregator.go
awrichar: I added another commit to log and drop the error along that specific path, which also likely solves the issues I was seeing. But I think it still feels pretty correct to return nil here, at least as the code stands right now. Figuring out how to spell retryable vs. non-retryable errors throughout this whole package has been challenging.

PR #1279, opened 2023-04-18
Separate "define" and "publish" for contract interfaces, and allow deleting unpublished interfaces

awrichar:
Part of the fix for #1220
~~In a chain with #1261~~
Depends on https://github.com/hyperledger/firefly-common/pull/64

### DIFF
--- a/internal/definitions/sender_tokenpool.go
+++ b/internal/definitions/sender_tokenpool.go
@@ -71,11 +71,22 @@ func (ds *definitionSender) getTokenPoolSender(ctx context.Context, pool *core.T
 	if err := pool.Validate(ctx); err != nil {
 		return wrapSendError(err)
 	}
-	existing, err := ds.database.GetTokenPoolByNetworkName(ctx, pool.Namespace, pool.NetworkName)
+	existing, err := ds.database.GetTokenPoolByNetworkName(ctx, ds.namespace, pool.NetworkName)
 	if err != nil {
 		return wrapSendError(err)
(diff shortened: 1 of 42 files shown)


internal/definitions/sender_tokenpool.go
```diff
@@ -71,12 +71,23 @@ func (ds *definitionSender) getTokenPoolSender(ctx context.Context, pool *core.T
...
 	}
+	if pool.Interface != nil && pool.Interface.ID != nil {
+		iface, err := ds.database.GetFFIByID(ctx, ds.namespace, pool.Interface.ID)
+		switch {
+		case err != nil:
+			return wrapSendError(err)
+		case iface == nil:
+			return wrapSendError(i18n.NewError(ctx, coremsgs.MsgContractInterfaceNotFound, pool.Interface.ID))
+		case !iface.Published:
+			return wrapSendError(i18n.NewError(ctx, coremsgs.MsgContractInterfaceNotPublished, pool.Interface.ID))
```
peterbroadhurst, line 88: Great to see this case covered 👍

peterbroadhurst (approved):
Looks great!

Only thing I noticed was that coverage dropped in the SQL package with those updates.
Happy to leave it with you whether to close that before merge.

awrichar:
I double-checked coverage on `ffi_sql.go` locally and it seems to be covered - so writing off the bot as a mistake. I'll check again after merging though.

PR #1280, opened 2023-04-18
Fix timing issue between Init and ConfigReload of Namespace Manager

peterbroadhurst:
Event after #1242 checked in, @awrichar saw an instance of #1250 in this test run, with slightly different symptoms https://github.com/hyperledger/firefly/actions/runs/4728313082/jobs/8389728445

~These are a nightmare to diagnose, as Go tests do not log start/end of tests (unless in full debug mode) so a `goroutine` bleading past the end of the test boundary is hard to debug. But with some staring, I believe the problem is that some tests was only waiting _once_ at the end for the reload, and because of that there's the possibility for timing to happen such that the reload is detected before the first initialization has completed.~

~This proposed change adds a wait in the middle~

> See additional diagnosis below
> - locking update to ensure state propagation issue due to a lack of locking between goroutines.
> - unit test file writing change to rename in, rather than write in place

```
panic: 
assert: mock: The method has been called over 1 times.
	Either do one more Mock.On("NamespaceRestarted").Return(...), or remove extra call.
	This call was unexpected:
		NamespaceRestarted(string,time.Time)
		0: "ns1"
		1: time.Date(2023, time.April, 18, 3, 40, 40, 19021351, time.Local)
	at: [/home/runner/work/firefly/firefly/internal/namespace/plugin.go:89 /home/runner/work/firefly/firefly/internal/namespace/manager.go:314 /home/runner/work/firefly/firefly/internal/namespace/retry.go:56 /home/runner/work/firefly/firefly/internal/namespace/manager.go:301 /home/runner/work/firefly/firefly/internal/namespace/asm_amd64.s:1571]
```

PR #1281, opened 2023-04-18
Address coverage gap, and issue with intent propagation found via it

peterbroadhurst:
In addressing all the coverage gaps I found in main (`internal/database`,`internal/events` and `internal/identity`) I found an issue with the new `ResolveQuerySigningKey` using the wrong intent on a default branch. The higher `sign` intent would have been used, rather than `query`.


peterbroadhurst:
Requires #1280 to merge for tests to pass

nguyer (approved):

PR #1282, opened 2023-04-18
Accept private groupinit messages from any signing key

### LINKED ISSUE #1247: Initializing a group using a non-registered key fails

awrichar:
Since we allow private messages to originate from an unregistered blockchain signing key,
we must also accept groupinit messages from the same (since the first private message
to a group will trigger group creation as a side-effect).

Fixes #1247


hfuss (approved):

peterbroadhurst (approved):
👍

PR #1285, opened 2023-04-18
Enable workflows for release branch

awrichar:


.github/workflows/go.yml
```diff
@@ -2,9 +2,9 @@ name: Go
 
 on:
   push:
-    branches: [main]
+    branches: [main, release-1.2]
```
nguyer, line 5: I am pretty sure if we use a filter here, we could have it account for future releases as well

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters

```suggestion
    branches: [main, release-**]
```

.github/workflows/go.yml
```diff
@@ -2,9 +2,9 @@ name: Go
...
   pull_request:
-    branches: [main]
+    branches: [main, release-1.2]
```
nguyer, line 7:
```suggestion
    branches: [main, release-**]
```

nguyer (approved):
Looks good for the immediate fix. I have a suggestion for a slight change that should (though untested by me) account for future release branches as well.

PR #1289, opened 2023-04-24
chore: remove deprecated config

EnriqueL8:
- Remove deprecated config for blockchain, tokens, factory, database, dex and shared storaged config

PR #1290, opened 2023-04-24
feat: Upgrade to latest ffresty with mTLS

EnriqueL8:
Depends on:
- https://github.com/hyperledger/firefly-common/pull/65
- https://github.com/hyperledger/firefly/pull/1289

Once those are in I'll update this one


internal/blockchain/ethereum/address_resolver.go
```diff
@@ -55,11 +55,18 @@ type addressResolverInserts struct {
 }
 
 func newAddressResolver(ctx context.Context, localConfig config.Section, cacheManager cache.Manager, enableCache bool) (ar *addressResolver, err error) {
+
+	client, err := ffresty.New(ctx, localConfig)
+
+	if err != nil {
+		return nil, err
```
peterbroadhurst, line 62: Looks like an extra test is needed to retain coverage and cover this branch.

Something like: https://github.com/hyperledger/firefly-signer/blob/bece7de7dc20d55bc99517a8255c2ec93f596581/internal/rpcserver/server_test.go#L63-L75

peterbroadhurst (changes requested):
Changes look great @EnriqueL8

If you wouldn't mind including the corresponding tests for coverage as well before we merge.

Not sure why the code coverage bot didn't run on your branch to help with that 🤔 

Toggling code coverage in your vscode for each of the affected packages should help though.

EnriqueL8:
@peterbroadhurst let's hold off on this one until I get the wsclient changes in

internal/blockchain/ethereum/config.go
```diff
@@ -74,6 +74,8 @@ const (
 func (e *Ethereum) InitConfig(config config.Section) {
 	e.ethconnectConf = config.SubSection(EthconnectConfigKey)
 	wsclient.InitConfig(e.ethconnectConf)
+	ffresty.InitConfig(e.ethconnectConf)
```
peterbroadhurst, line 77: Why is this now needed @EnriqueL8?

EnriqueL8: So we hide that underneath the init config from the wsclient since `wsClient.InitConfig` calls `ffresty.InitConfig`. So if we ever want to separate the configs for the ws and ffresty we would need to call it here, but since they are the same today I can remove this line

peterbroadhurst (approved):
Thanks @EnriqueL8

PR #1291, opened 2023-04-24
Use a more optimistic approach when inserting token transfer events

matthew1001:
This PR aims to improve the performance of handling token transfer event batches from the token connector.

It stems from some performance testing of FireFly's ability to process contract invocations & their respective token transfer events. Analysis of FireFly logs show that for a single token transfer event we go through the following DB calls:

1. Begin TX
2. Query token pool **<<-- could be cached**
3. Query token transfer  **<<-- no entry found in most cases**
4. Insert into blockchain events
5. Query token transfer again.  **<<-- In many cases this 2nd call can take 0.025 seconds on its own**
6. Insert token transfer
7. Query token balance
8. Insert token balance
9. Lock namespace
10. Insert events
11. Commit

There is a precedent in other areas of the codebase for using more "optimistic" approach for DB inserts that are likely to succeed first time without a clash. This follows an approach of "insert and handle conflicts" rather than "check for existing entry, then insert if not found".

This PR aims to improve `2` and `5/6`. In the case of `2` caching will be used to prevent having to call out to the DB for every event. For `5/6` we insert first and check for conflict errors. In the majority of cases this is expected to eliminate `5` entirely.

Notes on code changes:

1. In other places in the codebase where we do an optimistic `insert`, we don't support `update` - only `insert` and `get`. However in the case of token transfers we do allow a token transfer to be updated. This leaves us with the slightly oddly named `InsertUpdateOrGetTokenTransfer()` function which will:
   -  Insert a token transfer if none exists already
   -  Update a token transfer if it does already exist
   -  Return the existing token transfer if present
2. I've removed the previous `UpsertTokenTransfer()` since it is no longer needed
3. There is still a call to `GetTokenTransferByProtocolID()` (step `3` above). I feel like this could be removed entirely but I haven't done that yet. @awrichar could maybe do with some guidance on this part of the code: https://github.com/hyperledger/firefly/blob/5a78f152021e5cb0d32d8ce35cf28aeb03a17534/internal/events/tokens_transferred.go#L77 Removing that Get call seems obvious, but that affects the order of other calls (e.g. `maybePersistBlockchainEvent`, `emitBlockchainEventMetric`...). We could move `em.database.InsertUpdateOrGetTokenTransfer(ctx, &transfer.TokenTransfer)` to before https://github.com/hyperledger/firefly/blob/5a78f152021e5cb0d32d8ce35cf28aeb03a17534/internal/events/tokens_transferred.go#L84 but I didn't want to start reordering those calls without more input from someone like yourself.


matthew1001:
> I didn't want to start reordering those calls without more input from someone like yourself.

Indeed, having tried to remove the initial call to `GetTokenTransferByProtocolID` I see occasional errors in the FF logs relating to the web-socket client trying to ACK an event that doesn't exist:

`ERROR Sending protocol error to client: FF10176: Invalid data: FF10175: Acknowledgment does not match an inflight event + subscription pid=1 websocket=b28505cc-315f-4309-802c-1bb5e87bb472`

I assume that having combined the check for an existing transfer into the `insert` call prevents us from committing the new blockchain event if-and-only-if the transfer didn't exist previously. So for now I'm leaving the original commit which doesn't remove the initial `GetTokenTransferByProtocolID`.

internal/events/tokens_transferred.go
```diff
@@ -103,10 +103,17 @@ func (em *eventManager) persistTokenTransfer(ctx context.Context, transfer *toke
 	em.emitBlockchainEventMetric(transfer.Event)
 	transfer.BlockchainEvent = chainEvent.ID
 
-	if err := em.database.UpsertTokenTransfer(ctx, &transfer.TokenTransfer); err != nil {
+	existing, err := em.database.InsertUpdateOrGetTokenTransfer(ctx, &transfer.TokenTransfer)
```
awrichar, line 106: I agree that I think you should be able to remove the earlier `GetTokenTransferByProtocolID` and rely only on this call to find duplicates. The earlier inserts for the transaction and blockchain event should be idempotent and replayable in and of themselves.

It could cause an issue I'm not thinking of now... but I don't see any way that could cause an issue with websocket acks like [what you were describing](https://github.com/hyperledger/firefly/pull/1291#issuecomment-1525735229). Perhaps it could be a coincidence, and that is a separate issue?

matthew1001:
Caching of token pools now added under commit ff26e41c8b096c76e6a754994f60001fd9f00b8e

matthew1001:
As a side note I've raised https://github.com/hyperledger/firefly/issues/1294 because it looks like we could produce misleading blockchain event metrics with the code as it stands. I didn't want to tackle that here but could do it as a quick follow on piece of work.

internal/apiserver/swagger_check_test.go
```diff
@@ -65,5 +65,5 @@ func TestDiffSwaggerYAML(t *testing.T) {
 	expectedSwaggerHash := sha1.New()
 	expectedSwaggerHash.Write(existingSwaggerBytes)
 
-	assert.Equal(t, actualSwaggerHash.Sum(nil), expectedSwaggerHash.Sum(nil), "The swagger generated by the code did not match the swagger.yml file in git. Did you forget to run `make reference`?")
+	// assert.Equal(t, actualSwaggerHash.Sum(nil), expectedSwaggerHash.Sum(nil), "The swagger generated by the code did not match the swagger.yml file in git. Did you forget to run `make reference`?")
```
awrichar, line 68: This should not be removed

internal/assets/token_pool.go
```diff
@@ -157,6 +158,12 @@ func (am *assetManager) GetTokenPools(ctx context.Context, filter ffapi.AndFilte
 }
 
 func (am *assetManager) GetTokenPool(ctx context.Context, connector, poolName string) (*core.TokenPool, error) {
```
awrichar, line 160: Is this method actually used? It feels redundant with `GetTokenPoolByNameOrID`.

internal/database/sqlcommon/tokentransfer_sql.go
```diff
@@ -69,86 +70,89 @@ var (
...
+		return nil, s.CommitTx(ctx, tx, autoCommit)
+	}
+
+	// Do a select within the transaction to determine if the protocolID already exists
+	existing, err = s.GetTokenTransferByProtocolID(ctx, transfer.Namespace, transfer.Connector, transfer.ProtocolID)
+	if err != nil || existing != nil {
+
+		if existing != nil {
+			if _, err = s.UpdateTx(ctx, tokentransferTable, tx,
+				sq.Update(tokentransferTable).
```
awrichar, line 126: Did you determine if there's a situation where we should update an existing transfer? My thinking was that this shouldn't actually happen (and any code previously added for updates may need to be removed).

internal/assets/token_pool.go
awrichar: Removed this

internal/apiserver/swagger_check_test.go
matthew1001: Bah, shouldn't have been included in the commit. Been seeing issues with it on local builds so commented out but accidentally included. New commit re-instates it.

internal/database/sqlcommon/tokentransfer_sql.go
matthew1001: I'm not sure I'm 100% confident that we never update them, but I can't think of a reason. I've added commit 0ebd4adae8e73b89150bec35a398c864ffac81a3 to remove the update logic. I think I could do with your guidance if any more tidy up needs doing that might have been assuming token transfers can be updated?

awrichar (approved):
Looks great. Thanks for circling back on the insert-vs-update question - I agree with where it landed.

Issue #1121, opened 2022-12-16
FireFly crash if a token pool event is received with an empty data string


awrichar:
If "poolDataString" is an empty string and therefore "poolData" is nil, this line results in a nil pointer dereference.

https://github.com/hyperledger/firefly/blob/490745489fe98779cff07821d19ba7a83b8421bf/internal/tokens/fftokens/fftokens.go#L335

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xbef3d1]

goroutine 58 [running]:
github.com/hyperledger/firefly/internal/tokens/fftokens.(*FFTokens).handleTokenPoolCreate(0xc0002cf960, {0x1093520, 0xc0007e1ec0}, 0xc000d3a630, 0x0)
	/firefly/internal/tokens/fftokens/fftokens.go:335 +0x371
github.com/hyperledger/firefly/internal/tokens/fftokens.(*FFTokens).handleMessage(0xc0002cf960, {0x1093520, 0xc0007e1d00}, {0xc00087b000, 0x3f3, 0x400})
	/firefly/internal/tokens/fftokens/fftokens.go:511 +0x434
github.com/hyperledger/firefly/internal/tokens/fftokens.(*FFTokens).handleMessage(0xc0002cf960, {0x10935c8, 0xc00024d530}, {0xc000840100, 0x44c, 0x580})
	/firefly/internal/tokens/fftokens/fftokens.go:506 +0x878
github.com/hyperledger/firefly/internal/tokens/fftokens.(*FFTokens).eventLoop(0xc0002cf960)
	/firefly/internal/tokens/fftokens/fftokens.go:553 +0x28f
created by github.com/hyperledger/firefly/internal/tokens/fftokens.(*FFTokens).Init
	/firefly/internal/tokens/fftokens/fftokens.go:234 +0x296

```

EleisonC:
Hello @awrichar I am new to hyperledger and would like to pick this issue up

dwertent:
@EnriqueL8 It looks like this issue was resolved

PR #1296, opened 2023-05-03
Retry tokens events consistent with FFDX, as we cannot push back nack

peterbroadhurst:
Make this consistent with what we did for FFDX, and prevents a scenario on startup as seen in the integration test where a token event comes in before the namespace is started:

```
firefly_core_0_1  | [2023-04-30T00:35:08.431Z] DEBUG Calling TokenPoolCreated callback. Locator='address=0x8875e026c26cedc2ab9880ac56d31e328428db73&schema=ERC20WithData&type=fungible' TX=token_pool/66fffabe-5b11-4118-abc2-50b95877654d pid=1 proto=fftokens role=event-loop
firefly_core_0_1  | [2023-04-30T00:35:08.431Z] ERROR Event loop exiting (FF10446: Namespace 'default' is not started). Terminating server! pid=1 proto=fftokens role=event-loop
```

Also made a change to both FFDX and FFTokens to split `eventRetry` out from `retry` in the config, as I realized looking at this PR that the config was overlapping with the `ffresty` config for request retry.

### DIFF
--- a/internal/tokens/fftokens/fftokens.go
+++ b/internal/tokens/fftokens/fftokens.go
@@ -354,12 +354,15 @@ func (ft *FFTokens) handleTokenPoolCreate(ctx context.Context, data fftypes.JSON
 	blockchainEvent := data.GetObject("blockchain")
 
 	poolDataString := data.GetString("data")
-	if poolData == nil && poolDataString != "" {
-		// We want to process all events, even those not initiated by FireFly.
-		// The "data" argument is optional, so it's important not to fail if it's missing or malformed.
-		if err = json.Unmarshal([]byte(poolDataString), &poolData); err != nil {
-			log.L(ctx).Warnf("TokenPool event data could not be parsed - continuing anyway (%s): %+v", err, data)
-			poolData = &tokenData{}
+	if poolData == nil {
+		poolData = &tokenData{}
+		if poolDataString != "" {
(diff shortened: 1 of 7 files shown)


internal/tokens/fftokens/fftokens.go
```diff
@@ -346,12 +354,15 @@ func (ft *FFTokens) handleTokenPoolCreate(ctx context.Context, data fftypes.JSON
 	blockchainEvent := data.GetObject("blockchain")
 
 	poolDataString := data.GetString("data")
-	if poolData == nil && poolDataString != "" {
```
peterbroadhurst, line 349: In unit testing, I found the `poolData == nil && poolDataString == ""` case was uncaught, and resulted in a panic

peterbroadhurst:
https://github.com/hyperledger/firefly/pull/1296#issuecomment-1533252738 appears unrelated to the change, but does need diagnosis. Raised https://github.com/hyperledger/firefly/issues/1297

internal/tokens/fftokens/fftokens.go
awrichar: https://github.com/hyperledger/firefly/issues/1121? 🙏

peterbroadhurst:
fixes https://github.com/hyperledger/firefly/pull/1296

awrichar (approved):

PR #1305, opened 2023-05-17
Replace token pool "confirmed" state with "active" bool

awrichar:
Token pool states were originally given similar names to message states, but this has ended up being confusing (because the token pool creation flow also includes a message, but the token pool states have no relation to the identically named message states).

Old terminology:
When pool definition message is "confirmed", pool is created as "pending"
When pool gets "activated" by the connector, pool is moved to "confirmed"

New proposed terminology:
When pool definition message is "confirmed", pool is created as "not active"
When pool gets "activated" by the connector, pool is moved to "active"

My hope is that the new terminology is easier to follow, since it does not overlap with messaging states and more clearly denotes the action of "activating" a token pool. Also with #1261 in the works (which means a pool may get activated first and then _later_ be published by a broadcast message), I think it's better to separate the token pool and message terminology.

This is technically a breaking API change, although I'm not aware of any significant usefulness of the "state" field to
external applications (it's primarily an internal tracking field). The related _event_ called "token_pool_confirmed" will
remain unchanged - it's still consistent with other definitions like datatypes, identities, etc, and I think it's still
appropriate.


nguyer (commented):
I like the name change. I do have concerns about it not being backwards compatible though.

awrichar:
I think #1261 will be the bigger backwards-incompatible change in the next release... so might as well rope them together while you have to migrate your token pool code anyway, right?

awrichar:
@peterbroadhurst I guess my comment made it seem like the state was called `Not Active`. It's a bool as I've proposed it here, so it's either `Active` or not `Active`. I can draw a picture too.

nguyer (approved):

PR #1312, opened 2023-05-22
fix: bug with init config

EnriqueL8:
Draft PR where I would appreciate some pointers on my approach!


peterbroadhurst:
Issue related to this is #688

PR #1313, opened 2023-05-22
Do not allow a batch to contain messages from different signing keys

### LINKED ISSUE #1311: Private messages from different signing keys can get out of order
My app sends two different private messages very close together, from the same identity but signed by different keys.

Message `50cddd48-b6c2-449b-a243-f5e7b5c4ab6f` is sent first and gets nonce #0, pin hash `babf8903ad7adbb6b602a190e16c663562cd69e4ae340931fb79dcf86aaa0be5`
Message `2b21e7c5-d6c4-48e7-bd85-c770273f1a75` is sent second and gets nonce #1, pin hash `9f270b2af4693565e0fcb57372122643fefcefd951885707547af6a64c758288`

They're placed in two different batches due to the different signing keys, but the second one somehow gets ordered first, meaning its pin hash doesn't match the zerohash and it's ignored. So that message will never confirm, and I think this topic is now broken.
```
[2023-05-18T22:50:35.006Z] DEBUG Aggregating pin 0000000027 batch=ab47a8bd-7183-428b-9ef5-5a6b869ef1ea msg=2b21e7c5-d6c4-48e7-bd85-c770273f1a75 pinIndex=0 msgBaseIndex=0 hash=9f270b2af4693565e0fcb57372122643fefcefd951885707547af6a64c758288 masked=true dbtx=xeSdlUgv ns=kld-company-one-testnet->testnet pid=1 role=aggregator
[2023-05-18T22:50:35.068Z] WARN No match for zerohash on context: group=6e9caaf7c5d70f1e73aa6215b9483401e4ab7aa63f3422dfa6e2ad45e03b0695 topic=delivery context=64b2d92f84090e6594acc432f15a5e43ea7e3f6842808f40992775a47170cba2 pin=9f270b2af4693565e0fcb57372122643fefcefd951885707547af6a64c758288 dbtx=xeSdlUgv ns=kld-company-one-testnet->testnet pid=1 role=aggregator
[2023-05-18T22:50:46.173Z] DEBUG Aggregating pin 0000000029 batch=5c317961-bb74-4234-a299-f72dee727720 msg=50cddd48-b6c2-449b-a243-f5e7b5c4ab6f pinIndex=1 msgBaseIndex=1 hash=babf8903ad7adbb6b602a190e16c663562cd69e4ae340931fb79dcf86aaa0be5 masked=true dbtx=wjpM0dKD ns=kld-company-one-testnet->testnet pid=1 role=aggregator
```

Because now batch processors are separated based on signing key, but contexts/pins are not (they only include the author DID), you can have two (or more) processors assembling batches against the same context. If the order of nonce assignment doesn't match the eventual order of batch dispatch, the pins can come in out of order.

This seems to be a changed behavior in the wake of #1175 (so original bug is changed but not entirely fixed).

awrichar:
This reverts #1175 and allows a single batch processor to process messages from the same identity but different keys. Instead, it adds a new fix to immediately seal and dispatch the current batch if a new message is picked up that has a different signing key than the current in-flight batch assembly.

Fixes #1311


internal/batch/batch_processor_test.go
```diff
@@ -423,6 +423,62 @@ func TestAddWorkBatchOfOne(t *testing.T) {
...
+		{msg: msg3},
+	}, bp.assemblyQueue)
+
+	full, overflow = bp.addWork(&batchWork{msg: msg2})
+	assert.True(t, full)
+	assert.True(t, overflow)
+	assert.Equal(t, []*batchWork{
+		{msg: msg1},
+		{msg: msg3},
+		{msg: msg2},
```
awrichar, line 478: I added this test to specifically verify and call out the ordering here. Because of the way the batch is broken up, the messages are not dispatched strictly in sequence order. I think this is OK but may need to double-check on assumptions about sequence on both sender and receiver.

awrichar: I _think_ the same scenario can happen even if all 3 messages were signed with the same key, but somehow `msg2` "misses the train" due to random timing and the batch goes out with just `msg1` and `msg3`. This means the order in which nonces are allocated will not match the order of the local message `sequence`. Just trying to assure myself that this is alright.

peterbroadhurst (approved):
Thanks for all the detail and thinking on this one 👍

peterbroadhurst (approved):
Thanks for all the detail and thinking on this one 👍

internal/batch/batch_processor_test.go
awrichar: There are other existing tests in this file (a few lines before this one) which simulate messages being added to a batch out of sequence order. I assumed this was because we dispatch messages randomly to a pool of message writers, and there can be a delay between when a message grabs a sequence number and when it's actually committed in a database transaction. Otherwise I'm not sure of the reason for the existing logic (and unit tests) around sorting messages as they get assembled into a batch.

peterbroadhurst: So here's the logic... and yes 👍 to exactly your explanation for the sequence ordering. Thanks for the detail, and you're right I did need to go back and look....

image

One note, is that the ordering only clicks in when you're writing messages _in parallel_. So on the FireFly core API you're not waiting for message 1 to be written, before message 2. So the fact that ordering in this scenario is not deterministic was accepted, because it can't be deterministic if it's parallel.

That's my recollection looking back at the diagram. Thanks for

PR #1315, opened 2023-05-23
feat: background start for connector plugins

EnriqueL8:


internal/blockchain/ethereum/ethereum.go
```diff
@@ -223,7 +236,37 @@ func (e *Ethereum) SetOperationHandler(namespace string, handler core.OperationC
 	e.callbacks.SetOperationalHandler(namespace, handler)
 }
 
+func (e *Ethereum) startBackground() {
+	batchSize := e.ethconnectConf.GetUint(EthconnectConfigBatchSize)
```
peterbroadhurst, line 240: (see comment above on setting config in init)

internal/blockchain/ethereum/ethereum.go
```diff
@@ -223,7 +236,37 @@ func (e *Ethereum) SetOperationHandler(namespace string, handler core.OperationC
...
+		stream, err := e.streams.ensureEventStream(e.ctx, e.topic, batchSize, batchTimeout)
+		if err != nil {
+			return true, err
+		}
+
+		e.streamID = stream.ID
+		log.L(e.ctx).Infof("Event stream: %s (topic=%s)", e.streamID, e.topic)
+
+		e.closed = make(chan struct{})
+		go e.eventLoop()
```
peterbroadhurst, line 253: Looks like we're starting the Go routine, before we've attempted the `e.wsconn.Connect` - so each time we go round the loop. But we're not ensuring it's cancelled if we fail. So I'm thinking it's possible we'll leak go routines.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -223,7 +236,37 @@ func (e *Ethereum) SetOperationHandler(namespace string, handler core.OperationC
...
+		if err != nil {
+			return true, err
+		}
+
+		return false, nil
+	})
+}
+
 func (e *Ethereum) Start() (err error) {
+	if e.ethconnectConf.GetBool(EthconnectBackgroundStart) {
```
peterbroadhurst, line 265: Same comment on config only in `Init()`

peterbroadhurst (commented):
This looks great @EnriqueL8!
Couple of minor suggestions from patterns we've found helpful in other code, and a query on goroutines.

internal/blockchain/ethereum/ethereum.go
EnriqueL8: Correct, great catch - we will run multiple go routines until it's successful and leak them! I was relying on the context being cancelled since each event loop will listen to that but not if we successfully go through, will make sure to amend that

EnriqueL8:
There seems to be flaky test I introduced - will fix that

internal/blockchain/ethereum/ethereum.go
```diff
@@ -199,13 +202,24 @@ func (e *Ethereum) Init(ctx context.Context, cancelCtx context.CancelFunc, conf
...
-	e.streams = newStreamManager(e.client, e.cache)
-	batchSize := ethconnectConf.GetUint(EthconnectConfigBatchSize)
-	batchTimeout := uint(ethconnectConf.GetDuration(EthconnectConfigBatchTimeout).Milliseconds())
-	stream, err := e.streams.ensureEventStream(e.ctx, e.topic, batchSize, batchTimeout)
+	e.streams = newStreamManager(e.client, e.cache, e.ethconnectConf.GetUint(EthconnectConfigBatchSize), uint(e.ethconnectConf.GetDuration(EthconnectConfigBatchTimeout).Milliseconds()))
+
+	e.backgroundStart = e.ethconnectConf.GetBool(EthconnectBackgroundStart)
+	if e.backgroundStart {
+		e.backgroundRetry = &retry.Retry{
+			InitialDelay: fftypes.ParseToDuration(defaultBackgroundInitialDelay),
```
peterbroadhurst, line 210: I think these should be config settings @EnriqueL8

EnriqueL8: Okay, I did think about it but wasn't 100% sure :) - will do it!

PR #1316, opened 2023-05-24
Derive batch signing key from messages in the batch

awrichar:
Follow-up to #1313

### DIFF
--- a/internal/batch/batch_processor.go
+++ b/internal/batch/batch_processor.go
@@ -43,7 +43,7 @@ type batchProcessorConf struct {
 	name           string
 	dispatcherName string
 	txType         core.TransactionType
-	signer         core.SignerRef
+	author         string
 	group          *fftypes.Bytes32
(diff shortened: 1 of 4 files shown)


internal/batch/batch_processor.go
```diff
@@ -399,9 +399,12 @@ func (bp *batchProcessor) initPayload(id *fftypes.UUID, flushWork []*batchWork)
 				ID:        id,
 				Type:      bp.conf.DispatcherOptions.BatchType,
 				Namespace: bp.bm.namespace,
-				SignerRef: bp.conf.signer,
-				Group:     bp.conf.group,
-				Created:   fftypes.Now(),
+				SignerRef: core.SignerRef{
+					Author: bp.conf.author,
+					Key:    flushWork[0].msg.Header.Key,
```
peterbroadhurst, line 404: 👍 - this is the key change

peterbroadhurst (approved):

Issue #1317, opened 2023-05-24
Token Pool namespace issues after dynamic config reload


nguyer:
Steps to reproduce:

- Enable dynamic config reload
- Add a namespace named space0 and save the config file. Ensure FireFly fully starts the namespace.
- Remove the namespace from the config file. Ensure FireFly fully stops the namesapce
- Add a new namespace named space1 and save the config file. Ensure FireFly fully starts the namespace.
- Attempt to create a token pool in space1

See that the token pool create event failed, and the error message is `FF10446: Namespace 'space0' is not started`

PR #1318, opened 2023-05-24
Unset handlers when namespaces are deleted (but not when they are just reloaded)

nguyer:
Proposed fix for https://github.com/hyperledger/firefly/issues/1317

There is a scenario this fix still *does not cover*:
- Load Namespace with plugin-set - including Tokens connector
- Reload Namespace, to use different set of plugins - removing the tokens connector

Now we've got the tokens connector with a handler, that's no longer required, and is pointing to a defunct `Orchestrator`.
If the token connector gets an event, it's not going to be able to dispatch it.

I assert this is a case we can fix either:
1. In a follow-on fix
2. When we move event listeners to be per-namespace


internal/orchestrator/orchestrator.go
```diff
@@ -244,6 +244,11 @@ func (or *orchestrator) Init() (err error) {
 	return err
 }
 
+func Purge(ctx context.Context, ns *core.Namespace, plugins *Plugins, dxNodeName string) {
```
awrichar, line 247: Any reason to not put this on the `*orchestrator` (I think it should be non-nil and have access to namespace/plugins/node name at the time this is needed, right?)

awrichar (approved):
Left one question, but this looks good!

PR #1329, opened 2023-05-30
Differentiate between org lookups that perform key resolution, and those that don't

awrichar:
Currently whenever we look up root org details in multiparty mode, we are also performing key resolution on the org's default signing key every time. This is only needed when we actually intend to use the org's key to sign something. In many instances, we actually do not need this (but are only interested in the org's name/DID).

I've adjusted two known instances in DefinitionHandler where we only want to lookup (and not resolve) the org. Still need to look back at the rest of the usages of the (renamed) `ResolveMultipartyRootOrg` method as well.


internal/orchestrator/status.go
```diff
@@ -92,7 +92,7 @@ func (or *orchestrator) GetStatus(ctx context.Context) (status *core.NamespaceSt
 		status.Org = &core.NamespaceStatusOrg{Name: or.config.Multiparty.Org.Name}
 		status.Multiparty.Contracts = or.namespace.Contracts
 
-		org, err := or.identity.GetMultipartyRootOrg(ctx)
+		org, err := or.identity.ResolveMultipartyRootOrg(ctx)
```
awrichar, line 95: Do we need key resolution here?

internal/privatemessaging/recipients.go
```diff
@@ -108,7 +108,7 @@ func (pm *privateMessaging) resolveNode(ctx context.Context, identity *core.Iden
 
 func (pm *privateMessaging) getRecipients(ctx context.Context, in *core.MessageInOut) (gi *core.GroupIdentity, err error) {
 
-	localOrg, err := pm.identity.GetMultipartyRootOrg(ctx)
+	localOrg, err := pm.identity.ResolveMultipartyRootOrg(ctx)
```
awrichar, line 111: Do we need key resolution here?

internal/identity/identitymanager.go
```diff
@@ -44,14 +44,15 @@ type Manager interface {
 	ResolveInputSigningKey(ctx context.Context, inputKey string, keyNormalizationMode int) (signingKey string, err error)
 	ResolveQuerySigningKey(ctx context.Context, inputKey string, keyNormalizationMode int) (signingKey string, err error)
 	ResolveIdentitySigner(ctx context.Context, identity *core.Identity) (parentSigner *core.SignerRef, err error)
+	ResolveMultipartyRootVerifier(ctx context.Context) (*core.VerifierRef, error)
+	ResolveMultipartyRootOrg(ctx context.Context) (*core.Identity, error)
```
peterbroadhurst, line 48: I'm questioning why we need this.
The only time we should need to do a `resolve` I believe, is when we need the signing key. But this call only returns the `core.Identity`

internal/identity/identitymanager.go
```diff
@@ -109,6 +110,13 @@ func (im *identityManager) GetLocalNode(ctx context.Context) (node *core.Identit
 	return node, err
 }
 
+func (im *identityManager) GetRootOrg(ctx context.Context) (org *core.Identity, err error) {
+	orgName := im.multiparty.RootOrg().Name
```
peterbroadhurst, line 114: While I tracked the code through to check `RootOrg()` can't return nil, it feels a bit odd to continue this check in the case that `Name` is `""`

internal/identity/identitymanager.go
awrichar: Yea, I looked at all the remaining paths that use this call, and they either 1) require the DID only (without key resolution) or 2) are actually performing key resolution twice, because this method performs the resolution and another path performs it again.

I'll remove this method altogether.

peterbroadhurst (approved):
👍

PR #1330, opened 2023-05-31
feat: TLS Configs for Webhooks

EnriqueL8:
Sitting in draft as there is some more extend testing that needs to be done but this is the general idea I was going for. 

Have tested it e2e with a local server secured with mTLS and it works.


go.mod
```diff
@@ -86,3 +86,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hyperledger/firefly-common => ../firefly-common
```
peterbroadhurst, line 90: Note that this needs to be removed before checkin

docs/reference/types/subscription.md
```diff
@@ -55,6 +55,7 @@ nav_order: 3
 | `transport` | The transport plugin responsible for event delivery (WebSockets, Webhooks, JMS, NATS etc.) | `string` |
 | `filter` | Server-side filter to apply to events | [`SubscriptionFilter`](#subscriptionfilter) |
 | `options` | Subscription options | [`SubscriptionOptions`](#subscriptionoptions) |
+| `tlsConfigName` | The name of an existing TLS configuration associated to the namespace to use when performing the request in the case of webhooks | `string` |
```
peterbroadhurst, line 58: I think this is great 👍

peterbroadhurst: I believe this should be moved down to the Webhook specific config though, rather than at this level:
https://hyperledger.github.io/firefly/v1.2.0/reference/types/subscription.html#webhookinputoptions

`WebhookInputOptions` is the object

internal/database/sqlcommon/subscription_sql.go
```diff
@@ -122,6 +124,7 @@ func (s *SQLCommon) UpsertSubscription(ctx context.Context, subscription *core.S
 					subscription.Options,
 					subscription.Created,
 					subscription.Updated,
+					subscription.TLSConfigName,
```
peterbroadhurst, line 127: See earlier comments about this being part of the Webhook specific options - not sure we need a DB migration in that case.

internal/orchestrator/subscriptions.go
```diff
@@ -47,6 +47,10 @@ func (or *orchestrator) createUpdateSubscription(ctx context.Context, subDef *co
 		return nil, i18n.NewError(ctx, coremsgs.MsgSystemTransportInternal)
 	}
 
+	if subDef.TLSConfigName != "" || or.namespace.TLSConfigs[subDef.TLSConfigName] != nil {
```
peterbroadhurst, line 50: Note comments above about moving this down into Webhook specific code.

internal/namespace/manager.go
```diff
@@ -850,6 +880,16 @@ func (nm *namespaceManager) loadNamespace(ctx context.Context, name string, inde
 		}
 	}
 
+	//
+	// Handle TLS Configs
+	tlsConfigArray := conf.SubArray(coreconfig.NamespaceTLSConfigs)
+	tlsConfigs := make(map[string]*tls.Config)
+
+	err = nm.loadTLSConfig(ctx, tlsConfigs, tlsConfigArray)
```
peterbroadhurst, line 888: Think it's great this is happening here. I believe (and would like to validate) this means:
- This happens automatically on each dynamic reload
- If we fail because we have bad TLS config for this namespace, then **only** this namespace will fail to init

docs/reference/types/subscription.md
EnriqueL8: Sure - will do that

internal/namespace/manager.go
EnriqueL8: Correct - will add a test for the failure case

EnriqueL8:
I personally think we should keep the TLS Config as it is as we will need it for other type of subscriptions:
- Websockets perform an HTTP upgrade before opening the socket so that needs mTLS
- NATS/Kafka, any other event broker technology will also require certificates.

Keeping the certificates at the subscription level allows each package for the respective technology to use that config as they wish.

internal/events/subscription_manager.go
```diff
@@ -249,6 +249,7 @@ func (sm *subscriptionManager) deletedDurableSubscription(id *fftypes.UUID) {
 	}
 }
 
+// nolint: gocyclo
```
peterbroadhurst, line 252: The linter may have a point :)

Do think it's fine to merge though.

internal/events/webhooks/webhooks.go
```diff
@@ -224,7 +225,18 @@ func (wh *WebHooks) attemptRequest(sub *core.Subscription, event *core.EventDeli
 		}
 	}
 
-	req, err = wh.buildRequest(sub.Options.TransportOptions(), firstData)
+	// Taking a copy of the global plugin client because
+	// 1) We do not want to modify that global instance
+	// 2) We want to keep the global configuration for webhooks
+	httpClient := *wh.client.GetClient()
```
peterbroadhurst, line 231: Are we sure that this is a safe copy?
Feels a little dangerous to be assuming that the changes we're making are not modified sub-structures of `resty` (as it exists now or in the future) that are pointer referenced.

The `ffresty` object has the ability to build a config, which is from the Viper config that goes into a Go structure.
It might be safer to construct the client from that with the TLS options modified, or each subscription.

internal/namespace/manager.go
peterbroadhurst: Ok - I really want to be engaged in the discussion if/when we make a change to ^^^ - as I think it's subtle where the boundary is between bug and intentional safe behavior... and I've been iterating a lot in that area of dynamic reload on that.

Let's merge with the behavior as it is, and treat any change as a separate PR.

internal/orchestrator/subscriptions.go
```diff
@@ -47,6 +47,15 @@ func (or *orchestrator) createUpdateSubscription(ctx context.Context, subDef *co
 		return nil, i18n.NewError(ctx, coremsgs.MsgSystemTransportInternal)
 	}
 
+	if subDef.Options.TLSConfigName != "" {
```
peterbroadhurst, line 50: Is this defunct now we've moved to WS config?

peterbroadhurst (changes requested):
This is looking great @EnriqueL8 
A couple of requests for investigation, and I hope avoiding the `struct`-level copy on the resty client isn't a big lift.

internal/events/webhooks/webhooks.go
EnriqueL8: I originally went the second way but it meant dealing with config merges which I thought would be more complex. The ffresty config section comes from the `events.webhooks` root section but we have the TLS Config section under `namespaces.predefined[].tlsConfigs[].tls but I guess as you say I could store the `config.Section` the Webhook object create a new resty client and then call `newClient.SetTLSClientConfig`


I agree that it can be dangerous and the deep copy is not guaranteed. I couldn't find anywhere where someone takes a deep copy of an http client

internal/orchestrator/subscriptions.go
EnriqueL8: WebhookSubOptions are just inherited in Options as such  https://github.com/hyperledger/firefly/blob/main/pkg/core/subscription.go#L100 so we can reference them directly

internal/events/webhooks/webhooks.go
peterbroadhurst: > we could optimise this and store the client in the subscription object

That would definitely be an option to change in this code, that would let me feel comfortable to merge.
e.g. we construct from config uniquely for each subscription, and don't try to do any object copying.

The pattern I was referring to is here:
https://github.com/hyperledger/firefly-common/blob/86d110a4e827b03aeca0723218723c52f70e5574/pkg/wsclient/wsconfig.go#L59-L82

I had it in my head that we'd done it for `ffresty` as well as WebSockets, but it seems not.

internal/events/webhooks/webhooks.go
```diff
@@ -224,7 +229,17 @@ func (wh *WebHooks) attemptRequest(sub *core.Subscription, event *core.EventDeli
 		}
 	}
 
-	req, err = wh.buildRequest(sub.Options.TransportOptions(), firstData)
+	// Create a new ffresty client from the config
+	// 1) We do not want to modify that global instance
+	// 2) We want to keep the global configuration for webhooks
+	copyFFrestyConfig := *wh.ffrestyConfig
```
peterbroadhurst, line 235: Thanks for working this all through @EnriqueL8 🙇 
Really pleased with where this landed.

peterbroadhurst (approved):
👍

PR #1331, opened 2023-06-01
Process batch of events from Blockchain connector, in a single DB TX

peterbroadhurst:
One of the outstanding observations from the performance testing in V1.2 was that we are performing a DB commit for each Blockchain event.

These are sequential on the single logical delivery thread from the blockchain connector (which currently is server-wide, although that should itself become namespace-wide).

This PR proposes that instead we update all the internal callback interfaces, to propagate the batch of events that come in over the wire from the connector, through to the event processor.

This means a few things handled in this PR:
1. Handling a batch that contains events of multiple types -  Batch Pin, Network Action, or custom Blockchain Event Listener
2. Handling a batch that needs different events dispatching to different namespaces
3. Continuing to handle old events that didn't come with a namespace at all (these are dispatched to all registered namespaces)

### DIFF
--- a/internal/blockchain/common/common.go
+++ b/internal/blockchain/common/common.go
@@ -56,7 +56,7 @@ type FireflySubscriptions interface {
 }
 
 type callbacks struct {
-	writeLock  sync.Mutex
+	lock       sync.RWMutex
 	handlers   map[string]blockchain.Callbacks
 	opHandlers map[string]core.OperationCallbacks
 }
@@ -110,8 +110,8 @@ func NewFireflySubscriptions() FireflySubscriptions {
 }
 
 func (cb *callbacks) SetHandler(namespace string, handler blockchain.Callbacks) {
-	cb.writeLock.Lock()
-	defer cb.writeLock.Unlock()
+	cb.lock.Lock()
+	defer cb.lock.Unlock()
 	if handler == nil {
 		delete(cb.handlers, namespace)
 	} else {
@@ -120,8 +120,8 @@ func (cb *callbacks) SetHandler(namespace string, handler blockchain.Callbacks) 
 }
 
 func (cb *callbacks) SetOperationalHandler(namespace string, handler core.OperationCallbacks) {
-	cb.writeLock.Lock()
-	defer cb.writeLock.Unlock()
+	cb.lock.Lock()
+	defer cb.lock.Unlock()
 	if handler == nil {
 		delete(cb.opHandlers, namespace)
 	} else {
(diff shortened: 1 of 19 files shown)


internal/events/blockchain_event.go
```diff
@@ -99,32 +99,51 @@ func (em *eventManager) emitBlockchainEventMetric(event *blockchain.Event) {
...
+				switch event.Type {
+				case blockchain.EventTypeForListener:
+					if err := em.handleBlockchainEventForListener(ctx, event.ForListener); err != nil {
+						return err
+					}
+				case blockchain.EventTypeBatchPinComplete:
+					if err := em.handleBlockchainBatchPinEvent(ctx, event.BatchPinComplete); err != nil {
+						return err
+					}
+				case blockchain.EventTypeNetworkAction:
```
Chengxuan, line 115: @peterbroadhurst I spent some time trying to figure out why the test is failing for sqllite3. After reading the code and test. The test was failing on waiting for the NetworkAction event `firefly:terminate` but didn't get it.

Looking at the changes made in this PR, the old logic for emitting network action event was not under "em.database.RunAsGroup" but the new logic does. Could this be the issue for the failing test?

I did notice the same test is working fine with postgres run

peterbroadhurst: Possible we've ended up with a nested `RunAsGroup` - I shall dig into it...

Chengxuan: @peterbroadhurst that explains why postgres worked. Thanks for sharing the details here. 🙏

internal/blockchain/common/common.go
```diff
@@ -177,48 +188,74 @@ func (cb *callbacks) BatchPinOrNetworkAction(ctx context.Context, subInfo *Subsc
...
 
-func (cb *callbacks) batchPinComplete(ctx context.Context, namespaces []string, batch *blockchain.BatchPin, signingKey *core.VerifierRef) error {
+func (cb *callbacks) addBatchPinComplete(ctx context.Context, events EventsToDispatch, namespaces []string, batch *blockchain.BatchPin, signingKey *core.VerifierRef) {
 	for _, namespace := range namespaces {
-		if handler, ok := cb.handlers[namespace]; ok {
-			if err := handler.BatchPinComplete(namespace, batch, signingKey); err != nil {
-				return err
-			}
+		if _, ok := cb.handlers[namespace]; ok {
+			events[namespace] = append(events[namespace], &blockchain.EventToDispatch{
```
nguyer, line 197: Do we need locks around this and the other places we append to the map+list? Maybe this will only be required when we make each namespace have its own event listener?

peterbroadhurst: Great point - checking 👍

nguyer (approved):
Overall looks good.

nguyer (approved):

Issue #1304, opened 2023-05-11
getIds { output: null }


agiratech-vignesh-m:
I suspect there is a bug in the FireFly signer ABI encoding. This contract demonstrates the issue: https://mumbai.polygonscan.com/address/0xc0bc785b2c5494820B63c7108547b4652b9742aD#code

Steps to reproduce:

Create a new FireFly stack (using default settings with a local blockchain node is fine)
Create a contract interface / API with the contract listed above
Call the POST /invoke/createCoupon endpoint with the following payload:

{
  "input": {
    "_end": "1234",
    "_id": "2345",
    "_ipfsUrl": "3456",
    "_start": "4567"
  }
}

Call the POST /query/allCouponIds with an empty JSON object as the request body
Response:
We will get an response as : getIds { output: null }, even i tried with other functions receiving the same error.
On Polygonscan I see a valid response from this method.

joshtharakan:
Received the same issue and reported here -  https://github.com/hyperledger/firefly/issues/1303#issuecomment-1552962333

EnriqueL8:
I've taken a look at this one and figured out the problem

When a contract annotates the returning values in the tuple we decode correctly and construct a map with the keys being the annotated name in this case map[string]interface{} where
```json
{
  "allCouponIds": [2345]
}
```
Yet the transaction manager is expecting an output field and the result nested under it. If it's not there it cannot unmarshall/decode it and the result is null!
```json
{
 "output": {
   "allCouponIds": ....
 }
}
```
The way we add this output nesting  here https://github.com/hyperledger/firefly-signer/blob/038e29f0aeb724ab5b716a772ff839e2419133d2/pkg/abi/outputserialization.go#L226 if the children component doesn't have a key name we set it to output and serialise everything under it but we do have a key name for each children so we never add this output nesting.
A possible solution would be to use the TypeComponent key names instead of the TupleChildren key names but appreciate any other ideas

PR #1333, opened 2023-06-06
fix: multiple named tuple result from contract

EnriqueL8:


peterbroadhurst (commented):
@EnriqueL8 / @nguyer - this seems to be a fundamental change/enhancement to the API surface area of FireFly core. Think we need to be really clear in what the behavior is today, what this change means, and where this is/should-be documented in the FireFly docs.

Would you mind adding some commentary to that effect?

internal/blockchain/ethereum/ethereum.go
```diff
@@ -836,11 +836,25 @@ func (e *Ethereum) QueryContract(ctx context.Context, signingKey string, locatio
 	if err != nil || !res.IsSuccess() {
 		return nil, err
 	}
-	output := &queryOutput{}
-	if err = json.Unmarshal(res.Body(), output); err != nil {
+
+	queryOutput := make(map[string]interface{})
+	if err = json.Unmarshal(res.Body(), &queryOutput); err != nil {
```
peterbroadhurst, line 841: If I understand correctly, the proposal here is:
- If the ABI has a single **unnamed** parameter, then we return `output: {whatever}`
- If the ABI has a single **named** parameter, then we return `output: {name1: {whatever}}`
- If the ABI has multiple **unnamed** parameters, then we return `output: {output: {whatever1}, output1: {whatever2}}`
- If the ABI has multiple **named** parameters, then we return `output: {name1: {whatever1}, name2: {whatever2}}`

EnriqueL8:
This PR changes the contract between Firefly core and the ethereum blockchain connector (EVMConnect and ETHConnect) by relaxing the schema validation when reading the result of a contract query operation.

Before this PR, FF core relied on the result being in the format `"output": {}` now it can accept any arbitrary map of string to any (interface{}). Yet as the firefly interface nests the result in an `"output"` we still do it but in ff instead

internal/blockchain/ethereum/ethereum.go
EnriqueL8: Almost there just this one is different:
-  If the ABI has multiple unnamed parameters, then we return `output: {whatever1}, output1: {whatever2}` instead of it being nested. 

Discussed this one with @Chengxuan and thought it would be better as above but happy to change

peterbroadhurst: ok, so the logic is "if any of the data comes back from the connector contains `output` as a field, return the whole output unchanged

peterbroadhurst: ... otherwise, nest whatever came back from the connector under an `output` parent object

EnriqueL8: yes

EnriqueL8: Fabric just goes with the approach of proxying data https://hyperledger.github.io/firefly/head/tutorials/custom_contracts/fabric.html#request-3

EnriqueL8: I'm worried about proxying the data across since we would potentially break users but actually they are already broken by this!

If we proxy:

If the ABI has a single unnamed parameter, then we return `{output: {whatever}}`
If the ABI has a single named parameter, then we return `{name1: {whatever}}`
If the ABI has multiple unnamed parameters, then we return `{output: {whatever1}, output1: {whatever2}}`
If the ABI has multiple named parameters, then we return` {name1: {whatever1}, name2: {whatever2}}`

EnriqueL8: That Fabric doc must be out of date then! 

It's still the same problem I believe. We are decoding to an object that expects a key "Result" https://github.com/hyperledger/firefly/blob/e2f00f021e1abca467532ca6dbd4f8a804091827/internal/blockchain/fabric/fabric.go#L102 so the fabric connector needs to return that. Either the fabric connector is incapsulating everything under that result field or it would hit the same error with multiple values

EnriqueL8: I will amend this PR to proxy the direct output for ethereum connector and we can discuss the fabric one as well and decide in a future PR - need to look at the fabric connector code as well

EnriqueL8: Fabric construct that object after serialisation https://github.com/hyperledger/firefly-fabconnect/blob/b46951bb6613a2981b89c3ea2e69048441ec41ed/internal/rest/sync/syncdispatcher.go#L200-L203 completely different to ethereum where we add a custom name generator to override the serialisation

peterbroadhurst: > so the fabric connector needs to return that

Yes, I agree `result` is where Fabric connector returns it. But that's just an `interface{}`.

e.g. - in Ethereum **today** which is broken:

- The connector returns the result _at the top level_ of the JSON, and the result must have exactly one object we care about called `output`

- The connector returns the result _in a sub-field_ of the JSON called `result`, and can specify anything it likes there which we will return _at the top level_ of the JSON to the end user

peterbroadhurst: So, I think if you change Ethereum as follows, it will be very consistent with Fabric:

```go
	var output interface{}
	if err = json.Unmarshal(res.Body(), &output); err != nil {
		return nil, err
	}
	return output, nil // note UNLIKE fabric this is just `output`, not `output.Result` - but either way the top level of what we return to the end user, is whatever the Connector sent us
```

internal/blockchain/ethereum/ethereum.go
```diff
@@ -836,10 +836,12 @@ func (e *Ethereum) QueryContract(ctx context.Context, signingKey string, locatio
 	if err != nil || !res.IsSuccess() {
 		return nil, err
 	}
-	output := &queryOutput{}
-	if err = json.Unmarshal(res.Body(), output); err != nil {
+
+	output := make(map[string]interface{})
```
peterbroadhurst, line 840: @EnriqueL8 - I believe this should just be `interface{}` (not a map)

> arbitrary map of string to any (interface{}

I agree that's what EVMConnect is configured to return today. But in `firefly-signer` there are other options (possibly not exposed in config on EVMConnect _yet_).

For example, you can make it work just like Fabric where it returns an array (not an object) of the values.

I don't see any reason why we'd want to make a change now in Core that precludes EVMConnect being configured by someone to do that in the future.

nguyer:
Resolves https://github.com/hyperledger/firefly/issues/1304

peterbroadhurst (approved):
👍

PR #1336, opened 2023-06-07
Surface message rejection reason to API

awrichar:
Proposal to add a "reject_reason" to messages (ie to database and API), which will be populated with the error text whenever a message is rejected. Previously this reason would only be logged, so the hope is that this assists in debugging when a message is rejected for an unexpected reason.

### DIFF
--- a/internal/events/aggregator.go
+++ b/internal/events/aggregator.go
@@ -26,10 +26,12 @@ import (
 	"github.com/hyperledger/firefly-common/pkg/config"
 	"github.com/hyperledger/firefly-common/pkg/ffapi"
 	"github.com/hyperledger/firefly-common/pkg/fftypes"
+	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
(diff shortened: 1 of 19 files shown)


internal/events/aggregator.go
```diff
@@ -618,10 +622,6 @@ func (ag *aggregator) readyForDispatch(ctx context.Context, msg *core.Message, d
 		var handlerResult definitions.HandlerResult
 		handlerResult, err = ag.definitions.HandleDefinitionBroadcast(ctx, &state.BatchState, msg, data, tx)
 		log.L(ctx).Infof("Result of definition broadcast '%s' [%s]: %s", msg.Header.Tag, msg.Header.ID, handlerResult.Action)
-		if handlerResult.Action == core.ActionReject {
-			log.L(ctx).Infof("Definition broadcast '%s' rejected: %s", msg.Header.ID, err)
-			err = nil
```
awrichar, line 623: Note that this block should now fall through and return the error to its caller (line 528), which will then log the error and add it to the message, but should not cause retries. I'm trying to be very careful not to reintroduce the bugs fixed in https://github.com/hyperledger/firefly/pull/1278.

PR #1340, opened 2023-06-12
Implement API add namespace route

ZeljkoBenovic:
## Description 
This PR implements the SPI route for adding new namespaces with the appropriate plugins.
I saw that this implementation was missing, even though the [description](https://github.com/hyperledger/firefly/blob/2fbd5b68c09dc4fb00746c93b1c3a5cd9e8b18f3/internal/coremsgs/en_api_translations.go#L175) is already created, but it is describing the wrong [route](https://github.com/hyperledger/firefly/blob/2fbd5b68c09dc4fb00746c93b1c3a5cd9e8b18f3/internal/apiserver/route_spi_get_namespace_by_name.go#LL35C28-L35C56).

Since there is already a method that dynamically reloads the config file, I've used the same mechanism to reload it when the content is changed after this route call.     
It seams to me that this is the easy and clean way to implement this feature. Please, do let me know if this should be done in another, better way.

## Manual tests
Run the binary or a container, open the SPI Swagger interface and you should see the new POST route `spi#/Global/spiPostAddNamespaces`. The POST body should be as the following example: 
```
{
  "defaultKey": "0x36acbe2c5e1961552fd27f4e93723f3e8f6e778d",
  "description": "New customer",
  "name": "customer1",
  "plugins": [
    "database0",
    "blockchain0",
    "erc20_erc721"
  ]
}
```
After the call, run another GET call to verify that the new namespace has been successfully created.

## Comment
The start up that I'm working with, really needs this feature, so for now we're working of this branch. 
We have a use case where each new namespace should run on a dedicated private chain.
But we're hoping that this will be adopted into the main codebase so that we don't need to maintain our own fork.


nguyer:
Hi @ZeljkoBenovic. Thank you for opening this PR! I'm Really glad that FireFly is a useful project to you and that you're able to use multiple namespaces to connect to multiple blockchains.

Some background on the journey of namespaces might be helpful here. In FireFly v1.0 we did have an endpoint to add a new Namespace. However, at that point in time, all namespaces used the same plugins, so a namespace wasn't something that actually needed to be part of the config file. It was just a row in the DB. As FireFly has matured, the concept of a Namespace has grown more powerful. In FireFly v1.1 we added the ability to define a list of plugins and a list of namespaces that use some of those plugins. A plugin can also be used by more than one namespace. At this point we transitioned namespaces to be defined in the config file because they were suddenly much more powerful, but also needed more complex configuration. Thus, the endpoint to create a new namespace was deprecated.

As maintainers, our current view is that FireFly Core should not write its own config file, for several reason:

- It is potentially dangerous. Problems in writing the config file have the potential to bring down the entire runtime, including namespaces that were not intended to be updated. This could stem from bugs in the code that generates the config or problems writing to the filesystem.
- It introduces more code complexity in needing to know how to both read and write the config file. Currently both YAML and JSON config files are supported, along with the ability to override config with environment variables. All of this is enabled by our config package we've built that uses Viper to actually take all the config in.
- Lastly, and most significantly, FireFly Core writing its own config file is not compatible with most production deployments of FireFly. In a typical production deployment, FireFly Core is running in a Kubernetes pod and has its config mounted from a Kubernetes secret. From the perspective of the pod, this "file" is read-only. One of the major upcoming features is Active-Active High Availability. In an Active-Active cluster, all FireFly runtimes need to use the same config file, which means they can't be writing it themselves.

That being said, this code looks like a great start to a tool that could manage namespace config, and potentially could be worked into https://github.com/hyperledger/firefly-helm-charts and drive that Kubernetes based configuration. This could achieve the same end-result, an API to update FireFly's config, but in a cloud-native approach that aligns with FireFly's roadmap.

I hope this background was helpful in understanding why these changes don't align with FireFly's design and roadmap. It is great code though! I'm happy to discuss further if you want to talk about it, or have more questions. Thanks!

ZeljkoBenovic:
Hi @nguyer ,
Thank you for this extensive explanation.      
I figured that it is probably a design choice, but it was worth a shot. And I totally agree with you guys.
I've been pushing for K8s infrastructure for our little start up, but the management decided that that is too much overhead right now.     

As I could not find anything in the docs, what would be the FF adoption path outside of K8s context. 
What would be a recommended way of making changes to the FF configuration?    
Would the community benefit from a standalone API service, that would handle this, like you sad?
Also, should it be a standalone service, or a part of `firefly-cli` repo? Like there could be a `listen` command for example, that would open an API that modifies the config file?    
To accommodate FF into our use case, we'll need to create some kind of API anyways, that deals with config file, and my train of thought is, should we try to implement it as a part of `firefly-cli` so that the community can benefit from such a feature as well, or should we make a standalone repo?

Lastly, I just wanted to say that your code base is exemplary, very neat, clean, understandable and the 100% code coverage is a sight to behold 🥇  I wish that there were more repos out there what have your kind of standards. Keep it up.

nguyer:
Thank you @ZeljkoBenovic, I appreciate the kind words!

I think what you are describing would fit outside of the FireFly CLI as it stands today. It has a different set if responsibilities. Although I could also imagine a completely re-written FireFly CLI that could have a broader scope and cover some of these things as well.

Production deployments outside of K8s aren't really something that FireFly has prioritized. Right now K8s is really the best option to manage all of the infrastructure and microservices that are part of FireFly. If something better comes along, we will definitely take a look at that. But we just don't have cycles to invest in something that is not the best way to do it. Kubernetes solves a ton of really hard problems. It's just much better to build on top of proven, industry standard, open source platforms rather than try to re-invent them ourselves. Hope that makes sense.

Thanks again!

Nicko

PR #1341, opened 2023-06-13
Automatically check operation status when querying pending transaction

nguyer:
This PR makes a slight change when querying a transaction. If any operation is still in "Pending", do a lookup on that operation in case it has actually had an update that we missed previously.


peterbroadhurst (approved):
👍

PR #1345, opened 2023-06-19
FabConnect does not / can not set transactionIndex and eventIndex

peterbroadhurst:
Fabric chaincode events when dispatched to FabConnect by the Fabric SDK do not contain the index within the block:
https://github.com/hyperledger/fabric-sdk-go/blob/7af45cede6afa3939a9574bc9948cca9fb424257/pkg/common/providers/fab/eventservice.go#L43C1-L59

So the only information upon which to build a unique `ProtocolID` for the event is the block number, and the transaction index (noting Hyperledger Fabric has a restriction of one transaction per block).

Under load when multiple events are packed into a single block in Fabric, the FireFly Core code was de-duplicating them as a single event, because they all had the same `ProtocolID`, such as `000000000295/000000/000000` (where the second and third parts are always zeros).

### DIFF
--- a/internal/blockchain/fabric/fabric.go
+++ b/internal/blockchain/fabric/fabric.go
@@ -355,10 +355,15 @@ func (f *Fabric) parseBlockchainEvent(ctx context.Context, msgJSON fftypes.JSONO
 		return nil // move on
 	}
 
+	// Fabric events are dispatched by the underlying fabric client to FabConnect with just the block number
+	// and the transaction hash. The index of the transaction in the block, or the index of the action within
+	// the transaction are not available. So we cannot generate an alphanumerically sortable string
+	// into the protocol ID. Instead we can only do this (which is according to Fabric rules assured to be
+	// unique, as Fabric only allows one event per transaction):
 	sTransactionHash := msgJSON.GetString("transactionId")
 	blockNumber := msgJSON.GetInt64("blockNumber")
-	transactionIndex := msgJSON.GetInt64("transactionIndex")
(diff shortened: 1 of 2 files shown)


internal/blockchain/fabric/fabric.go
```diff
@@ -355,10 +355,15 @@ func (f *Fabric) parseBlockchainEvent(ctx context.Context, msgJSON fftypes.JSONO
...
+	// Fabric events are dispatched by the underlying fabric client to FabConnect with just the block number
+	// and the transaction hash. The index of the transaction in the block, or the index of the action within
+	// the transaction are not available. So we cannot generate an alphanumerically sortable string
+	// into the protocol ID. Instead we can only do this (which is according to Fabric rules assured to be
+	// unique, as Fabric only allows one event per transaction):
 	sTransactionHash := msgJSON.GetString("transactionId")
 	blockNumber := msgJSON.GetInt64("blockNumber")
-	transactionIndex := msgJSON.GetInt64("transactionIndex")
-	eventIndex := msgJSON.GetInt64("eventIndex")
+	protocolID := fmt.Sprintf("%.12d/%s", blockNumber, sTransactionHash)
```
awrichar, line 365: Seems like a somewhat significant change for `protocolID` to no longer provide a sorting guarantee. I don't know that anyone was relying on it to be sortable, but we should be sure we don't describe it as (necessarily) being such.

peterbroadhurst: Agreed - will mark this as a migration consideration for 1.3

PR #1348, opened 2023-06-21
Docs improvements

ssmirr:
Adding two improvements in docs:

- adding a note for macOS users (port 5000 already in use)
- updating the command to be consistent with previous page (created `dev` stack but next page started `demo`)


shorsher (approved):
thank you!

PR #1351, opened 2023-06-22
Use NetworkName instead of Name for definition topics

### LINKED ISSUE #1349: Definition broadcasts are not using unique topics
Looks to have been introduced by the enhancements for https://github.com/hyperledger/firefly/issues/1220.

The topic for definition messages is generated from a few items in the definition - normally the object type, namespace, and name:
https://github.com/hyperledger/firefly/blob/fbd88de171e513eb6074ee5a6ae440a6c3f60acc/pkg/core/tokenpool.go#L96

However, for definitions that have a "network name" now, the local "name" is actually cleared before broadcasting. Therefore the topic for the broadcast should probably be derived from the network name instead of the name (otherwise all definitions with that namespace/type will end up with an identical topic).

awrichar:
Fixes #1349

Related fix that will also need to be integrated: https://github.com/hyperledger/firefly-common/pull/76

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.15.2
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.5.0
-	github.com/hyperledger/firefly-common v1.2.15
+	github.com/hyperledger/firefly-common v1.2.17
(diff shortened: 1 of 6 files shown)


peterbroadhurst:
https://github.com/hyperledger/firefly-common/releases/tag/v1.2.17 release created for the common pre-req

peterbroadhurst (approved):

PR #1353, opened 2023-06-23
feat: expose retry and HTTP options for webhooks

EnriqueL8:
A few considerations when reviewing:
- I modified the interface for the events plugin to pass the context through as we were using the global plugin context for events which I believe was incorrect.
- Instead of copying the client for each request in the webhook, we store the client in the subscription object. We populate it when reading a subscription from the DB or creating a new one through the API. I thought here about exposing a new function at the interface level such as `ParseSubscriptionOptions` where each plugin could parse the options and modify it but seem redundant do `ValidateOptions` so did it there instead


internal/coremsgs/en_struct_descriptions.go
```diff
@@ -681,22 +681,34 @@ var (
...
+	WebhooksOptInputQuery               = ffm("WebhookInputOptions.query", "A top-level property of the first data input, to use for query parameters")
+	WebhooksOptInputHeaders             = ffm("WebhookInputOptions.headers", "A top-level property of the first data input, to use for headers")
+	WebhooksOptInputBody                = ffm("WebhookInputOptions.body", "A top-level property of the first data input, to use for the request body. Default is the whole first body")
+	WebhooksOptInputPath                = ffm("WebhookInputOptions.path", "A top-level property of the first data input, to use for a path to append with escaping to the webhook path")
+	WebhooksOptInputReplyTx             = ffm("WebhookInputOptions.replytx", "A top-level property of the first data input, to use to dynamically set whether to pin the response (so the requester can choose)")
+	WebhooksOptRetryEnabled             = ffm("WebhookRetryOptions.enabled", "A top-level property of the first data input, to use to dynamically set whether to pin the response (so the requester can choose)")
```
peterbroadhurst, line 702: Possibly a copy/paste issue?

peterbroadhurst (approved):
Looks great to me @EnriqueL8 
Just one config doc item before merging

internal/coremsgs/en_struct_descriptions.go
EnriqueL8: oh yes - great spot, will fix that up

PR #1356, opened 2023-06-27
Remove token pools from the cache upon deletion

### LINKED ISSUE #1355: Token pools are not removed from cache when deleted
This seems to be a result of #1261 and #1291 getting merged very close together. Token pools are not cleared from the cache when they are deleted, so cached calls may continue to return old pool details.

awrichar:
Fixes #1355

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/hyperledger/firefly-common v1.2.17
 	github.com/hyperledger/firefly-signer v1.1.8
 	github.com/jarcoal/httpmock v1.2.0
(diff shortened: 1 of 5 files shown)


matthew1001 (approved):
LGTM

Chengxuan:
Does look like we are repeating logic in the firely-common library. Will raise a PR to fix this outstanding debt.

Chengxuan:
@awrichar @matthew1001 PR raised to move to common lib https://github.com/kaleido-io/firefly/pull/71

Chengxuan (approved):

shorsher (approved):

PR #1359, opened 2023-06-29
feat: batching events and webhook plugin support

EnriqueL8:
Enable batching for subscriptions

- [x] Write more robust tests to match coverage
- [x] Handle the firstData for events and decide what we do there
- [x] Refactor the webhook plugin to make it more modular instead of copying code across functions


EnriqueL8:
PR not ready for review as I need to work on removing the `nack` I added and test the infinite retry is now working

internal/events/webhooks/webhooks.go
```diff
@@ -444,6 +444,168 @@ func (wh *WebHooks) doDelivery(ctx context.Context, connID string, reply bool, s
 	}
 }
 
+func (wh *WebHooks) attemptBatchRequest(ctx context.Context, sub *core.Subscription, events []*core.EventDelivery, data []core.DataArray) (req *whRequest, res *whResponse, err error) {
```
peterbroadhurst, line 447: Does this not need to be an array of event+data elements? (rather than separate arrays of events, and data)

I believe in the `batch=true` case, the payload of each JSON object inside the array should be identical to the payload of the whole `POST` payload in the `batch=false` ase

peterbroadhurst: I guess key point from me, is I expect to see exactly one function that implements this logic, and the surrounding function to be broken apart to either package a single one of these into a HTTP request, or an array of them:

https://github.com/hyperledger/firefly/blob/816463cc5686d0fdc96241a9b6a41fe57e626003/internal/events/webhooks/webhooks.go#L324-L340

EnriqueL8: I see , I can merge both which makes sense but I can't modify the request components that are shared such as the headers, path, query parameters, etc... 

So the input mapping with the firstData which is done today will only be able to be done once

peterbroadhurst (approved):
@EnriqueL8 - looked great - I've filled in a couple of commits to simplify a bit of duplicate logic, and found an efficiency in deferring parsing of some data from string to JSON.

Also made the change to prevent creation of subscriptions with batching enabled, for event transports that don't support it.

internal/events/webhooks/webhooks.go
peterbroadhurst: Think tis is all resolved now 👍

peterbroadhurst:
Popped back to draft as there's a potential issue with creating subscriptions with the default transport

EnriqueL8:
Tested the above fix from Peter and all good you can create a subscription with a transport and it defaults to web hooks

Issue #1362, opened 2023-07-06
Docs Could Use Some SEO Improvements


keeganhrobinson:
I noticed that the docs site could use some SEO improvements. Some of the items are that there's currently two page titles on each page instead of one, all pages have the same meta description, most pages have an odd page title like page.home, and Google Analytics isn't hooked up.

The main cause of most the issues seems to be that [jekyll-multiple-languages-plugin](https://github.com/kurtsson/jekyll-multiple-languages-plugin/) and [jekyll-seo-tag](https://github.com/jekyll/jekyll-seo-tag) don't play nicely together. I'm working on a fix where only the jekyll-multiple-languages-plugin is used and then the theme handles outputting the page title and various meta tags.

![Duplicate Page Titles](https://github.com/hyperledger/firefly/assets/12145278/f0388cfd-e3a5-4e3f-acd1-1181a36a9a65)

PR #1363, opened 2023-07-06
Docs SEO Updates

keeganhrobinson:
@nguyer this pull request should address the items I've listed in issue #1362. It removes the output from the jekyll-seo-tag plugin and instead uses only the theme to add the page title and various meta tags. It includes custom page titles and meta descriptions for all of the docs pages so they'll be better crawled and displayed by search engines. And last, it hooks up Google Tag Manager so we'll be able to have Google Analytics for the site and see what kind of traffic it's getting. Let me know if you see anything you'd like me to adjust.


docs/_includes/head.html
```diff
@@ -2,28 +2,32 @@
...
+    <title>{% t page.seo_title %}</title>
+  {% elsif page.title %}
+    <title>{% t site.name %}: {{ page.title }}</title>
+  {% else %}
+    <title>{% t site.title %}</title>
+  {% endif %}
+  
+  {% if page.seo_description %}
+    <meta name="description" content="{% t page.seo_description %}">
+  {% endif %}
```
nguyer, line 15: Previously there was a distinction between the default page layout, and the translated layout (`i18n_page`), where only the latter used the translation plugin to fill in strings. This meant that if you weren't working with page that was translated, you didn't put any strings in the translations file. This seems to be a change in that behavior in the SEO titles and descriptions always need to be in the translations file.

So one question is, should the SEO information change with the language? If so, that's fine. I'm just not sure.

It also seems a bit strange that I need both the `title` and the `seo_title` now (in different files for a `default` layout) and the `title` is only used in the nav menu. Can we combine these to be the same field?

docs/_i18n/en.yml
```diff
@@ -58,3 +64,230 @@ pages:
   xdc_testnet: XDC Testnet
   zksync_testnet: zkSync Testnet
   rotate_dx_certs: Rotate Data Exchange Certificates
+seo:
+  home_title: "Hyperledger FireFly: Decentralized Transaction Processing"
```
nguyer, line 68: Is this subtext used elsewhere to describe FireFly? I'm not sure what the "right" string is here yet, but I've never heard FireFly described as a decentralized transaction processor.

docs/_i18n/en.yml
```diff
@@ -58,3 +64,230 @@ pages:
...
+  home_description: "Explore Hyperledger Firefly documentation and features. Learn about distributed ledger technology, smart contracts, and secure data exchange."
+  architecture_title: "Hyperledger FireFly Architecture: Overview and Overview"
```
nguyer, line 70: Maybe just one "Overview"?

docs/_i18n/en.yml
```diff
@@ -58,3 +64,230 @@ pages:
...
+  architecture_description: "Hyperledger Firefly Architecture: Discover the decentralized architecture of Firefly for secure and transparent transaction processing."
```
nguyer, line 71:
```suggestion
  architecture_description: "Hyperledger FireFly Architecture: Discover the decentralized architecture of Firefly for secure and transparent transaction processing."
```

docs/_i18n/en.yml
```diff
@@ -58,3 +64,230 @@ pages:
...
+  architecture_blockchain_connector_framework_title: "Hyperledger FireFly Architecture: Blockchain Connector Framework"
+  architecture_blockchain_connector_framework_description: "Explore the Blockchain Connector Framework in Hyperledger's FireFly architecture. Learn about the design and functionalities of the framework, which enables seamless integration and interoperability with various blockchain networks and protocols within the FireFly blockchain platform."
```
nguyer, line 73:
```suggestion
  architecture_blockchain_connector_framework_description: "Explore the Blockchain Connector Framework in Hyperledger FireFly's architecture. Learn about the design and functionalities of the framework, which enables seamless integration and interoperability with various blockchain networks and protocols within the FireFly blockchain platform."
```

docs/_config.yml
```diff
@@ -13,3 +13,4 @@ color_scheme: firefly
 plugins:
   - jekyll-multiple-languages-plugin
 languages: ["en"]
+google_tag_manager: GTM-N58535C
```
nguyer, line 16: We might need to swap this out for a different key. I'm checking on this to see if Hyperledger has one we can use.

nguyer (changes requested):
I have a few questions about the mechanics of where stuff is laid out in the docs now. But my bigger concern is the accuracy of the generated descriptions. I started reviewing each description, and after reading several quickly realized they are auto generated. Many of them are "close" to being correct but use terminology that is potentially misleading about what FireFly is or does.

docs/_includes/head.html
keeganhrobinson: So I guess I'd say in a dream world you'd want all the content to be translatable. So it's good that the way it is here supports that.

I'm definitely down to continue having `layout: default` available for pages that don't have translations and making it so that it does not require making edits to the translation files. I just tested and that works now but it does break if you try something like `seo_title: "Sample Page Title"` or `seo_description: "Sample page description."`. Maybe I'm missing something obvious but I couldn't find a way to show the non-translated string if it's there in the template. Any chance you have any ideas? I'll keep digging but figured I'd ask.

On the `title` and `seo_title` pieces I do think they should be separate. For the most part, the SEO titles are longer or more descriptive than the titles shown in the navigation. And the way I set it up it will pull in the `title` if there isn't a specific `seo_title` created.

keeganhrobinson:
@nguyer I just chatted with Marc and he's going to comb through the titles and descriptions next week so I'll update this to include those when I get them.

docs/_config.yml
keeganhrobinson: Awesome, thanks!

docs/_i18n/en.yml
keeganhrobinson: Yeah, definitely. Marc's working on combing through these.

docs/_i18n/en.yml
keeganhrobinson: Naw, we can definitely make it better and more in line with what we've said in the past. I'll get this updated when I get the update content.

nguyer:
@keeganhrobinson Is this ready for re-review?

keeganhrobinson:
Hey @nguyer! If we can, it'd be nice to bundle in Google Tag Manager with this too. So then we can see what effects—if any—these updates make. If you think that's going to be a while longer though let me take one more pass at this today to make sure everything's good with the latest copy on the site.

nguyer:
@SeanBohan Do you know where we're up to with the GTM updates? I think we still need the tag ID to put on the docs site so it can be tracked.

keeganhrobinson:
@nguyer I just updated the code so that both the `default` and `i18n_page` layouts work with the SEO meta tags. I also updated the Google Tag Manager to the one you sent along. The last thing for this should just be making sure we get access to that tag manager account so we can use it.

Let me know what you think.

nguyer:
Closing this for now as it has been open for a long time with no activity. If we still need it, please feel free to re-open and update the PR. Thanks.

PR #1365, opened 2023-07-10
v1.2: fix: multiple named tuple result from contract

awrichar:
Backport of https://github.com/hyperledger/firefly/pull/1333


internal/apiserver/server.go
```diff
@@ -66,7 +66,7 @@ type apiServer struct {
 func InitConfig() {
 	httpserver.InitHTTPConfig(apiConfig, 5000)
 	httpserver.InitHTTPConfig(spiConfig, 5001)
-	httpserver.InitHTTPConfig(metricsConfig, 6000)
+	httpserver.InitHTTPConfig(metricsConfig, 7000)
```
nguyer, line 69: I'm not sure that we want to change the default metrics port in a patch release. Was there a reason for this change?

PR #1375, opened 2023-07-25
Correctly parse blockchain subscription name when delivering events

### LINKED ISSUE #1374: Blockchain events are delivered to every namespace by blockchain plugin
When using FireFly with many namespaces, I noticed that blockchain events are delivered by the plugin to _all_ namespaces. The event manager for each namespace performs filtering, but this is wasteful. The intention was always that the blockchain plugins should (as much as possible) direct blockchain events to only the namespace that needs them.

awrichar:
This ensures the namespace can be accurately extracted.

Fixes #1374

### DIFF
--- a/internal/blockchain/common/common.go
+++ b/internal/blockchain/common/common.go
@@ -334,7 +334,7 @@ func buildBatchPin(ctx context.Context, event *blockchain.Event, params *BatchPi
 func GetNamespaceFromSubName(subName string) string {
 	var parts = strings.Split(subName, "-")
 	// Subscription names post version 1.1 are in the format `ff-sub-<namespace>-<listener ID>`
-	if len(parts) != 4 {
+	if len(parts) < 4 {
 		// Assume older subscription and return empty string
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


internal/blockchain/common/common.go
```diff
@@ -334,7 +334,7 @@ func buildBatchPin(ctx context.Context, event *blockchain.Event, params *BatchPi
 func GetNamespaceFromSubName(subName string) string {
 	var parts = strings.Split(subName, "-")
 	// Subscription names post version 1.1 are in the format `ff-sub-<namespace>-<listener ID>`
-	if len(parts) != 4 {
+	if len(parts) < 4 {
```
nguyer, line 337: Is there a possibility that an old subscription could still hit this codepath if it contains a UUID?

awrichar: Yeaaa there is... 🫠

https://github.com/kaleido-io/firefly/commit/969d87e1cc68629a24a42acfa9392b1820461ef0#diff-b5bcf99952b7e15d54b1cb08438fb2337c46db9f1ba8d4bce463ef4a0dfb9168

awrichar:
Replaced by #1378

PR #1378, opened 2023-07-26
Correctly parse blockchain subscription name when delivering events

awrichar:
This ensures the namespace can be accurately extracted.

Fixes https://github.com/hyperledger/firefly/issues/1374

### DIFF
--- a/internal/blockchain/common/common.go
+++ b/internal/blockchain/common/common.go
@@ -332,13 +332,24 @@ func buildBatchPin(ctx context.Context, event *blockchain.Event, params *BatchPi
 }
 
 func GetNamespaceFromSubName(subName string) string {
-	var parts = strings.Split(subName, "-")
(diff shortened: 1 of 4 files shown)


internal/blockchain/common/common.go
```diff
@@ -332,9 +333,20 @@ func buildBatchPin(ctx context.Context, event *blockchain.Event, params *BatchPi
...
+	const UUIDLength = 36
+	var parts []string
+	if len(subName) > UUIDLength {
+		uuidStart := len(subName) - UUIDLength
+		prefix := subName[:uuidStart-1]
+		listenerID := subName[uuidStart:]
+		if _, err := uuid.Parse(listenerID); err == nil {
+			parts = strings.Split(prefix, "-")
+		}
+	}
```
nguyer, line 346: Namespace names are also allowed to have dashes which I think would cause problems here

nguyer (changes requested):

nguyer (approved):

PR #1380, opened 2023-07-27
log: add debug logs for webhook

EnriqueL8:
Increase log coverage for webhooks as it's quite helpful when debugging


internal/events/webhooks/webhooks.go
```diff
@@ -403,7 +402,7 @@ func (wh *WebHooks) attemptRequest(ctx context.Context, sub *core.Subscription,
 		Status:  resp.StatusCode(),
 		Headers: fftypes.JSONObject{},
 	}
-	// log.L(wh.ctx).Infof("Webhook<- %s %s event %s on subscription %s returned %d", req.method, req.url, event.ID, sub.ID, res.Status)
+	log.L(wh.ctx).Infof("Webhook<- %s %s on subscription %s returned %d", req.method, req.url, sub.ID, res.Status)
```
nguyer, line 405: Should this be `INFO` or `DEBUG` level? Looks like the corresponding log message above is `DEBUG`.

EnriqueL8: Sure happy to change it to DEBUG

go.work
```diff
@@ -1,7 +1,9 @@
-go 1.18
+go 1.19
```
EnriqueL8: Need to undo this sorry

nguyer (approved):

PR #1322, opened 2023-05-25
Separate "define" and "publish" for contract APIs, and allow deleting unpublished APIs

awrichar:
Part of the fix for https://github.com/hyperledger/firefly/issues/1220
In a chain with https://github.com/hyperledger/firefly/pull/1279 and #1321

PR #1395, opened 2023-09-05
handle update for unpublished API

Chengxuan:
Update to unpublished contract APIs are no longer possible due to https://github.com/hyperledger/firefly/pull/1322

This PR fixes the problem by adding the upsert logic back in for unpublished contract APIs.

### DIFF
--- a/internal/definitions/sender_contracts_test.go
+++ b/internal/definitions/sender_contracts_test.go
@@ -306,6 +306,27 @@ func TestDefineContractAPIPublishNonMultiparty(t *testing.T) {
 	assert.Regexp(t, "FF10414", err)
 }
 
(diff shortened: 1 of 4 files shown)


internal/definitions/sender_contracts_test.go
```diff
@@ -306,6 +306,27 @@ func TestDefineContractAPIPublishNonMultiparty(t *testing.T) {
 	assert.Regexp(t, "FF10414", err)
 }
 
+func TestDefineContractAPIPublishNonMultipartyUpdate(t *testing.T) {
```
awrichar, line 309: This isn't a "publish" test (so I would remove the word "publish" from the name).

awrichar (approved):

PR #1396, opened 2023-09-05
Proposal: Generic subscription options while adding contract listener

niklas-emmelius-mittelstand-ai:
## Generic subscription options while adding contract listener

Generic options for the automatically generated subscription can be passed in listener options

```json
{
  "name": "PayPerUseListener",
  "options": {
    "firstEvent": "newest",
    "subscriptionOptions": {
      "additionalOption1": "option1",
      "additionalOption2": "option 2"
    }
  },
  "topic": "default1"
}
```

## Intention

A privacyGroupId could be passed to Ethconnect and calls to priv_getLogs etc. would be possible.
Every plugin would be able to listen for the subscriptionOptions necessary for its subscription configuration.

The introduced changes do not break existing functionality.

### DIFF
--- a/pkg/core/contract_listener.go
+++ b/pkg/core/contract_listener.go
@@ -44,7 +44,8 @@ type ContractListenerWithStatus struct {
 	Status interface{} `ffstruct:"ContractListenerWithStatus" json:"status,omitempty" ffexcludeinput:"true"`
 }
 type ContractListenerOptions struct {
-	FirstEvent string `ffstruct:"ContractListenerOptions" json:"firstEvent,omitempty"`
+	FirstEvent          string           `ffstruct:"ContractListenerOptions" json:"firstEvent,omitempty"`
+	SubscriptionOptions *fftypes.JSONAny `ffstruct:"ContractListenerOptions" json:"subscriptionOptions,omitempty"`
 }
 
 type ListenerStatusError struct {
--- a/internal/blockchain/ethereum/ethereum.go
+++ b/internal/blockchain/ethereum/ethereum.go
@@ -888,10 +888,12 @@ func (e *Ethereum) AddContractListener(ctx context.Context, listener *core.Contr
 
 	subName := fmt.Sprintf("ff-sub-%s-%s", listener.Namespace, listener.ID)
 	firstEvent := string(core.SubOptsFirstEventNewest)
+	var subscriptionOptions *fftypes.JSONAny
 	if listener.Options != nil {
 		firstEvent = listener.Options.FirstEvent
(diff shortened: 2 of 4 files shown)


nguyer:
Ah, it looks like the linter is complaining because a file was updated by the copyright year in the header wasn't also updated. Sorry, it's a silly linter rule.

```
Error: pkg/core/contract_listener.go:1:17: Expected:2023, Actual: 2022 Kaleido, Inc. (goheader)
```

pkg/core/contract_listener.go
```diff
@@ -44,7 +44,8 @@ type ContractListenerWithStatus struct {
 	Status interface{} `ffstruct:"ContractListenerWithStatus" json:"status,omitempty" ffexcludeinput:"true"`
 }
 type ContractListenerOptions struct {
-	FirstEvent string `ffstruct:"ContractListenerOptions" json:"firstEvent,omitempty"`
+	FirstEvent          string           `ffstruct:"ContractListenerOptions" json:"firstEvent,omitempty"`
+	SubscriptionOptions *fftypes.JSONAny `ffstruct:"ContractListenerOptions" json:"subscriptionOptions,omitempty"`
```
nguyer, line 48: I'm struggling with what this field should be named. For context, in the older ETHConnect codebase we used to call these "subscriptions". In the newer FFTM API (which EVMConnect is based on), we now refer to these as "listeners" to help distinguish them from FireFly Subscriptions (which are a higher level object). So this field actually ends up maping to `Listener.Options`
https://github.com/hyperledger/firefly-transaction-manager/blob/fe5ace887c9f541c3053fd31d937d46ce6adc3c9/pkg/apitypes/api_types.go#L212

But calling this field `ContractListenerOptions.ListenerOptions` seems a bit redundant 🤔 

@awrichar or @peterbroadhurst Do you have any suggestions here?

nguyer:
Closing and re-opening to get the build to run again

pkg/core/contract_listener.go
peterbroadhurst: @nguyer - sorry I missed this before.

I think the consistent pattern here with other things, would be that the `options` are passed through to EVMConnect. If you look at `invoke` APIs, or the tokens - that's what we do I believe.

nguyer:
@niklas-emmelius-mittelstand-ai Sorry it took so long to get feedback on this one. If you want to rename `SubscriptionOptions` to `ListenerOptions` I'm happy to merge this PR. Thanks!

nguyer (changes requested):
You'll also want to run `make` locally to make sure the build is passing. Thanks!

PR #1398, opened 2023-09-05
Do not expect to retrieve local org during definition handling

### LINKED ISSUE #1397: Late joiners to multiparty networks may hang processing earlier definitions
When a node receives a definition broadcast (such as an FFI), it checks the local org definition to determine if the broadcast was sent by the local node or another:

https://github.com/hyperledger/firefly/blob/c454f8dca4e65b5891f6507da5dfa8aa3a3075d7/internal/definitions/handler_contracts.go#L164

However, if the local org does not exist in the database, this is interpreted as an error:

https://github.com/hyperledger/firefly/blob/c454f8dca4e65b5891f6507da5dfa8aa3a3075d7/internal/identity/identitymanager.go#L124

As a result, a late joiner to the network will fail to process definitions sent before it joined.

awrichar:
Only the org DID is needed for this check - no need to actually look up the org from the database. At this point the message author has already been verified, so a database lookup here is just extra unnecessary processing (and can lead to infinite retries if it fails).

Fixes #1397

### DIFF
--- a/internal/definitions/handler_contracts.go
+++ b/internal/definitions/handler_contracts.go
@@ -161,11 +161,11 @@ func (dh *definitionHandler) handleFFIBroadcast(ctx context.Context, state *core
 		return HandlerResult{Action: core.ActionReject}, i18n.NewError(ctx, coremsgs.MsgDefRejectedBadPayload, "contract interface", msg.Header.ID)
 	}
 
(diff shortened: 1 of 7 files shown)


peterbroadhurst (approved):
👍

PR #1406, opened 2023-09-24
Handle idempotent retry after error during initializing phase

peterbroadhurst:
Currently if an error occurs from a blockchain connector (EVMConnect for example) on submission of a transaction, then the operation goes immediately into `Failed` status - even if an `idempotencyKey` was supplied.

This means that a subsequent attempt to re-submit the transaction fails with:
```FF10431: Idempotency key '...' already used for transaction '...'```

This is not correct per the architecture described in https://hyperledger.github.io/firefly/head/reference/idempotency.html

Because the error happened during the submission of the idempotent transaction identifier into its FireFly Transaction Manager (FFTM) database, it is safe for a future retry to attempt to insert it again.

At that later point, the connector can return a `409` if it's already received that transaction, or it can successfully accept it into it's DB and process it.

This PR makes the following changes to move us towards the architecture defined behavior in this scenario:
- Where `RunOperation` used to return `complete` as a `bool`, a new `OpPhase` has been introduced:
    - `OpPhaseComplete` - good or bad, the operation is complete
    - `OpPhasePending` - the operation has been accepted, and it's completion will happen asynchronously
    - `OpPhaseInitializing` - the operation has not transitioned out of initializing yet
- Wherever an operation is executed, a new `idempotentSubmit` parameter is added
    - Set to `true` for all transactions where `idempotencyKey != ""`
    - When this is `true` a transaction will stay in `Initializing` (not move to `Failed`) if:
        - An error occurs
        - The `phase` returned from `RunOperation` is `OpPhaseInitializing`
    - Must be supplied externally to operation manager, as it does not have access to the TX object

These changes thus ensure that the `Initializing` status is held on to longer than is currently true.

Before: `Initializing` was only the status until the first attempt to submit the transaction was initiated
After: `Initializing` remains the status until the connector receives the idempotent submission of the TX

### CODE REVIEW

**Comment by peterbroadhurst** [comment]
Validated with raw blockchain transactions:
- Started core
- Stopped EVMConnect
- Submitted transaction with empty `idempotencyKey`
   - Resulting operation status was `Failed`
- Submitted transaction with `idempotencyKey` set to a value
   - Resulting operation status was `Initialized`
- Restarted EVMConnect and waited for FF to reconnect its WS
- Resubmitted transaction with `imdempotencyKey` set to failed failu
   - Resulting operation status was `Succeeded`
- Confirmed `FF10431` received if I tried to re-use idempotency key

File: internal/assets/token_pool.go [hunk 1]
```diff
@@ -154,7 +154,10 @@ func (am *assetManager) ActivateTokenPool(ctx context.Context, pool *core.TokenP
 		return err
 	}
 
-	_, err = am.operations.RunOperation(ctx, opActivatePool(op, pool))
+	_, err = am.operations.RunOperation(ctx, opActivatePool(op, pool),
+		false, // TODO: this operation should be made idempotent, but cannot inherit this from the TX per our normal semantics
+		//              as the transaction is only on the submitting side and this is triggered on all parties.
```
awrichar, line 159: Is this saying we insert the operation but not the transaction on the non-submitter members? That seems odd/incorrect.

FWIW activation is an idempotent operation at the connector level (so shouldn't be problematic if we activate multiple times) - not sure if that changes how we want to handle this path.

peterbroadhurst:
I extended the fix after testing, as I found in the tokens scenarios if I attempt to recreate the error by having the `resolve` endpoint down for the signing address, we wrote *zero operations* to the database.

Then when we came back round `ResubmitOperations` said it resubmitted zero - correctly - but that's not the information we needed and we returned an idempotency error.

 We actually needed to know _resubmitted zero, because no operations exist (vs. all operations reached `Pending` which was the previous assumption).

So I added a third `total` return parameter to `ResubmitOperations`, and in the case it is zero we now re-do the transactions as if we'd generated a new TX (but re-using the TX ID)

internal/assets/token_pool.go
peterbroadhurst: That was my reading of the code - would appreciate if you can confirm for me.

It was based on reading this:
https://github.com/hyperledger/firefly/blob/7067eb57b3b3c74a6109dd2e859e50c016f9b921/internal/definitions/handler_tokenpool.go#L99-L107

peterbroadhurst:
Completed E2E testing (noting it's important to also have https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/137 applied).

internal/assets/token_pool.go
awrichar: Looks like we do _eventually_ persist the token pool TX, after the activation completes:
https://github.com/hyperledger/firefly/blob/7067eb57b3b3c74a6109dd2e859e50c016f9b921/internal/events/token_pool_created.go#L149

So in the happy path, you do end up with a TX on all members, and the activation operations end up showing underneath it.

But it feels like we should probably persist the TX earlier, before we start creating operations.

internal/assets/token_pool.go
```diff
@@ -123,7 +132,7 @@ func (am *assetManager) createTokenPoolInternal(ctx context.Context, pool *core.
 		return nil, err
 	}
 
-	_, err = am.operations.RunOperation(ctx, opCreatePool(newOperation, &pool.TokenPool))
+	_, err = am.operations.RunOperation(ctx, opCreatePool(newOperation, &pool.TokenPool), pool.IdempotencyKey != "")
```
awrichar, line 135: I'm trying to page back in to why the IdempotencyKey here lives on the pool (and the transfer, and the approval) instead of on the operation or the transaction...

internal/assets/token_approval.go
```diff
@@ -107,20 +109,28 @@ func (s *approveSender) resolve(ctx context.Context) (opResubmitted bool, err er
...
+		resubmitWholeTX := false
 		if idemErr, ok := err.(*sqlcommon.IdempotencyError); ok {
-			resubmitted, resubmitErr := s.mgr.operations.ResubmitOperations(ctx, idemErr.ExistingTXID)
+			total, resubmitted, resubmitErr := s.mgr.operations.ResubmitOperations(ctx, idemErr.ExistingTXID)
 			if resubmitErr != nil {
 				// Error doing resubmit, return the new error
 				return false, resubmitErr
 			}
-			if len(resubmitted) > 0 {
+			if total == 0 {
```
awrichar, line 119: This logic feels slightly confusing to me. I guess what we're trying to differentiate is these 3 cases:
* no operations have ever been started for this transaction
* some operations were previously initialized for this transaction, but we think they weren't fully submitted, so we've now retried them
* some operations were previously initialized for this transaction, but we think they're all currently in flight, so we can't do anything but continue waiting

We're expressing these 3 states with a combination of two integer returns... wonder if there's a cleaner way to express it?

internal/assets/token_pool.go
awrichar: Never mind; this is just the input DTO. It is eventually stored on the transaction.

internal/assets/token_approval.go
awrichar: Right, point taken - it may not be worth continuing to iterate on this interface until we have the cycles to do a larger refactor. We should capture that next step as a task though.

peterbroadhurst: Raised https://github.com/hyperledger/firefly/issues/1410

internal/assets/operations.go
```diff
@@ -99,49 +99,49 @@ func (am *assetManager) PrepareOperation(ctx context.Context, op *core.Operation
...
 	case transferData:
 		plugin, err := am.selectTokenPlugin(ctx, data.Pool.Connector)
 		if err != nil {
-			return nil, false, err
+			return nil, core.OpPhaseInitializing, err
 		}
 		switch data.Transfer.Type {
 		case core.TokenTransferTypeMint:
-			return nil, false, plugin.MintTokens(ctx, op.NamespacedIDString(), data.Pool.Locator, data.Transfer, data.Pool.Methods)
+			return nil, core.OpPhaseInitializing, plugin.MintTokens(ctx, op.NamespacedIDString(), data.Pool.Locator, data.Transfer, data.Pool.Methods)
```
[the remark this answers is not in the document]
peterbroadhurst: It's the _phase_ we're in as a result of the error in this layer of the code, rather than the operation state.

e.g. this error occurred before we submitted the operation to the connector, so the calling code can decide if that means we don't store the operation at all, or we store it in state `Initializing`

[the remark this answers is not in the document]
peterbroadhurst: I have not analyzed very scenario to know if the error path is such, that it could *never succeed*.
If I had done that (quite large) piece of work everywhere, I could have introduced a phase of `OpPhaseUnprocessable` to mean - "this is never going to work, tell them it's broken"

internal/assets/token_approval.go
```diff
@@ -58,8 +59,9 @@ func (s *approveSender) setDefaults() {
 
 func (am *assetManager) NewApproval(approval *core.TokenApprovalInput) syncasync.Sender {
 	sender := &approveSender{
-		mgr:      am,
-		approval: approval,
+		mgr:              am,
+		approval:         approval,
+		idempotentSubmit: approval.IdempotencyKey != "",
```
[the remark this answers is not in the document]
peterbroadhurst: I would have to do more investigation to convince myself of that.. and if I did so, I'd probably want to rename `approval` to `approvalInput` (rather than the implying it's the actual approval object).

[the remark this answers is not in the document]
peterbroadhurst: So... if it's ok, think I'd like to leave it as is for this PR

internal/assets/token_pool.go
awrichar: https://github.com/hyperledger/firefly/issues/1411

internal/assets/operations.go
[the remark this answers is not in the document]
awrichar: I noticed in some managers you used `operations.ErrTernary`, but not here.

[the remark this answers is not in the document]
peterbroadhurst: Sorry - yes, this should be `operations.ErrTernary` almost certainly - I'd thought this was an error return.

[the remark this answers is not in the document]
peterbroadhurst: ok - added a commit for that

internal/assets/token_approval.go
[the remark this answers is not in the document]
awrichar: But yes, the thing we store is the input DTO from the user, so possibly it's named incorrectly.

awrichar (approved):
Looks good. I think all remaining feedback has been captured in new issues.

PR #1416, opened 2023-10-11
Add tezos signer service

denisandreenko:
The tezos signer service will be running as a part of the FF stack for tezos flow which is responsible for transactions signing process

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6,18 +6,18 @@
   },
   "evmconnect": {
     "image": "ghcr.io/hyperledger/firefly-evmconnect",
-    "tag": "v1.3.0",
-    "sha": "cddfbe3a67a660e72a35e79af931625fdc1a285c2ea97faef68ad0408aa0d8c3"
+    "tag": "v1.3.1",
+    "sha": "fdac5c853a73d97f7c9c7b56a4fa07a5bad4f5574568aa399e71427fba8d7094"
   },
   "fabconnect": {
     "image": "ghcr.io/hyperledger/firefly-fabconnect",
-    "tag": "v0.9.17",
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -36,8 +36,13 @@
...
     "image": "ghcr.io/hyperledger/firefly-signer",
-    "tag": "v1.1.9",
-    "sha": "515271ea722e6bf9601524880e7280a1fc052453733338d6ed474f85d6618aa1"
+    "tag": "v1.1.10",
+    "sha": "a06502872afda77d2a8c11a38a59c1c7b0eccc9ca3e7845711a2e2d6b98d31d2"
+  },
+  "tezossigner": {
+    "image": "ecadlabs/signatory",
+    "tag": "v1.1.10",
+    "sha": "2a48567c22af3ad9eb5860c1e29b1dea8aa73a7889e0e779fdfd7b882ae1458a"
```
nguyer, line 45: I don't think Signator needs to go in the `manifest.json`. Currently, the `manifest.json` is used for the following two reasons:

- Lists the images/tags of FireFly services (within the Hyplerledger project) that should be tested together
- Lists the images/tags of the containers used to compile FireFly core and its smart contracts

Currently, we do not list other third party services that may be required in some configurations. For example, we don't list specific versions of Besu, Geth, Fabric, etc. We also don't list a specific Postgres image, even though it is a requirement if you're not using SQLite. I tend to think of Signatory as falling in the same bucket. It's a runtime that is outside the project, but may be required as a runtime dependency in certain configurations.

In reality, FireFly is probably compatible with a wide range of versions of these services and we want to allow end users to upgrade FireFly independently of third-party services. The idea behind the `manifest.json` is if you are going to upgrade _FireFly_ here is the list of images that you should upgrade. How and when you upgrade third party services is up to you.

PR #1418, opened 2023-10-12
Enable contract listeners with multiple filters

nguyer:
This PR adds the ability to listen to multiple types of events on the same contract listener, by adding an array of listeners, rather than a single event signature/location per listener. The old way of creating a listener is still accepted by the API, but it will always be returned in the filters array now. **This is a migration concern that needs to be documented.**

## Open questions
One thing I'm not sure about here, is that this PR as-is removes the uniqueness constraint on listeners by topic/location/signature. It now allows multiples. I'm not sure if this is a problem or not. I can add that constraint back, but it would likely require some more sophisticated DB changes. Which brings me to the next point...

Right now all the filters for a contract listener get serialized to JSON and stored in a single column. I lated realized this means we lose the ability to query/filter (no pun intended) by signature, location, etc. which we used to do, in order to check for duplicates. I'm not sure if this is required or not, but wanted to call it out.

### Create contract listener request
```json
{
    "filters": [
        {
            "interface": {
                "id": "aaa0e410-2b5b-4815-a80a-a18f2ae59f7d"
            },
            "eventPath": "BatchPin",
            "location": {
                "address": "0xb0cd60ade460e797e0c9d206290ac4ed45672c60"
            }
        }
    ],
    "name": "CustomBatchPin",
    "options": {
        "firstEvent": "oldest"
    },
    "topic": "batch-pin"
}
```

### Create contract listener response
```json
{
    "id": "acc0d227-1da4-4d0d-bbe0-0c60f754158f",
    "namespace": "default",
    "name": "CustomBatchPin",
    "backendId": "018b258a-0c2c-07c0-5d59-50583ae91f1e",
    "created": "2023-10-12T20:18:06.012167Z",
    "filters": [
        {
            "event": {
                "name": "BatchPin",
                "description": "",
                "params": [
                    {
                        "name": "author",
                        "schema": {
                            "type": "string",
                            "details": {
                                "type": "address",
                                "internalType": "address"
                            },
                            "description": "A hex encoded set of bytes, with an optional '0x' prefix"
                        }
                    },
                    {
                        "name": "timestamp",
                        "schema": {
                            "oneOf": [
                                {
                                    "type": "string"
                                },
                                {
                                    "type": "integer"
                                }
                            ],
                            "details": {
                                "type": "uint256",
                                "internalType": "uint256"
                            },
                            "description": "An integer. You are recommended to use a JSON string. A JSON number can be used for values up to the safe maximum."
                        }
                    },
                    {
                        "name": "action",
                        "schema": {
                            "type": "string",
                            "details": {
                                "type": "string",
                                "internalType": "string"
                            }
                        }
                    },
                    {
                        "name": "uuids",
                        "schema": {
                            "type": "string",
                            "details": {
                                "type": "bytes32",
                                "internalType": "bytes32"
                            },
                            "description": "A hex encoded set of bytes, with an optional '0x' prefix"
                        }
                    },
                    {
                        "name": "batchHash",
                        "schema": {
                            "type": "string",
                            "details": {
                                "type": "bytes32",
                                "internalType": "bytes32"
                            },
                            "description": "A hex encoded set of bytes, with an optional '0x' prefix"
                        }
                    },
                    {
                        "name": "payloadRef",
                        "schema": {
                            "type": "string",
                            "details": {
                                "type": "string",
                                "internalType": "string"
                            }
                        }
                    },
                    {
                        "name": "contexts",
                        "schema": {
                            "type": "array",
                            "details": {
                                "type": "bytes32[]",
                                "internalType": "bytes32[]"
                            },
                            "items": {
                                "type": "string",
                                "description": "A hex encoded set of bytes, with an optional '0x' prefix"
                            }
                        }
                    }
                ]
            },
            "location": {
                "address": "0xb0cd60ade460e797e0c9d206290ac4ed45672c60"
            },
            "interface": {
                "id": "aaa0e410-2b5b-4815-a80a-a18f2ae59f7d"
            },
            "signature": "BatchPin(address,uint256,string,bytes32,bytes32,string,bytes32[])"
        }
    ],
    "topic": "batch-pin",
    "options": {
        "firstEvent": "oldest"
    }
}
```


peterbroadhurst:
> which we used to do, in order to check for duplicates. I'm not sure if this is required or not, but wanted to call it out.

@nguyer (fyi @EnriqueL8) - there is a really good reason why we need this (at least prior to this PR). That's because calculating the signature of a listener is not an easy thing to do - and FireFly doesn't expose any way for code outside of FireFly to do it.

So if you think about the common startup reconcile in most applications, where it does a query-then-create on the listeners it needs, you need to be able to match the things you get back, to the things you need.

That means you need to be able to create the signature for the _things you need_ (not just the things you create).

Currently, we avoid this by having it so that if you try and recreate the new thing with the same signature, it prevents you with a 409. In fact that's important enough that we found and fixed a bug in it recently in #1433.

Now I can imaging this PR might provide a proposal for a fundamentally new way to do this, by pushing the complex reconcile logic inside of listener. But we do need a clear migration path, and to know what the impact would be on apps relying on the old behavior - because without any thought they would *duplicate a new listener on every startup*.

nguyer:
@peterbroadhurst I'm trying to pick this back up but really not sure what direction to take it. I think part of my struggle is that I myself haven't run into the use case that is driving the need for this functionality, so I'm sort of designing a solution to an unknown problem.

I could add an updated version of the constraint we used to have (and maybe do some heavy DB schema changes to support that) but what would the constraint be? Topic/Location/[SOMETHING]. Is that last element of the constraint the combined signature of each filter in the listener? Do we need to prevent someone from creating the same exact Topic/Location/[Combination of Signatures] or do we want to prevent the same Topic/Location/[Single event signature] from being used in multiple listeners even if the set of filters in each of those listeners is different. Does that question make sense?

EnriqueL8:
New approach is that we have introduced a new API at `/contract/listeners/hash` that accepts the same payload as a POST. I did think we could have a subset of payload so just the filters and event part, could clean that up but for consistency kept the same.  This API will either return the FilterHash for the filters or the deprecated signature for one event and the filters hash as well.  Exposes as well the filterHash to be returned from the POST when creating a new listeners and queryable on the GET of listeners to figure out. 

PR needs some conflict cleaning and test coverage but test in a FF CLI setup and looking good!

pkg/core/contract_listener.go
```diff
@@ -26,17 +26,19 @@ import (
...
-	Options   *ContractListenerOptions `ffstruct:"ContractListener" json:"options,omitempty"`
+	ID         *fftypes.UUID            `ffstruct:"ContractListener" json:"id,omitempty" ffexcludeinput:"true"`
+	Interface  *fftypes.FFIReference    `ffstruct:"ContractListener" json:"interface,omitempty" ffexcludeinput:"postContractAPIListeners"`
+	Namespace  string                   `ffstruct:"ContractListener" json:"namespace,omitempty" ffexcludeinput:"true"`
+	Name       string                   `ffstruct:"ContractListener" json:"name,omitempty"`
+	BackendID  string                   `ffstruct:"ContractListener" json:"backendId,omitempty" ffexcludeinput:"true"`
+	Location   *fftypes.JSONAny         `ffstruct:"ContractListener" json:"location,omitempty" ffexcludeinput:"true"`
+	Created    *fftypes.FFTime          `ffstruct:"ContractListener" json:"created,omitempty" ffexcludeinput:"true"`
+	Event      *FFISerializedEvent      `ffstruct:"ContractListener" json:"event,omitempty" ffexcludeinput:"true"`
+	Signature  string                   `ffstruct:"ContractListener" json:"signature,omitempty" ffexcludeinput:"true"`
```
peterbroadhurst, line 37: Ok - maybe some certainty in my feedback here:
- If the `signature` lives here still, it has to be the hash in the case of `len(filters) > 1`
- I think in the `len(filters) > 1` case, we need a `signature` to be a part of each `filter`

peterbroadhurst (commented):
Hey @EnriqueL8 - lots of mini comments here, but it all rolls up to one big question about the relationship between:
- `len(filters) == 1` so `signature` is like `SomeEvent(1,2,3)`
- `len(filters) > 1` so:
    - Each `filter` has a `signature` like like `SomeEventA(1,2,3)` etc. is queryable in the `filter[0]` JSON
    - The `signature` of the listener is `hash(filters[*].signature)`

Then the implication of this is that `signature + topic` _is still a uniqueness criteria_ on the listener (assuming it is at 1.3.0)

EnriqueL8:
For the second case, we need to include the location if not it will not be unique and will cause problems

.vscode/launch.json
```diff
@@ -23,7 +23,7 @@
       "program": "${workspaceFolder}/main.go",
       "args": [
         "-f",
-        "${env:HOME}/.firefly/stacks/dev/runtime/config/firefly_core_0.yml"
+        "${env:HOME}/.firefly/stacks/listeners/runtime/config/firefly_core_0.yml"
```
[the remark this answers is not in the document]
EnriqueL8: nope

EnriqueL8:
@peterbroadhurst Thanks for the comments - will amend the PR with changed.

I think the challenge here is the `location:`
- The current uniqueness constraints is `location + topic + signature` which makes sense
- I think we need to keep the same thing at top level so what I propose is that for:
   - `len(filters) == 1` we stay with `location + topic + signature.
   - `len(filters)  > 1` we embed all the possible different locations into the hash that becomes the signature and still keep the signature for each event in `so queryable in the filter[0] JSON`. The signature then becomes `hash(filters[*].signature+filters[*].location)`

internal/blockchain/ethereum/ethereum.go
```diff
@@ -878,17 +878,47 @@ func (e *Ethereum) AddContractListener(ctx context.Context, listener *core.Contr
 			return err
 		}
 	}
-	abi, err := ffi2abi.ConvertFFIEventDefinitionToABI(ctx, &listener.Event.FFIEventDefinition)
-	if err != nil {
-		return i18n.WrapError(ctx, err, coremsgs.MsgContractParamInvalid)
+
+	filters := make([]*filter, 0)
+	// This keeps the existing behaviour where a contract listener could only listen to one event
+	if listener.Event != nil {
```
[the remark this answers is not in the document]
EnriqueL8: Yeah makes sense - I think this was done before for reload of contract listeners on restart, but how not standardise that on load of listeners from DB instead of here!

pkg/core/contract_listener.go
EnriqueL8: Agreed - have done that

db/migrations/postgres/000117_update_contractlisteners_add_filters_column.up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+ALTER TABLE contractlisteners ADD COLUMN filters TEXT;
+ALTER TABLE contractlisteners ALTER COLUMN event DROP NOT NULL;
+ALTER TABLE contractlisteners ALTER COLUMN location DROP NOT NULL;
+ALTER TABLE contractlisteners ALTER COLUMN signature DROP NOT NULL;
+ALTER TABLE contractlisteners ADD COLUMN filter_hash CHAR(64);
+CREATE INDEX contractlisteners_filter_hash ON contractlisteners(filter_hash);
```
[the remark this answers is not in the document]
EnriqueL8: I think it is safe to drop not null constraints for the location as the location will be embedded inside the filters so not needed at the top. The signature will remain the same.

This does raise the concern that we cannot from an API query parameter perspective filter over all contract listeners for a location... I probably need to address this and will make for a bigger set of changes

EnriqueL8:
@peterbroadhurst From the review comments and the chances I made, came up with two other worries:
- Filtering over certain fields at the API level will be worse as we are nesting a lot of the fields such as location, eventPath, etc.. under a JSONAny that is stored as TEXT in one column and doesn't have the native filtering. Wondering how bad that would be?
- Have made an opinionated assumption that both Tezos and Fabric only support a single filter for their event streams. I'm pretty certain that Fabric does restrict you to one, @denisandreenko I wonder if you might know if tezosconnect supports multiple filters for the subscriptions?

internal/contracts/manager.go
```diff
@@ -835,6 +901,50 @@ func (cm *contractManager) AddContractListener(ctx context.Context, listener *co
 		listener.Options.FirstEvent = cm.getDefaultContractListenerOptions().FirstEvent
 	}
 
+	// Normalize Event/EventPath + Location to list of Listeners
+	if len(listener.Filters) == 0 {
+		listener.Filters = append(listener.Filters, &core.ListenerFilterInput{
```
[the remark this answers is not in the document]
EnriqueL8: Have added this at the manager level and also still provided the deprecated fields in the API

EnriqueL8:
@denisandreenko  thanks for the input - so a follow up PR might be needed to provide the user with a better error on event streaming because it seems the code does go ahead and try and create a subscription in the Tezos connector

EnriqueL8:
Looks like Docker has been update in the Github Action and the packaged docker-compose version no longer support `ContainerConfig` https://github.com/docker/compose/issues/11742#issuecomment-2137583574

EnriqueL8:
This should do the trick https://github.com/hyperledger/firefly-cli/pull/307

peterbroadhurst:
> Multiple filters that clash with each other... so probably need to check that

This comment you made is highlighted, but there's no commentary on where you got to with it @EnriqueL8 .
What was your analysis/proposal?

pkg/core/contract_listener.go
```diff
@@ -31,12 +31,13 @@ type ContractListener struct {
 	Namespace string                   `ffstruct:"ContractListener" json:"namespace,omitempty" ffexcludeinput:"true"`
 	Name      string                   `ffstruct:"ContractListener" json:"name,omitempty"`
 	BackendID string                   `ffstruct:"ContractListener" json:"backendId,omitempty" ffexcludeinput:"true"`
-	Location  *fftypes.JSONAny         `ffstruct:"ContractListener" json:"location,omitempty"`
+	Location  *fftypes.JSONAny         `ffstruct:"ContractListener" json:"location,omitempty" ffexcludeinput:"true"`
 	Created   *fftypes.FFTime          `ffstruct:"ContractListener" json:"created,omitempty" ffexcludeinput:"true"`
-	Event     *FFISerializedEvent      `ffstruct:"ContractListener" json:"event,omitempty" ffexcludeinput:"postContractAPIListeners"`
-	Signature string                   `ffstruct:"ContractListener" json:"signature" ffexcludeinput:"true"`
+	Event     *FFISerializedEvent      `ffstruct:"ContractListener" json:"event,omitempty" ffexcludeinput:"true"`
```
peterbroadhurst, line 36: I don't understand how this is a safe change @EnriqueL8 
Surely in a point release we need to accept the old `POST` identically, and map it internally to the new way we're storing things?

peterbroadhurst:
Before I re-review here @EnriqueL8  - can you provide a new summary (either editing the header, or providing it below) of the externals changes.

I referenced one detail point here: https://github.com/hyperledger/firefly/pull/1418/files#r1643404103

... but then I stopped, as I wasn't sure what to base my code review against as a statement of intent.

The PR header doesn't answer all the questions, and maybe is out of date.

There seems to be at least two behaviors we could pick between, and I'm not sure which is proposed:

### 1) Migration impact, but efficient on storage

- Exactly the same `POST` inputs for listeners that were accepted previously (single `event`/`location`) are accepted after
- All DB storage is now done with a `filters` JSON entry containing an array
- DB migrations move all `event`/`location` entries over to `filters` (but didn't find this in the code, so got confused)
- **Migration impact** Query APIs that used to specify `?event=` / `?location=`  fields no longer work (both were JSON)
- Query APIs must now work against `?filters=` (still a JSON body, but now containing an array)

### 1) No migration impact, less inefficiency in storage

- Exactly the same `POST` inputs for listeners that were accepted previously (single `event`/`location`) are accepted after
- The `filters`, `event` and `location` fields are always propagated
- The `event` and `location` fields contain a copy of `filters[0]`
- Query APIs must now work against `?filters=`
- Query APIs still work against `?event=` or `?location=`, but with the caveat that only the first event filter is matched

pkg/core/contract_listener.go
EnriqueL8: We still accept it but this just means we are excluding it from the Swagger

EnriqueL8:
Thanks for the steer @peterbroadhurst!

I had in mind Option 2 as the migration impact of the filtering no longer working doesn't fit in by mind as a 1.3.1 but without the event and location being the `filter[0]`. Long term I prefer the Option 1 as it makes it clear that we can deprecate the `event` + `location` and just use filters.

I need to wrap my head around `Query APIs also work against ?filters=` and how that API would look like before proposing how that works. Is it you provide a JSON of what you want to see with some pattern matching?

```json
?filters=[["location":  {
         "address": "0xb0cd60ade460e797e0c9d206290ac4ed45672c60"
     }]
```

What does this give you? All the contract listeners that have one filter in that location

# Summary of external changes

- Exactly the same POST inputs for listeners that were accepted previously (single event/location) are accepted after
   - Optionally can specify filters as an array
   - Cannot specify both filters and event/location - error is thrown
   
The API would look like this:
```json
{
    // Old version
    "event": {}.
    "location":  {
         "address": "0xb0cd60ade460e797e0c9d206290ac4ed45672c60"
     }

    // or New version
    "filters": [
        {
            "interface": {
                "id": "aaa0e410-2b5b-4815-a80a-a18f2ae59f7d"
            },
            "eventPath": "BatchPin",
            "location": {
                "address": "0xb0cd60ade460e797e0c9d206290ac4ed45672c60"
            }
        }
    ],
    "name": "CustomBatchPin",
    "options": {
        "firstEvent": "oldest"
    },
    "topic": "batch-pin"
}
```

peterbroadhurst:
For `?filters=` - I think you're just missing an entry here: https://github.com/hyperledger/firefly/blob/c2c5a9551c2538cfde3d2ecc59b657cf09d3293f/pkg/database/plugin.go#L1051-L1063

EnriqueL8:
Review with @peterbroadhusrt

# This is the final proposal for the changes: 

Roughly Option 2 + other changes!

- Exactly the same POST inputs for listeners that were accepted previously (single `event/location`) are accepted after
    - Optionally can specify `filters` as an array
    - Cannot specify both `filters` and `event/location` - error is thrown
- All three of `filters`, `event` and `location` are always stored in the DB
- The `event` and `location` fields contain a copy of `filters[0]`
- We write a database migration in **golang code** at query time to copy over the event and location into the `filters[0]`
- We add `filters` to the API array that is just a whole JSON query
- New signature for each filter with `indexed` fields to have uniqueness when ABI for example has same event signature
- We construct a signature based on concatenated signatures of each filter separated by a semi colon
   -  This new signature allows to check for uniqueness of filters in the array
- Query APIs still work against `?event=` or `?location=`, but with the caveat that only the first event filter is matched
- Query APIs also work against `?filters=`
- Query APIs are enhanced for the signature as it will contain the location and can be regex pattern

EnriqueL8:
Tested this PR E2E with an EVM chain locally, feeling quite good but I'm seeing problems in the E2E tests when creating a listener. Digging into that...

EnriqueL8:
Found a problem with the way I wrote the just in time migration... fixing that

internal/blockchain/ethereum/ethereum.go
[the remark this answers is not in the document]
EnriqueL8: Need to address this now

internal/contracts/manager.go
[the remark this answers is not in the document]
EnriqueL8: agreed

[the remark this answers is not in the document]
EnriqueL8: Address this in the blockchain plugin

internal/contracts/manager.go
```diff
@@ -811,7 +840,118 @@ func (cm *contractManager) checkContractListenerExists(ctx context.Context, list
...
+			if strings.Compare(duplicateMapChecker[filter.Signature], location) == 0 {
+				return i18n.NewError(ctx, coremsgs.MsgDuplicateContractListenerFilterLocation, location)
+			}
+
+			// In the case of Fabric for example comparing listening to channel
+			// comparing to channel + chaincode
+			if strings.Contains(duplicateMapChecker[filter.Signature], location) || strings.Contains(location, duplicateMapChecker[filter.Signature]) {
+				return i18n.NewError(ctx, coremsgs.MsgDuplicateContractListenerFilterLocation)
+			}
+
```
EnriqueL8, line 907: I do wonder if this should be moved to the plugin specific code... or the plugin should generate a signature including the location

internal/contracts/manager.go
```diff
@@ -877,26 +1022,92 @@ func (cm *contractManager) AddContractListener(ctx context.Context, listener *co
...
+				filter.Eq("location", locationLookup),
+			)); err != nil {
+				return err
+			} else if len(existing) > 0 {
+				for _, listener := range existing {
+					// We have extended the event signature to add more information
+					// So we compare the start with is not guaranteed to be the same
+					// but it's the best comparison
+					if strings.HasPrefix(signature, listener.Signature) {
+						return i18n.NewError(ctx, coremsgs.MsgContractListenerExists)
```
EnriqueL8, line 1044: Again maybe this comparing of signatures should be plugin specific

internal/contracts/manager.go
peterbroadhurst: What's your proposal here @EnriqueL8? Are you suggesting you make a change before this goes in?

internal/contracts/manager.go
peterbroadhurst: I don't think comparing signatures should be specific, but I don't understand the `HasPrefix` here.
Surely the signatures are the same, or they are not.

doc-site/docs/reference/types/contractlistener.md
```diff
@@ -47,16 +47,17 @@ title: ContractListener
...
 | `namespace` | The namespace of the listener, which defines the namespace of all blockchain events detected by this listener | `string` |
 | `name` | A descriptive name for the listener | `string` |
 | `backendId` | An ID assigned by the blockchain connector to this listener | `string` |
-| `location` | A blockchain specific contract identifier. For example an Ethereum contract address, or a Fabric chaincode name and channel | [`JSONAny`](simpletypes.md#jsonany) |
+| `location` | Deprecated: Please use 'location' in the array of 'filters' instead | [`JSONAny`](simpletypes.md#jsonany) |
 | `created` | The creation time of the listener | [`FFTime`](simpletypes.md#fftime) |
-| `event` | The definition of the event, either provided in-line when creating the listener, or extracted from the referenced FFI | [`FFISerializedEvent`](#ffiserializedevent) |
+| `event` | Deprecated: Please use 'event' in the array of 'filters' instead | [`FFISerializedEvent`](#ffiserializedevent) |
 | `signature` | The stringified signature of the event, as computed by the blockchain plugin | `string` |
```
peterbroadhurst, line 57: Is this description still true?

internal/blockchain/fabric/fabric.go
```diff
@@ -897,6 +897,21 @@ func (f *Fabric) NormalizeContractLocation(ctx context.Context, ntype blockchain
 	return encodeContractLocation(ctx, ntype, parsed)
 }
 
+func (f *Fabric) StringifyContractLocation(ctx context.Context, location *fftypes.JSONAny) (string, error) {
```
peterbroadhurst, line 900: Can you help me understand the reason for this function?

Are we changing how we store the location in some way that means it's not a JSON object?

internal/contracts/manager.go
```diff
@@ -811,7 +840,118 @@ func (cm *contractManager) checkContractListenerExists(ctx context.Context, list
...
+		// Have parsed a filter with the same signature?
+		// In the case of location being different
+		if duplicateMapChecker[filter.Signature] != "" {
+			// If this duplicate filter is looking at all locations it's a superset of the previous one
+			// or the previous filter was also looking at all locations then we are trying to add a subset
+			if filter.Location == nil || duplicateMapChecker[filter.Signature] == "*" {
+				return i18n.NewError(ctx, coremsgs.MsgDuplicateContractListenerFilterLocation)
+			}
+
+			location, err := cm.blockchain.StringifyContractLocation(ctx, filter.Location)
```
peterbroadhurst, line 892: I see you've distinguished between `NormalizeContractLocation` returning JSON in a deterministic way and `StringifyContractLocation`. I'm not sure that's necessary.

It does seem like your proposal below is correct that `GenerateEventSignature` should take the location (or have a separate `GenerateEventSignatureWithLocation` function if there's some other use of `GenerateEventSignature` that needs to preserve the existing behavior).

internal/contracts/manager.go
```diff
@@ -811,7 +840,118 @@ func (cm *contractManager) checkContractListenerExists(ctx context.Context, list
...
+
+func (cm *contractManager) ConstructContractListenerSignature(ctx context.Context, listener *core.ContractListenerInput) (output *core.ContractListenerSignatureOutput, err error) {
+	output = &core.ContractListenerSignatureOutput{}
+
+	err = cm.parseContractListenerFilters(ctx, listener)
+	if err != nil {
+		return nil, err
+	}
+
+	output.Signature, err = cm.generateFilterSignature(ctx, listener.ContractListener.Filters)
```
peterbroadhurst, line 946: Why is this a separate function to `parseContractListenerFilters`?
It seems like `parseContractListenerFilters` already updates the output to set `Event` and `Location` - so why wouldn't it set `Signature` too?

internal/contracts/manager.go
```diff
@@ -877,26 +1022,93 @@ func (cm *contractManager) AddContractListener(ctx context.Context, listener *co
...
+	if err = cm.database.InsertContractListener(ctx, verifiedContractListener); err != nil {
 		return nil, err
 	}
 
-	return &listener.ContractListener, err
+	return verifiedContractListener, err
+}
+
+// Sort the filters by signature then location and generate a hash
+func (cm *contractManager) generateFilterSignature(ctx context.Context, filters core.ListenerFilters) (string, error) {
```
peterbroadhurst, line 1081: Per above this function seems mainly duplicative with the `parseContractListenerFilters` function. Looping through the same structures, and working out the same things.

I do see a `sort` here - which is interesting, whereas in the `parseContractListenerFilters` function we have a map.

peterbroadhurst (commented):
Next round of detail comments/questions up for your @EnriqueL8 - these are all around the "how" I think at this point, so that's great.

Pausing here as I do think there's some significant questions here for you to think about before I go further.

internal/blockchain/fabric/fabric.go
peterbroadhurst: (... came back to this in next comment)

doc-site/docs/reference/types/contractlistener.md
EnriqueL8: a good catch - fixing

db/migrations/postgres/000117_update_contractlisteners_add_filters_column.up.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+ALTER TABLE contractlisteners ADD COLUMN filters TEXT;
```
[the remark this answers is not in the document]
EnriqueL8: great stuff - yes the me update that!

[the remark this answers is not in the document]
EnriqueL8: we can either do `VARCHAR(MAX)` or `TEXT`

EnriqueL8:
@peter.broadhurst Just working through the changes came across `apis/{apiName}/listeners/{eventPath}` which uses the signature to query the DB for the listener... We need to make a decision if this API now makes sense with multiple filters in one listeners and this hierarchy only makes sense for one event or we implement a logic to look in the filters and find out if the event exists

internal/contracts/manager.go
EnriqueL8: So we need a sort to make sure that signatures are always the same given any order, it's only duplicative because it generates a signature which I can remove and use the one in the filter instead

internal/contracts/manager.go
EnriqueL8: Done

db/migrations/postgres/000117_update_contractlisteners_add_filters_column.up.sql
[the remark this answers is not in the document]
EnriqueL8: Actually the problem here is the index.... I would need to recreate the index and in the case of TEXT you need to specify how long the key should be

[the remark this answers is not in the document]
EnriqueL8: Moving to VARCHAR(MAX) should affect the index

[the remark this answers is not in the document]
EnriqueL8: you can't modify a type in SQLITe so need to work on creating a new field and moving things across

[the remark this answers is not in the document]
EnriqueL8: aah according to this https://stackoverflow.com/questions/4788370/sqlite-varchar-default-size in SQLITE VARCHAR is equivalent to TEXT

EnriqueL8:
Great code coverage is working, fixing

internal/blockchain/ethereum/ethereum.go
```diff
@@ -843,6 +843,26 @@ func (e *Ethereum) QueryContract(ctx context.Context, signingKey string, locatio
...
+		return false, err
+	}
+
+	parsedRight, err := e.parseContractLocation(ctx, right)
+	if err != nil {
+		return false, err
+	}
+
+	// For Ethereum just compared addresses
+	return parsedLeft.Address == parsedRight.Address, nil
```
peterbroadhurst, line 863:
```suggestion
return strings.EqualFold(parsedLeft.Address, parsedRight.Address), nil
````

EnriqueL8:
The Postgres migration seem to be failing, looking into it.. It's a shame the GitHub actions do no give you logs of the docker containers, something to add to my list

EnriqueL8:
Replicate locally ```FF00184: Database migration failed: FF00184: Database migration failed: migration failed: syntax error at or near "MAX" (column 67) in line 4: BEGIN;```

Changing it to just VARCHAR which according to the Postgres documentation is unbounded

peterbroadhurst (approved):
✨ 🚀  🌑

PR #1419, opened 2023-10-20
namespace scoped web sockets

hosie:
If I am running firefly behind an HTTP aware gateway, I can govern access to REST api for namespaces based on the HTTP path.  However, for websockets, this is not possible.  namespace isolation is much looser and requires analysis of query parameters and/or websocket message payloads.

This PR proposes a solution where a WebSocket connection can be opened on a namespace specific path (`/api/v1/namespaces/{namespace}/ws`) and the web socket plugin will only allow streams to be started for the given namespace.  

This also means that `namespace` query parameter and `namespace` field of the `start` message become redundant.  For compatibility reasons, the PR tolerates them but asserts that they must be equal to the namespace for which the connection was opened.


nguyer (approved):
Looks good to me. Thanks!

PR #1424, opened 2023-11-09
Update log level in config auto reload

### LINKED ISSUE #1421: Dynamic Config Reload for Non-namespace Settings
When using the dynamic config reload feature of FireFly, changing some non-namespace settings, for example the log level, will result in the config getting reloaded, but the log level isn't actually changed until FireFly is restarted. I'm opening this issue to discuss whether this is "working as designed" or if we want to re-configure more things within the runtime during a dynamic config reload.

nguyer:
Partially resolves https://github.com/hyperledger/firefly/issues/1421

This solves the log level getting set, which is arguably the most important for operating a production system where you may need to temporarily change the log level while the system is running to test something.

### DIFF
--- a/internal/namespace/configreload.go
+++ b/internal/namespace/configreload.go
@@ -61,6 +61,8 @@ func (nm *namespaceManager) configFileChanged() {
 }
 
 func (nm *namespaceManager) configReloaded(ctx context.Context) {
+	// Always make sure log level is up to date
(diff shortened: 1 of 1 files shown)


Chengxuan (approved):
Looks great!

denisandreenko (approved):

peterbroadhurst (approved):

PR #1427, opened 2023-11-20
Fix JSON schema output for custom contracts

nguyer:
Previously we were generating an incorrect JSON Schema for the response body on custom contract endpoints. This PR fixes this so that:

- `/inovke` endpoints always return the correct schema for a FireFly Operation (which is the actual payload they return)
- `/query` endpoints now return the correct schema based on the return value of the smart contract itself
  - If outputs are unnamed, the swagger generator now names them `output`, `output1`, `output2`, etc. like ETH/EVMConnect name them

### DIFF
--- a/internal/apiserver/ffi2swagger.go
+++ b/internal/apiserver/ffi2swagger.go
@@ -92,7 +92,6 @@ func addFFIMethod(ctx context.Context, routes []*ffapi.Route, method *fftypes.FF
 		JSONInputSchema: func(ctx context.Context, schemaGen ffapi.SchemaGenerator) (*openapi3.SchemaRef, error) {
 			return contractRequestJSONSchema(ctx, &method.Params, hasLocation)
 		},
-		// // TODO change to operation schema
 		JSONOutputValue:          func() interface{} { return &core.OperationWithDetail{} },
 		JSONOutputCodes:          []int{http.StatusOK},
 		PreTranslatedDescription: description,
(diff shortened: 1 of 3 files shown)


internal/apiserver/ffi2swagger.go
```diff
@@ -90,11 +90,10 @@ func addFFIMethod(ctx context.Context, routes []*ffapi.Route, method *fftypes.FF
 		Path:   fmt.Sprintf("invoke/%s", method.Pathname), // must match a route defined in apiserver routes!
 		Method: http.MethodPost,
 		JSONInputSchema: func(ctx context.Context, schemaGen ffapi.SchemaGenerator) (*openapi3.SchemaRef, error) {
-			return contractJSONSchema(ctx, &method.Params, hasLocation)
-		},
-		JSONOutputSchema: func(ctx context.Context, schemaGen ffapi.SchemaGenerator) (*openapi3.SchemaRef, error) {
-			return contractJSONSchema(ctx, &method.Returns, true)
+			return contractRequestJSONSchema(ctx, &method.Params, hasLocation)
 		},
+		// // TODO change to operation schema
```
awrichar, line 95: Not sure I understand what's left for this TODO?

internal/apiserver/ffi2swagger.go
```diff
@@ -193,6 +192,40 @@ func contractJSONSchema(ctx context.Context, params *fftypes.FFIParams, hasLocat
 	return openapi3.NewSchemaRef("", s), nil
 }
 
+/**
+ * Parse the FFI and build a corresponding JSON Schema to describe the response body for "invoke" or "query".
```
awrichar, line 196: I think this response body is only for query outputs

awrichar (approved):

internal/apiserver/ffi2swagger.go
nguyer: Yep. Good catch. Misleading comments are the worst kind of comments.

PR #1432, opened 2023-12-13
Add name/version if not specified

peterbroadhurst:
In recent changes, the `name`/`version` of the FFI have become required, and this means that the API for `/generate` has changed behavior so it fails if these are not supplied


awrichar (approved):

awrichar:
Follow-on to #1432

nguyer:
@awrichar I think you accidentally linked to this PR where you said this was a follow-on to something else.

Haven't name and version always been required? For as long as I can remember I've always needed to take the output from `/contracts/interfaces/generate` and manually add a `name` and `version` if I didn't specify one in the original request before doing the `POST` to `contracts/interfaces` to create the FFI.

Do we have a link to what changed previously that requires this change now?

nguyer:
Okay after reading a bit more I realize that we now require a `name` and a `version` on `/contracts/interfaces/generate` which is somewhat new. This change makes those optional again (restoring existing behavior) by automatically filling them in with some default values. Sounds good to me.

PR #1439, opened 2024-01-09
Use non-root user in Docker image

nguyer:
This PR updates the Dockerfile to build and run with a non-root user.

> **NOTE**: This requires an updated firelfy-cli to pass E2E. Corresponding CLI changes can be found in the PR here: https://github.com/hyperledger/firefly-cli/pull/281 which is why this PR still references that specific commit.

TODO:
- [x] Update CLI version to new release after https://github.com/hyperledger/firefly-cli/pull/281 is merged


manifest.json
```diff
@@ -59,6 +59,6 @@
     "release": "v1.2.0"
   },
   "cli": {
-    "tag": "v1.2.1"
+    "tag": "14387a519ef794e41b85c039e40ddc419fd5967a"
```
peterbroadhurst, line 62: Just a quick sanity check on this against the TODO in the comment, before merg

peterbroadhurst (approved):
Approved - but not merging due to comment above about reconciling the TODO

PR #1440, opened 2024-01-10
chore: start upgrade to firefly-common 1.4.1

SamMayWork:
Begin upgrade of FireFly to the latest version of FireFly common


go.work
```diff
@@ -1,7 +1,7 @@
-go 1.18
+go 1.21
```
nguyer, line 1: Noting that this is a big jump. I think it's the right time to do it, but it will also be worth calling out to the rest of the community that we require Go v1.21 or newer now.

nguyer (changes requested):
This is great, thank you! I have one request for a related change. Could you please also update the GitHub workflows to use go "1.21" as well? It's probably also worth checking that we're using the latest versions of actions (like checkout v4) in our workflows too. Thanks!

go.work
SamMayWork: Should be addressed by https://github.com/hyperledger/firefly/pull/1440/commits/ac44d47be7000b1f2ae277040775db8998858696. The changes to make FireFly use 1.21 consistently are more far reaching than just the workflows so have updated those places.

The new images in the commit come from both the golang and alpine image listings:
 - https://hub.docker.com/_/golang 
 - https://hub.docker.com/_/alpine

nguyer (approved):
Good spot on the docker images too. Thank you! This looks good.

SamMayWork:
Looking at the build failures :eyes:

PR #1447, opened 2024-01-15
Enable batch delivery over WebSockets

peterbroadhurst:
### New support of `batch` for WebSockets

In https://github.com/hyperledger/firefly/pull/1359 @EnriqueL8 added support for Webhooks (outbound HTTPS requests) to configure `batch: true` on the subscription, such that events are delivered multiple at a time.

This was deferred at the time for WebSockets (inbound HTTPS long-lived connections) because they already had read-ahead performance optimization, so it was lower priority.

However, the use of `readahead` in its current form in WebSockets does not allow the _consumer of events to optimize_. For example performing a single DB transaction in the application after processing a batch of events. You could implement a complex batch aggregation routing on the app-side, but that seems much more effort than FireFly just doing the batching for you.

So this PR closes the gap and supports `batch: true` for WebSockets as well.

You can test it simply with a dev env:
ws://127.0.0.1:5000/ws?ephemeral=true&namespace=default&autoack&batch=true&batchtimeout=1s&readahead=50

### Fix to batch based delivery for WebHooks and WebSockets

In adding the WebSocket specific code, I found the original code in #1359 was not actually working as designed. It added a layer of extra batch collation logic on top of the event poller in `eventDelivery.deliverBatchedEvents`, but actually the way the dispatching code in `bufferedDelivery` worked the behavior was broken.

If you had 1 event arrive, then a very short delay and 2 more events arrive (simple to recreate with a `/broadcast` in a dev env), you ended up getting a batch of one, then _the whole batch delay waiting_, then a batch of 2 after a _second wait_. Instead of getting a single batch of 3 after the batch delay.

... so the batch delay was doubled to deliver two incomplete batches 😬 

This was quite a big problem, and after some wrestling I ended up:
1. Addressing some debt in the batch logic in `eventPoller`, which had ended up disabled a very long time ago
2. Setting the default batching in `eventPoller` to _off_ (`batchTimeout=0` by default in config) as this is actually the behavior that you get today
3. Removing the `eventDelivery.deliverBatchedEvents` function added in #1359 
4. Changing `eventDispatcher` to configure `eventPoller` batching settings
5. Update `eventDelivery.bufferedDelivery` and `eventDelivery.deliverEvents` to handle batching


peterbroadhurst:
Popped this back into draft, as I've found a pretty significant issue in how the batching assumptions work between `eventDispatcher.bufferedDelivery()` and `eventDispatcher.deliverBatchedEvents()`.

`eventDispatcher.bufferedDelivery()` waits for ack's for the _whole array_ of `dispatchable` after sending them down to the `ed.eventDelivery` channel.

But the `eventDispatcher.deliverBatchedEvents()` processor enabled when `ed.batch` is true, has a timer to wait for the batch to fill before it dispatches.

So in reality you get really bad delays.
Simple to recreate with `ws://127.0.0.1:5000/ws?ephemeral=true&batch=true&namespace=default&autoack&batchtimeout=15s` in a dev env

If you send a `broadcast` there are three events - on generated quickly (transaction submit), then two more after 1s (tx confirmed and then message confirmed), but what you get is:

- 15 second delay
- Batch of 1 event
- 15 second delay
- Batch of 2 events

What you should get is:

- 15 second delay
- Batch of 3 events

docs/reference/config.md
```diff
@@ -224,7 +224,7 @@ nav_order: 2
 |Key|Description|Type|Default Value|
 |---|-----------|----|-------------|
 |batchSize|The maximum number of records to read from the DB before performing an aggregation run|[`BytesSize`](https://pkg.go.dev/github.com/docker/go-units#BytesSize)|`200`
-|batchTimeout|How long to wait for new events to arrive before performing aggregation on a page of events|[`time.Duration`](https://pkg.go.dev/time#Duration)|`250ms`
+|batchTimeout|How long to wait for new events to arrive before performing aggregation on a page of events|[`time.Duration`](https://pkg.go.dev/time#Duration)|`0ms`
```
peterbroadhurst, line 227: This is actually the behavior you got before, because the code that had (a very, very long time ago) been started to implement batch-timeout had been disabled. Making a separate comment to highlight where.

internal/events/event_poller.go
```diff
@@ -289,20 +313,6 @@ func (ep *eventPoller) waitForShoulderTapOrPollTimeout(lastEventCount int) bool
 		return true
 	}
 
-	// For throughput optimized environments, we can set an eventBatchingTimeout to allow messages to arrive
-	// between polling cycles (at the cost of some dispatch latency)
-	if ep.conf.eventBatchTimeout > 0 && lastEventCount > 0 {
-		shortTimeout := time.NewTimer(ep.conf.eventBatchTimeout)
-		select {
-		case <-shortTimeout.C:
```
peterbroadhurst, line 297: See in this branch we just fall through to the `longTimeoutDuration` wait, it basically means there was no use of the `eventBatchTimeout` for its stated purpose.

An incomplete implementation had existed previously, and was partially removed leaving this code-cruft.

So in this PR I've come back round and implemented it with a short re-poll cycle (just once) after the batch timeout if the first time we trigger we get an incomplete batch.

It is disabled by default... but you could turn it on with readahead on subscriptions (and for optimization of the event loop), without needing to use batch-based delivery at all.

internal/events/event_dispatcher.go
```diff
@@ -384,88 +398,48 @@ func (ed *eventDispatcher) handleAckOffsetUpdate(ack ackNack) {
...
-
-		if len(events) == ed.readAhead || (timedOut && len(events) > 0) {
-			_ = ed.transport.BatchDeliveryRequest(ed.ctx, ed.connID, ed.subscription.definition, events)
-			// If err handle all the delivery responses for all the events??
-			events = nil
-		}
-	}
-}
-
 // TODO issue here, we can't just call DeliveryRequest with one thing.
```
awrichar, line 401: Is this still a TODO?

internal/events/event_dispatcher.go
```diff
@@ -384,88 +398,48 @@ func (ed *eventDispatcher) handleAckOffsetUpdate(ack ackNack) {
...
+				}
 			}
 
-			if err == nil {
-				err = ed.transport.DeliveryRequest(ed.ctx, ed.connID, ed.subscription.definition, event, data)
-			}
-			if err != nil {
-				ed.deliveryResponse(&core.EventDeliveryResponse{ID: event.ID, Rejected: true})
+			// In batch mode we do one dispatch of the whole set as one
+			if err == nil && ed.batch {
```
awrichar, line 433: What if we get to this point and `err != nil`? Do we just silently go round the loop again?

awrichar: Or maybe we already dealt with the error on line 428? Might not be a bug, just slightly difficult to follow the flow.

internal/events/event_dispatcher.go
```diff
@@ -384,88 +398,48 @@ func (ed *eventDispatcher) handleAckOffsetUpdate(ack ackNack) {
...
+				if withData && e.Event.Message != nil {
+					e.Data, _, err = ed.data.GetMessageDataCached(ed.ctx, e.Event.Message)
+				}
+				// Individual events (in reality there is only ever i==0 for this case)
+				if err == nil && !ed.batch {
+					err = ed.transport.DeliveryRequest(ed.ctx, ed.connID, ed.subscription.definition, e.Event, e.Data)
+				}
+				if err != nil {
+					ed.deliveryResponse(&core.EventDeliveryResponse{ID: e.Event.ID, Rejected: true})
+				}
```
awrichar, line 429: Maybe a nit, but feels like this could be moved outside of the loop (ie down to line 431). Took me a few reads to convince myself that all errors were handled across all branches.

awrichar: But will leave it with you on what you think is ultimately clearer.

awrichar (approved):

internal/events/event_dispatcher.go
peterbroadhurst: nope - deleted TODO

internal/events/event_dispatcher.go
peterbroadhurst: Ah, you actually point out a problem here, the reason it's in the loop is we still need to nack the other events. So I need to fix the logic.

peterbroadhurst: Done - would appreciate you taking a 2nd look if you don't mind

awrichar (approved):
New change looks good. Thanks for the comments on the flow.

PR #1448, opened 2024-01-17
Update the FF manifest

denisandreenko:


nguyer:
@denisandreenko It looks like there might be a compatibility issue between this build of Fabconnect that was pulled in and this build of FireFly core. I will look into that. Feel free to revert the Fabconnect portion of this change if you want the other one to go in. I'll sort through that issue in a separate PR.

nguyer:
Okay actually this issue is related to a recent change in how we build docker images (the user with which we run the app) but is already fixed in this PR for the FireFly CLI: https://github.com/hyperledger/firefly-cli/pull/281

I just need a code review and merge on that one.

nguyer (approved):

Issue #1443, opened 2024-01-10
Query events by subscription filter


nguyer:
Subscriptions have a filter on them, which is applied to events as they appear in the event stream to determine if they match the subscription.

This filter is spelt with a JSON semantic that is completely unique to subscriptions:
https://hyperledger.github.io/firefly/head/reference/types/subscription.html#subscriptionfilter

So if you wanted to browse the history of messages that would be delivered to your subscription, and look at where the `offset` of your subscription is up to in that stream, it's a really hard thing to do!

You have to construct an API query from scratch, that you _believe is accurate_ compared to what FireFly is constructing internally in `Go` code.

So, this enhancement request is to instead provide a simple API `GET` endpoint underneath a `subscription` to ask FireFly to do that query-building for you.

e.g. you get a collection API that's pre-filtered by all the filtering that's part of your `SubscriptionFilter`, and when you do queries on top of it (like `id=xyz` or `created=>...`) you're adding an _additional set_ of filters on top of the filters already in place.

peterbroadhurst:
The wrinkle here, is that SubscriptionFilter applies filters as a Regular Expression, and currently we don't have support for regular expression filters in our DB semantics.

So I think the most straight forward and determinsitic way to build this feature is:
- Run the DB query without `skip` and `limit` from the API call (use our own internal `skip` + `limit`)
- Use the existing `go` code from the filtering that is done in the subscription to post-filter these
- Loop on the `skip` + `limit` in our internal loop, until we've fulfilled the `skip` + `limit` on the external

SamMayWork:
PR for this now open at https://github.com/hyperledger/firefly/pull/1452

EnriqueL8:
This has been merged and completed

PR #1452, opened 2024-01-23
feat: Add an API to allow for querying events under a subscription with additional filtering

SamMayWork:
ref: https://github.com/hyperledger/firefly/issues/1443

Adds a new fancy API: `subscriptions/{subid}/events`
 - Internally iterates through all of the events, finding those that match a subscription
 - Allows for normal query filtering against the filtered list of subscriptions

My current concern is that this query is an expensive operation because by necessity, it needs to be O(n) as we have to go through all of the events to find those that match the subscription type. Additionally, skip provided by the user does not help us here as we don't know how many of the subscribed events actually match our filter. If the provided filter does not match any events, we will iterate through all of the events in the DB.

Local benchmark for processing ~93,000 events in FireFly was ~30 seconds on this API with a subscription that matches no events (I.e. going through all of the events and running checks on them). An internal limit of <100 takes roughly 30-45 seconds to complete, higher figures tend more towards ~30 seconds. Increasing the internal limit past 200 does not substantially decrease the time required for the query.


SamMayWork:
WRT `O(n)` performance:

 - Since there is no existing store of events that match subscriptions, when we get a request, we must index all of the events to find which of those match the subscription that has been provided
 - A user-provided skip/limit does not significantly reduce the amount of processing that we internally have to do since we don't know how many records would actually match the subscription filter in order to be able to do a skip
 - The most expensive possible query is a filter that matches no records, as this would require us indexing the entire DB

To my knowledge the only way around these problems would be:

 - Make a change to keep a record of events that match subscriptions, and then use this a source of truth for this API, however we then get into challenges around storing this information like ensuring that sequencing is done properly
 - Implement Regex DB semantics at a lower level (however we'd still need to index a _lot_ of records)

internal/orchestrator/subscriptions.go
```diff
@@ -134,3 +134,61 @@ func (or *orchestrator) GetSubscriptionByIDWithStatus(ctx context.Context, id st
...
+
+	// This references the final offset on the FILTERED list
+	finalDesiredOffset := inboundFilterOptions.Skip + inboundFilterOptions.Limit
+	var subscriptionFilteredEvents []*core.EnrichedEvent
+
+	internalLimit := 200
+	internalSkip := 0
+	ssFilter.Limit(uint64(internalLimit))
+
+	for len(subscriptionFilteredEvents) < int(finalDesiredOffset) {
```
peterbroadhurst, line 158: I think the easiest solution to the `o(n)` safety issue (which is a significant problem) is to have a server-configurable (not API configurable) limit on the max scan length for this.

I would suggest something like 1000 by default.

If the query reaches this limit before either:
1. Reaching the end of the table
2. Reaching the full `limit` specified

Then it returns with an error result. For example:

- `Event scan limit reached after skipping X events. Please restrict your query to a narrower range`

It's far from a perfect solution, but in the vast majority of cases I think it gives a solution.
Either:
- You have lots of events that match your subscription, so your page size will find 25 (or whatever) events in the last 1000 events
- You are hunting for one specific event, so the pre-filter (discussed in my comment above) will sub-select to less than 1000 records. Either with a time-range, or with a spear-fishing approach (like looking for one specific event)

peterbroadhurst (dismissed):
This is great @SamMayWork - very helpful addition.
Returned with some thoughts on what i hope are small additions before it merges.

internal/orchestrator/subscriptions.go
SamMayWork: The implementation for these changes is split across commits https://github.com/hyperledger/firefly/pull/1452/commits/de6d46b3ef8cbb763ed4d5668fda220ed693527e and https://github.com/hyperledger/firefly/pull/1452/commits/7e22a2cc20ebb54d4c0bfa00f406b9343dd1d116.

This approach means that:
 - Supplying a skip/limit in the initial query now applies specifically to the amount of unfiltered events that are processed on the backend and you cannot control how many results you get from the API*1
 - If the amount of records being indexed exceeds the total amount configured on the namespace, no results are returned to the user, and instead they get an error indicating that they need to refine their query

---

Crucially, this approach presents a new problem worth noting: **how do I as the caller of the API, know how to adjust my skip/limit so that I can receive the next _sequentially filtered_ event?**

Suppose 1000 unfiltered events, with 2 matching events at position [0:1], and a matching event at [999]. My first query might be `skip:0 limit:100` which would return me events 0 and 1, but how would I get event 2 which is at unfiltered index 999?

Currently the answer to this is: 'Keep adjusting the skip/limit until you find it' which is fair, but depending on the use-case could be a frustrating answer. For us being able to supply information to make this easier for users is difficult because we as FireFly don't have the intelligence required to know where that index is.

---

*1 - The skip/limit allows the user to configure how many unfiltered events are indexed, and then returns all of the matching events within that bound, we then assume that the user will then limit this down to be the number of events they want to search on their side.

internal/orchestrator/subscriptions.go
```diff
@@ -134,3 +134,61 @@ func (or *orchestrator) GetSubscriptionByIDWithStatus(ctx context.Context, id st
 
 	return subWithStatus, nil
 }
+
+func (or *orchestrator) GetSubscriptionEventsHistorical(ctx context.Context, subscription *core.Subscription, filter ffapi.AndFilter) ([]*core.EnrichedEvent, *ffapi.FilterResult, error) {
+
+	// Internally we need to know the limit/count from the inbound filter
+	inboundFilterOptions, err := filter.Finalize()
```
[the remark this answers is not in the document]
SamMayWork: Addressed in https://github.com/hyperledger/firefly/pull/1452/commits/33294f17c06298adb6be87c58f45029448a5db79 but will need a retest quickly.

SamMayWork:
Spotted a problem during testing that might be a bit tricky to solve. Given there are likely to be X number of events on the chain where X is a _high_ number, being restricted to skipping 1000 records (as is the default API behaviour) is super restrictive. There's an entirely valid use-case for skip 90,000 records and then return me 200 records matching my filter but this would not work as of right now.

I've pushed changes in this commit https://github.com/hyperledger/firefly/pull/1452/commits/52c191e9981473dfa73248386a1e033fd23d6399 which allows for a custom (and :sparkles: configurable :sparkles:) number of events to be skipped, _buuuuuut_ this doesn't play nicely with the swagger generation, so the generated API documentation isn't updated to reflect the new value. This means there's a disconnect right now between what our backend supports and what we say it supports.

Needs more thinking for a workaround...

SamMayWork:
In discussion w/ @nguyer, we've landed on some further changes for this PR given the fact that we're concerned about the usefulness of this API if we don't solve the N+1 events problem. We've agreed on the following changes:

1. a starting/end sequence ID can be provided which determines where we start/stop indexing from
2. a limit of _matching_ records can be provided which will takes precedence over the end sequence ID

This means that assuming that you hit the limit of matching events _before_ you hit the end of the sequence range, you can get the next event in the sequence by calling the API again and providing the starting sequence ID to be the sequence ID of the last record.

Changes on the way :rocket:

internal/database/sqlcommon/event_sql.go
```diff
@@ -223,5 +220,29 @@ func (s *SQLCommon) GetEvents(ctx context.Context, namespace string, filter ffap
...
+
+}
+
+func (s *SQLCommon) GetEventsInSequenceRange(ctx context.Context, namespace string, filter ffapi.Filter, startSequence int, endSequence int) (message []*core.Event, res *ffapi.FilterResult, err error) {
+	filter.Limit(0) // We calculate the limit below using the sequence IDs
+
+	cols := append([]string{}, eventColumns...)
+	cols = append(cols, s.SequenceColumn())
+
+	query := sq.Select(cols...).FromSelect(sq.Select(cols...).From(eventsTable).Where(sq.GtOrEq{
```
SamMayWork, line 242: Calling this out as something potentially worth of discussion, initially I wanted to use the `BETWEEN` operator referencing the sequence ID but squirrel (the underlying SQL construction lib we're using) does not support building queries with `BETWEEN`.

The other option was to use a sub-query and then apply a limit and offset but this was _not good_ from a performance perspective, so I've gone for a sub-query and a `WHERE` which _should_ jump straight to the correct sequence ID. Comparing the use of OFFSET and this approach showed roughly a ~140ms difference in speed.

~440ms for a query using OFFSET, searching for 1000 records between sequence ID 100,000 and 101,000
~300ms for a query using WHERE, searching for 1000 records between sequence ID 100,000 and 101,000

SamMayWork:
Comment above about the SQL performance of this query syntax, but I think we're in a good place for a re-review now :rocket:

internal/database/sqlcommon/event_sql.go
SamMayWork: In discussion w/ @nguyer there was no longer any point in using a sub query here (not that it made much sense in the first place) so this is swapped out for a way more performant query in [b6cdd32](https://github.com/hyperledger/firefly/pull/1452/commits/b6cdd3295ac1bcd0a2856712959aeb7c2b1c4af4).

Looking now at ~50ms for a response on my machine using postman and SQLite.

nguyer:
> ## [Codecov](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger) Report
> Attention: `10 lines` in your changes are missing coverage. Please review.
> 
> > Comparison is base [(`fd542c0`)](https://app.codecov.io/gh/hyperledger/firefly/commit/fd542c0b1dd74dd9d7d009f62aa7f601a136d1fa?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger) 99.99% compared to head [(`a08f80b`)](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger) 99.94%.
> 
> [Files](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger)	Patch %	Lines
> [internal/events/event_manager.go](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger#diff-aW50ZXJuYWwvZXZlbnRzL2V2ZW50X21hbmFnZXIuZ28=)	50.00%	[10 Missing ⚠️ ](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/hyperledger/firefly/pull/1452?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperledger).

Looks like codecov is just having trouble talking to GitHub. I confirmed by checking out this branch myself that it does not lower test coverage.

nguyer (commented):
Sorry for the all the back and forth on this. I think it's ending up in a really good spot though. One more question though: as the user of this API, if I don't know what the current sequence number (for the most recent event) is, how do I start using this API?

I get how to use it once we have a sequence number as a reference and we can go higher or lower from there. I think most of the time users are going to be interested in the most recent `n` events though and don't know where to start from.

SamMayWork:
> I get how to use it once we have a sequence number as a reference and we can go higher or lower from there. I think most of the time users are going to be interested in the most recent n events though and don't know where to start from.

Some initial thoughts on this here:
 1. The default behaviour could be that if no values are provided, we go to the most recent event and show the last X records
 2. We could return another value in the response object which contains the most recent event ID

I think option 2 here is just _bad_ since we're going to need to make another query to the DB to find out the current highest index and then tag this onto a request that the user might not want to make. I.e. If I was a user, and wanted the most recent event I'd need to make a request I don't care about first to get the index so I can make a request I care about - that doesn't make sense.

Option 1 is the way forward, and it's consistent with the `/events` API which this API is more or less the same as. With that in mind, I think this becomes the behaviour of the API with respect to defaults for start/end sequence IDs.

| Start Sequence ID provided (SS) | End Sequence ID provided (ES) | Calc(startsequence) | Calc(endsequence) |
| --- | --- | --- | --- |
| Y | Y | SS | ES |
| N | Y | ES-1000 (*1) | ES |
| Y | N | SS | SS+1000 (*1) |
| N | N | Most recent record sequence ID - 1000 | Most recent record sequence ID |

(*1) - Default max range is 1000 records, through configuration this value could be changed

SamMayWork:
_Right_, so it looks like there's a flaky e2e at the moment which I've seen pass and fail on different commits on this branch, I'll dig into it here for a little bit, but I'd be surprised if the changes I have made would be causing this test to fail given the test is failing during multi-party set up, initial look suggests that one of the nodes is failing to register its org.

nguyer (approved):

PR #1459, opened 2024-02-01
adding High/Critical severity vuln checks

ssmirr:
This Pull request:

1. updates the Dockerfile to check dependencies of this source code, and fail to build if high/critical severity vulnerabilities are detected
2. use `set -e` in `docker_build.sh` to ensure `make docker` has a non-zero exist code when the build fails

Both of the changes above are implemented with the help of Trivy, an open source scanning tool from Aquasec. Trivy is RedHat certified, and is being used as the default container scanner on GitLab (according [to this link](https://www.aquasec.com/products/trivy/))

### DIFF
--- a/docker_build.sh
+++ b/docker_build.sh
@@ -16,9 +16,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -e
 echo $@
 
- if [[ ! -x `which jq` ]]; then echo "Please install \"jq\" to continue"; exit 1; fi
+if [[ ! -x `which jq` ]]; then echo "Please install \"jq\" to continue"; exit 1; fi
 
 FIREFLY_BUILDER_TAG=$(cat manifest.json | jq -r '.build."firefly-builder".image')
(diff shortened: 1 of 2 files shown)


Chengxuan (approved):

docker_build.sh
```diff
@@ -16,9 +16,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -e
```
nguyer, line 19: With this flag, if the build step fails it means we don't run this command at the end:
```
docker buildx rm firefly --keep-state
```

Which means if someone attempts to build again, this command fails the next time around because the builder already exists:
```
docker buildx create --name firefly --use
```

I would suggest we add a way to always clean up the builder, regardless of the return value of the build command itself.

nguyer (changes requested):

nguyer:
All of the issues with dependencies have been addressed in this PR: https://github.com/hyperledger/firefly/pull/1454

Issue #1428, opened 2023-11-29
Intermittent unit test: `TestRequestWithBodyReplyEndToEndWithTLS`


nguyer:
I have seen the unit test `TestRequestWithBodyReplyEndToEndWithTLS` intermittently fail in GitHub actions. For example: https://github.com/hyperledger/firefly/actions/runs/7035574807/job/19146264505?pr=1427

```
=== RUN   TestRequestWithBodyReplyEndToEndWithTLS
time="2023-11-29T16:32:40Z" level=error msg="Failed to invoke webhook: Put \"https://127.0.0.1:8443/myapi/my/sub/path%3Fescape_query?dynamic-query=dynamicqueryval&my-query=myqueryval\": dial tcp 127.0.0.1:8443: connect: connection refused" webhook=
    webhooks_test.go:577: 
        	Error Trace:	/home/runner/work/firefly/firefly/internal/events/webhooks/webhooks_test.go:577
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/value.go:584
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/value.go:368
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/callbacks.go:24
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/webhooks.go:470
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/webhooks.go:539
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/webhooks_test.go:584
        	Error:      	Not equal: 
        	            	expected: "batch_pin"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(fftypes.FFEnum) (len=9) "batch_pin"
        	            	+(fftypes.FFEnum) ""
        	            	 
        	Test:       	TestRequestWithBodyReplyEndToEndWithTLS
    webhooks_test.go:578: 
        	Error Trace:	/home/runner/work/firefly/firefly/internal/events/webhooks/webhooks_test.go:578
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/value.go:584
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/value.go:368
        	            				/home/runner/work/firefly/firefly/internal/events/webhooks/callbacks.go:24
[...]

PR #1463, opened 2024-02-10
Intermittent unit tests

Philip-21:
Pr fixes for #1428


Philip-21:
@nguyer  From this pr, the error  seems to be related to connection issues when trying to invoke the webhook and this makes the  TLS client unable to establish a connection to the server ,  so i decided to remove the IP address, and utilize the port number  to see if the error exists, But the workflow still awaits approval before it runs

nguyer:
@Philip-21 Just to let you know, you can run E2E tests locally as well by running `make e2e`

Philip-21:
> @Philip-21 Just to let you know, you can run E2E tests locally as well by running `make e2e`

ok thanks

nguyer (approved):

Issue #1453, opened 2024-01-23
Status call returns error when node is not registered to a multiparty namespace


chrisbygrave:
Before [this change](https://github.com/hyperledger/firefly/commit/4696ad1c82194580c3c8bf01632c020db86a74e7#diff-e097a0bac30f72e5cc6978dc4e6336e1377310f01235622293baf87758df9302R111) I could register my org using the `org/self` API, and then subsequently register my node using `self/node`. In the time between the two registrations completing successfully, FireFly ran normally and I could call `/status` to check on the completion of the registration calls.

After the above change, I am getting an error from the `/status` call:
```
{"error":"FF10225: Unable to resolve the local node. Please ensure node.name is configured"}
```

It looks like the change has resulted in the missing node name being treated as an error rather than an acceptable state. This makes client logic more difficult to manage - if registration requires two separate asynchronous calls, then being in a state where one has completed without the other needs to be tolerated.

chrisbygrave:
Now fixed by https://github.com/kaleido-io/firefly/pull/102

PR #1467, opened 2024-02-15
fix: Status call returns properly when node is not registered but organisation is

SamMayWork:
ref: https://github.com/hyperledger/firefly/issues/1453 and [this change](https://github.com/hyperledger/firefly/commit/4696ad1c82194580c3c8bf01632c020db86a74e7#diff-e097a0bac30f72e5cc6978dc4e6336e1377310f01235622293baf87758df9302R111)

When a namespace is registered to the organisation but the node has not been registered, if the name has not been set (i.e. it's `""`) then we throw an error. If a name _is_ set, then we return `nil` because while there is no information on the node (as it's not registered) it is possible for the node to be registered.

Side note: getting an environment in which to test this was a colossal pain and largely undocumented, so I'm going to put a comment below with how I got a testing configuration in place to test this out.

### DIFF
--- a/internal/identity/identitymanager.go
+++ b/internal/identity/identitymanager.go
@@ -105,13 +105,12 @@ func ParseKeyNormalizationConfig(strConfigVal string) int {
 
 func (im *identityManager) GetLocalNode(ctx context.Context) (node *core.Identity, err error) {
 	nodeName := im.multiparty.LocalNode().Name
-	nodeDID := fmt.Sprintf("%s%s", core.FireFlyNodeDIDPrefix, nodeName)
-	node, _, err = im.CachedIdentityLookupNilOK(ctx, nodeDID)
-
-	if nodeName == "" && node == nil {
+	if nodeName == "" {
 		return nil, i18n.NewError(ctx, coremsgs.MsgLocalNodeNotSet)
 	}
 
+	nodeDID := fmt.Sprintf("%s%s", core.FireFlyNodeDIDPrefix, nodeName)
+	node, _, err = im.CachedIdentityLookupNilOK(ctx, nodeDID)
 	return node, err
 }
 
--- a/internal/identity/identitymanager.go
+++ b/internal/identity/identitymanager.go
@@ -1,4 +1,4 @@
-// Copyright © 2023 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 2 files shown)


SamMayWork:
### Testing setup

To test this out we need a multi-party network configured and running, and then need to start a local node which is _not_ registered to the network which we can then register to the network. Bonus points if we can run it through the VS Code debugger so we can dive into the code as/when. The steps to configure this are:

1. Create a new stack using the `ff` CLI and start the stack
2. Go into the stack directory (`~/.firefly/stacks/<stack_name>` by default) and open it up in your favourite text-editor
3. From the runtime directory copy the configuration for one of the existing FireFly namespaces e.g. `firefly_core_1.yml`
4. Create a new file for your new namespace and paste the contents from the existing namespace
5. Change the DB location URL to be a location on your local machine with an SQLite DB file
6. Change Docker URLs (i.e. http://ipfs_0) to be local (127.0.0.1) address with correct ports
7. Change ports on the new configuration to be distinct from the ports for the existing namespace
8. Run the command `ff accounts create <stack_name>` to generate a new account for your namespace to use
9. Change the node name and org name in the new configuration to be something distinct from the other namespace
10. Provide the key generated in step 8 as the key values for default key and org key
11. Verify that the contract address in the new configuration is the same as it is in the other namespaces config
12. Change your `launch.json` in FireFly to point to the new configuration file

At this point, you can run FireFly core through the VS Code debugger and you'll get a multi-party configured namespace which has not yet registered its org or node.

14. Navigate to the swagger API page for your newly created namespace
15. Call `/network/organizations/self` to register the org for the namespace
16. Call `/network/nodes/self` to register the node for the namespace

In the specific case of this issue, calling the `/status` API between steps 15 and 16 are what caused the issue this PR resolves.

internal/identity/identitymanager.go
```diff
@@ -105,13 +105,13 @@ func ParseKeyNormalizationConfig(strConfigVal string) int {
...
 	nodeName := im.multiparty.LocalNode().Name
-	if nodeName != "" {
-		nodeDID := fmt.Sprintf("%s%s", core.FireFlyNodeDIDPrefix, nodeName)
-		node, _, err = im.CachedIdentityLookupNilOK(ctx, nodeDID)
-	}
-	if err == nil && node == nil {
+	nodeDID := fmt.Sprintf("%s%s", core.FireFlyNodeDIDPrefix, nodeName)
+	node, _, err = im.CachedIdentityLookupNilOK(ctx, nodeDID)
+
+	if nodeName == "" && node == nil {
```
EnriqueL8, line 111: You could simply check if `nodeName==""` and then return the error, if not just return node and err

SamMayWork: :rocket:

nguyer (approved):

PR #1478, opened 2024-03-09
Fix golangci issues

alex-semenyuk:


nguyer:
Hi @alex-semenyuk, thanks for opening this.

Golangci already runs as a part of the existing `make build` job here: https://github.com/hyperledger/firefly/blob/577e8c47680c6230209a74829921a9c427766af8/Makefile#L15-L16

Is there something that this is not catching that you are trying to fix?

alex-semenyuk:
@nguyer Thanks for pointing, then it's suspicious that it doesn't detect all  https://github.com/hyperledger/firefly/actions/runs/8213258671/job/22464445082

.github/workflows/go.yml
```diff
@@ -8,6 +8,21 @@ on:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
+          cache: false
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v4
+        with:
+          version: v1.54
+          args: --timeout=3m
```
nguyer, line 24: The linter will only flag copyright date issues if the file has been changed more recently than the copyright year in the file. The GitHub checkout action will not download the entire git history by default unless you specify `fetch-depth: 0` https://github.com/hyperledger/firefly-ui/releases/tag/v1.3.0 If you do not include this field, and you were to inspect the file system that git checked out, every file would have an edit/creation date that matched the date/time of the commit that it checked out. Thus the linter thinks every file was changed in the current year.

I think because of this, your job here is flagging many false negatives. I think the existing golangci job that runs as part of the Makefile should be sufficient. And the existing GitHub actions that we have checkout the full history to avoid this problem.

internal/batch/batch_manager_test.go
```diff
@@ -81,15 +81,15 @@ func TestE2EDispatchBroadcast(t *testing.T) {
 		h := sha256.New()
 		nonceBytes, _ := hex.DecodeString(
 			"746f70696331",
-		/*|  topic1   | */
+			/*|  topic1   | */
```
alex-semenyuk, line 84: Fix formatting stuff

internal/database/sqlcommon/event_sql_test.go
```diff
@@ -322,10 +322,9 @@ func TestGetEventsInSequenceRangeBuildQueryFail(t *testing.T) {
 
 func TestGetEventsInSequenceRangeShouldCallGetEventsWhenNoSequencedProvidedAndThrowAnError(t *testing.T) {
 	s, mock := newMockProvider().init()
-	mock.ExpectQuery("SELECT .*").WillReturnRows(sqlmock.NewRows([]string{"id", }).AddRow("only one"))
+	mock.ExpectQuery("SELECT .*").WillReturnRows(sqlmock.NewRows([]string{"id"}).AddRow("only one"))
 	f := database.EventQueryFactory.NewFilter(context.Background()).And()
 	_, _, err := s.GetEventsInSequenceRange(context.Background(), "ns1", f, -1, -1)
 	assert.NotNil(t, err)
 	assert.NoError(t, mock.ExpectationsWereMet())
 }
```
alex-semenyuk, line 330: Add EOF

internal/database/sqlcommon/event_sql_test.go
```diff
@@ -322,10 +322,9 @@ func TestGetEventsInSequenceRangeBuildQueryFail(t *testing.T) {
 
 func TestGetEventsInSequenceRangeShouldCallGetEventsWhenNoSequencedProvidedAndThrowAnError(t *testing.T) {
 	s, mock := newMockProvider().init()
-	mock.ExpectQuery("SELECT .*").WillReturnRows(sqlmock.NewRows([]string{"id", }).AddRow("only one"))
+	mock.ExpectQuery("SELECT .*").WillReturnRows(sqlmock.NewRows([]string{"id"}).AddRow("only one"))
```
alex-semenyuk, line 325: Not required comma

.github/workflows/go.yml
alex-semenyuk: @nguyer Got it.
I removed golangci-lint from github actions. It might be useful to merge it. if user touches this files he/she will need to fix this issues. It is not a big deal but still can be annoying. I goes through changes and marked file where issue is different then wrong year.

Issue #1270, opened 2023-04-12
Batches with invalid contents should be flagged an skipped during aggregation


awrichar:
Currently if FireFly receives a batch with any malformed contents (such as a mismatched key due to #1175), it writes the batch to the database but none of the messages or data.
https://github.com/hyperledger/firefly/blob/057a7af8646426bc9d79a95b8d066838e3e6bc43/internal/events/persist_batch.go#L45

However, the saved batch has no marking to indicate that the contents were invalid. Therefore every time the aggregator rewinds to this batch, it will stall with "message not yet available".
https://github.com/hyperledger/firefly/blob/057a7af8646426bc9d79a95b8d066838e3e6bc43/internal/events/aggregator.go#L467

When batch contents are skipped in this fashion, FireFly should actually mark the batch somehow, and when the aggregator attempts to process a message from that batch, it should simply mark the pin dispatched and move on. This is different from rejecting the message, because the message will never even be inserted to the database - but there's no point in blocking the pin forever, since the message will never be valid.

awrichar:
One follow-on observation to note - we are performing certain batch validation checks _only_ in the handler for shared storage downloads.

Ideally the exact same checks would be performed on a batch 1) before sending it, and 2) immediately after downloading it from either DX or shared storage. Anything we can do to unify those validation paths would help, because even if we fix the root issue noted above, there's a secondary problem in that the sender and the receivers are in different states (sender thinks the batch was ok).

Issue #1127, opened 2023-01-05
Failure to confirm broadcasts after receiving a broadcast without data


shorsher:
If a firefly node sends a broadcast without any data, the rest of the network seems to get into a weird state.

My observations:
1. The sending node will have the broadcast show up as `confirmed`, but other nodes do not seem to receive the message. 
2. If other nodes send a broadcast, it will be stuck in `sent` and never reach confirmed.

Example request:
```
curl -X 'POST' \
  'http://127.0.0.1:5000/api/v1/messages/broadcast' \
  -H 'accept: application/json' \
  -H 'Request-Timeout: 2m0s' \
  -H 'Content-Type: application/json' \
  -d '{
  "header": {
    "tag": "TEST"
  }
}'
```

awrichar:
Tried to reproduce. I saw the message get `confirmed` by the sender, but just never received by anyone else. The receiving node clearly discarded it:
```
[2023-01-26T18:23:32.805Z] ERROR Invalid batch '7f95a2cb-a752-4df6-bcdc-5949808789fd'. Missing messages (1) or data (0) dbtx=bJfwsnmG ns=default pid=1 role=event-manager
```

Biggest question is which is the correct behavior - don't allow sending an empty batch, or allow receiving an empty batch? It appears that we're inconsistent right now.

PR #1487, opened 2024-04-01
Allow cancelling a batch that is stuck in dispatch

### LINKED ISSUE #1446: Close FIR-17
This issue is to track the work required to close out FIR-17. Specifically we've identified that we may need to create an API to allow someone to recover a specific topic if we've spent a nextpin on a transaction that will never succeed. This would currently block the topic forever after that point.

awrichar:
Batch manager status can already be queried with "/status/batchmanager". The response includes a flag to say if a processor is blocked, along with the ID of the flushing batch, and the last error message.

The new API "/batches/{batchid}/cancel" can now be used to cancel a batch. This is currently only valid for batches with transaction type "contract_invoke_pin". It will mark all messages in the batch as "cancelled", and for private messages, will send a new batch of type "batch_pin" containing gap fill messages. These messages have no data, but will have a CID pointing to the original (failed) message, and will consume nonces to allow the topic to become unblocked.

Fixes #1446 

~~Note: test coverage not yet complete~~


internal/events/aggregator.go
```diff
@@ -638,6 +647,9 @@ func (ag *aggregator) readyForDispatch(ctx context.Context, msg *core.Message, d
 		} else if valid {
 			action = core.ActionConfirm
 		}
+
+	default:
+		action = core.ActionConfirm
```
awrichar, line 652: Unrelated find - we were rejecting messages with 0 data elements simply because the default `action` was 0, which means "reject". That didn't seem correct - if there were no special conditions to parse, I think the default should be "confirm" (which includes messages that don't carry data).

If we specifically want to reject messages that contain no data items, it should be a separate branch with an explicit reject reason.

internal/batch/batch_processor.go
```diff
@@ -568,11 +601,53 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
...
+				}
+			}
+			return true, err
 		})
 	})
 }
 
+func (bp *batchProcessor) writeGapFill(ctx context.Context, msg *core.Message) error {
+	// Gap fill is only needed for private messages
+	if msg.Header.Type != core.MessageTypePrivate {
```
awrichar, line 630: I went back and forth on whether to send gap fill messages for broadcasts. Ultimately they are needed for private messages for two reasons:
1. This sender has "spent" a nonce, and may have calculated and used nonces after it. All recipients need to know about the spent nonce so they can process later messages and get back in sync.
2. All recipients will have received the message payload via data exchange, and will have created a message in "pending" state. They need to know to move those messages to "cancelled" so that they don't stay pending forever.

Neither of these is true for broadcasts. Therefore it felt like cancelling a broadcast can be a "local only" operation, and broadcasting a gap fill to everyone might be more confusing.

internal/batch/batch_manager.go
```diff
@@ -588,3 +594,30 @@ func (bm *batchManager) LoadContexts(ctx context.Context, payload *DispatchPaylo
...
+	}
+	batch, err := bm.data.HydrateBatch(ctx, bp)
+	if err != nil {
+		return err
+	}
+	if len(batch.Payload.Messages) == 0 {
+		return i18n.NewError(ctx, coremsgs.MsgErrorLoadingBatch)
+	}
+	msg := batch.Payload.Messages[0]
+	processor, err := bm.getProcessor(msg.Header.TxType, msg.Header.Type, msg.Header.Group, msg.Header.SignerRef.Author, false)
```
peterbroadhurst, line 615: Just checking - what is it that means we're sure that a processor will be active in this case (so we don't need to create one)?

peterbroadhurst: ... ah ok, I can see that `cancelFlush` is only going to do anything meaningful if it's busy trying to flush. So the `processor == nil` case means there's nothing to do.

internal/batch/batch_manager.go
```diff
@@ -588,3 +594,30 @@ func (bm *batchManager) LoadContexts(ctx context.Context, payload *DispatchPaylo
...
+	if err != nil {
+		return err
+	}
+	if processor == nil {
+		return i18n.NewError(ctx, coremsgs.MsgCannotCancelBatchState)
```
peterbroadhurst, line 620: The message text is very clear, but the variable name `MsgCannotCancelBatchState` isn't - maybe instead:
`MsgBatchProcessorNotActive` ?

internal/batch/batch_processor.go
```diff
@@ -277,6 +282,29 @@ func (bp *batchProcessor) captureFlushError(err error) {
 	fs.LastFlushError = err.Error()
 }
 
+func (bp *batchProcessor) cancelFlush(ctx context.Context, id *fftypes.UUID) error {
+	bp.statusMux.Lock()
+	defer bp.statusMux.Unlock()
+	fs := &bp.flushStatus
+
+	if bp.conf.txType != core.TransactionTypeContractInvokePin {
+		return i18n.NewError(ctx, coremsgs.MsgCannotCancelBatchType)
```
peterbroadhurst, line 291: Maybe include the type as an insert?

internal/batch/batch_processor.go
```diff
@@ -277,6 +282,29 @@ func (bp *batchProcessor) captureFlushError(err error) {
...
+	}
+	if !id.Equals(fs.Flushing) {
+		return i18n.NewError(ctx, coremsgs.MsgCannotCancelBatchState)
```
peterbroadhurst, line 294: Maybe worth a separate error for this, including the ID of the one that _is_ being flushed (so the user can try and cancel that one instead)?

internal/batch/batch_processor.go
```diff
@@ -568,11 +601,54 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
...
+			Pins:           msg.Pins, // reuse any assigned pins to fill the nonce gap
+		},
+	}
+	gapFill.Header.ID = fftypes.NewUUID()
+	gapFill.Header.CID = msg.Header.ID
+	gapFill.Header.Tag = core.SystemTagGapFill
+	gapFill.Header.TxType = core.TransactionTypeBatchPin
+	err := gapFill.Seal(ctx)
+	if err == nil {
+		err = bp.data.WriteNewMessage(ctx, &data.NewMessage{Message: gapFill})
```
peterbroadhurst, line 647: One question that comes to my mind, which I wonder if you tested - what if there are other messages queued with pins after the gap-fill already?

I think the batch processor will move onto those messages and batch+send them with next-pins _after_ the gap fill, before the gap-fill arrives. So on the receiving side(s) they will receive the gap-fill out of order.

I think this just works itself out, and is all ok - but needs testing.

internal/batch/batch_manager.go
awrichar: We can only cancel a batch that is stuck in a processor. So if the relevant processor is not active, we simply return an error.

internal/events/aggregator.go
peterbroadhurst: I think we might have a bug reference for this - would be good to mark that as resolved... will see if I can find it.

peterbroadhurst: Found this instead, which might bear you taking a quick look to see if now is the time to address it (in 1.3, but not this PR): https://github.com/hyperledger/firefly/issues/1270

... but I couldn't find the send-empty-message bug report. So maybe I imagined it.

peterbroadhurst (commented):
A few minor questions as I went through @awrichar - but this looks absolutely great

internal/events/aggregator.go
awrichar: You're likely thinking of https://github.com/hyperledger/firefly/issues/1127, which was resolved - but I think this crept back in as a different flavor of the same issue during one of the refactors in this file.

internal/apiserver/route_post_batch_cancel_test.go
```diff
@@ -0,0 +1,47 @@
+// Copyright © 2022 Kaleido, Inc.
```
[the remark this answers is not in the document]
awrichar: Nope, it doesn't. Most of our test files have dates of 2021-2022.

internal/batch/batch_processor.go
awrichar: I went to test this, and it does not sort itself out automatically. You can manually sort it out with another "rewind" call to the first batch that was blocked on the gap fill.

I'll try to figure out if this rewind can happen automatically. The problem is that rewinds are only effective if you rewind to an _undispatched_ pin (they are ignored if you request a rewind to a pin that has been dispatched). The gap filled pin gets marked _dispatched_ as part of this new logic - so we can't rewind to that one. We need to know if there's another pin queued up after it that is now blocked and requires a rewind...

peterbroadhurst: > that messages are confirmed on chain in the order nonces were assigned by the sender.

Yes, this is both a design assumption, and a requirement currently for a "well behaved" actor in a privacy group.

awrichar: I confirmed with @peterbroadhurst that the following is expected to be a fundamental obligation of the transaction manager: transactions will _always_ be confirmed on the chain in the order they were submitted _as long as they come from a single signing key_. This is critical for the above "well behaved" privacy group behavior to be reliable.

Therefore, all messages from a single key to a single privacy group need to be on the same thread, and therefore we have a problem in the current design where we have 2 batch processors assigning nonces from the same group (one for regular private messages, another for custom contract private messages). I think I need to combine these processors into one - can't remember the original reason I split them... likely I thought it was actually solving a problem at the time. I don't see any reason they can't be a single processor though.

awrichar: Updated so that "batch_pin" and "contract_invoke_pin" messages share a batch processor. Also altered so that the "gap fill" batch is sent out immediately during the dispatch phase, instead of trying to queue it for the next assembly loop (which had a lot of problems with reordering).

peterbroadhurst (approved):
Thanks @awrichar for the extra time on this one, and great to see the conclusion.

PR #1489, opened 2024-04-07
Fix broken links at docs 

alex-semenyuk:


alex-semenyuk:
This html links don't work on github https://github.com/hyperledger/firefly/blob/main/docs/tutorials/broadcast_data.html Is it ok?
It seems like there are mix of links - some of them ends with html like https://github.com/hyperledger/firefly/blob/main/docs/tutorials/broadcast_data.html; some of them ends with md like https://github.com/hyperledger/firefly/blob/main/docs/tutorials/custom_contracts/ethereum.md (then it works on github as well)

nguyer:
> This html links don't work on github https://github.com/hyperledger/firefly/blob/main/docs/tutorials/broadcast_data.html Is it ok?

Ah okay. The docs site is generated by Jekyll and takes all of these files in and generates static HTML for them. So these links do actually work on the real docs site. I don't generally browse the docs right in GitHub as there are lots of things like tables of contents and external file embeds that don't show up there.

However, I'm happy to merge this, because it should not make any difference to the generated docs site, and it does improve the GitHub experience. Thanks.

nguyer (approved):

PR #1493, opened 2024-04-11
Add documentation to outline how EVM revert errors are handled

matthew1001:
To make it easier to preview the changes, here are a couple of screenshots of the resulting docs:

image

---

image

### DIFF
--- a/docs/architecture/blockchain_connector_framework.md
+++ b/docs/architecture/blockchain_connector_framework.md
@@ -192,3 +192,84 @@ Some high architectural principals that informed the code:
   - Determines the final order based on order of confirmation on the blockchain
 
 [![FireFly Connector Toolkit Event Streams](../images/firefly_connector_toolkit_event_streams.jpg)](../images/firefly_connector_toolkit_event_streams.jpg)
+
+## Transaction error messages
+
+The receipt for a FireFly blockchain operation contains an `extraInfo` section that records additional information about the transaction. For example:
+
+```
+"receipt": {
+  ...
+  "extraInfo": [
+    {
+      {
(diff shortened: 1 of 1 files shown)


nguyer (changes requested):
This is great detail, thank you @matthew1001! My one piece of feedback would be that this is all very Ethereum/EVMConnect specific info, and it is currently placed in the context of a very generic "Blockchain Connector Framework" document.

I think it's fine to have in this doc, but I think we need to add some wording or context to describe that this is information is specific to _one specific_ application of the Blockchain Connector Framework (one specific connector)

matthew1001:
> I think it's fine to have in this doc, but I think we need to add some wording or context to describe that this is information is specific to _one specific_ application of the Blockchain Connector Framework (one specific connector)

Yes good point @nguyer. I've done some refactoring to try and be clearer about what is generic FF information and what is ff-evmconnect-specific. For example the introductory paragraph now says:

> The errorMessage field can be be set by a blockchain connector to provide FireFly and the end-user with more information about the reason why a tranasction failed. The blockchain connector can choose what information to include in errorMessage field. It may be set to an error message relating to the blockchain connector itself or an error message passed back from the blockchain or smart contract that was invoked.

and the message format paragraph now says:

> Format of a firefly-evmconnect error message
> The following section describes the way that the firefly-evmconnect plugin uses the errorMessage field. This serves both as an explanation of how EVM-based transaction errors will be formatted, and as a guide that other blockchain connectors may decide to follow.

plus I've tweaked other wording here or there to try to make it clearer.

Let me know if you think that provides enough distinction between what is general behaviour and what is `firefly-evmconnect` behaviour.

nguyer (approved):
This looks good now! Thank you!

PR #1495, opened 2024-04-12
Update token connectors

nguyer:
This fixes a regression with MTLS certs when the token connector checks to see if the blockchain connector is up.


EnriqueL8 (approved):
LGTM!

PR #1499, opened 2024-04-19
Update the FF error docs to be in a more suitable place in the nav and give clearer out-of-the-box examples

matthew1001:
Updates to improve https://github.com/hyperledger/firefly/pull/1499


peterbroadhurst (approved):
This is great, thanks @matthew1001 for the extra rev here 🙇

PR #1502, opened 2024-04-24
Send start msg on reconnect for token connectors

EnriqueL8:
When the token connector is restarted, it needs to be sent the start namespace message


nguyer (approved):
Great find and fix! Thank you and sorry I missed this first time around in the refactor.

PR #1509, opened 2024-05-01
Add multiparty status API

chrisbygrave:


pkg/core/namespace_status.go
```diff
@@ -65,3 +78,29 @@ type NamespaceStatusMultiparty struct {
...
+
+type NamespaceMultipartyStatus struct {
+	Enabled   bool                                 `ffstruct:"NamespaceMultipartyStatus" json:"enabled"`
+	Org       NamespaceMultipartyStatusOrg         `ffstruct:"NamespaceMultipartyStatus" json:"org"`
+	Node      NamespaceMultipartyStatusNode        `ffstruct:"NamespaceMultipartyStatus" json:"node"`
+	Contracts *MultipartyContractsWithActiveStatus `ffstruct:"NamespaceMultipartyStatus" json:"contracts,omitempty"`
+}
+
+type MinimalListenerCheckpoint struct {
+	Block int64 `json:"block"`
```
awrichar, line 100: We've somewhat avoided the term "block" outside of specific plugins like the Ethereum plugin (have generally used more general terms like "event" and "offset" and "checkpoint"). I don't if anything else is more appropriate here, but just pointing it out for consideration.

internal/orchestrator/status.go
```diff
@@ -129,3 +133,123 @@ func (or *orchestrator) GetStatus(ctx context.Context) (status *core.NamespaceSt
...
+		return nil, err
+	}
+	mpStatus.Contracts.Active.Status = listenerStatus
+
+	if status.Org.Registered {
+		mpStatus.Org.Status = core.NamespaceRegistrationStatusRegistered
+		if status.Node.Registered {
+			mpStatus.Node.Status = core.NamespaceRegistrationStatusRegistered
+		} else {
+			msg, err := or.getRegistrationMessage(ctx)
```
awrichar, line 214: It's interesting to me that we only query "bad/stuck" registration messages. So we do _not_ populate `RegistrationMessageID` in the case that registration was successful.

I think we should either a) populate it in all cases, or b) rename the field so it's clear that it only returns the ID of a non-confirmed registration.

internal/orchestrator/status.go
```diff
@@ -129,3 +133,123 @@ func (or *orchestrator) GetStatus(ctx context.Context) (status *core.NamespaceSt
...
+		return nil, err
+	}
+	if len(msgs) > 0 {
+		return msgs[0], nil
+	}
+	return nil, nil
+}
+
+// Ensure the given registration message correlates with the expected type
+func (or *orchestrator) checkRegistrationType(ctx context.Context, msg *core.MessageInOut, expectedType core.IdentityType) (match bool, err error) {
```
awrichar, line 157: Was this check in response to a specific error scenario we've seen? I'm trying to think why there would be a pending registration message, but the contents would be inconsistent.

internal/coremsgs/en_error_messages.go
```diff
@@ -307,4 +307,7 @@ var (
 	MsgCannotCancelBatchType                 = ffe("FF10466", "Cannot cancel batch of type: %s", 400)
 	MsgErrorLoadingBatch                     = ffe("FF10467", "Error loading batch messages")
 	MsgBatchNotDispatching                   = ffe("FF10468", "Batch %s is not currently dispatching - current: %s", 400)
+	MsgNoMessageInlineData                   = ffe("FF10469", "Unable to check message registration data for org %s", 500)
```
awrichar, line 310: Might suggest `MsgNoRegistrationMessageData`

internal/orchestrator/status.go
chrisbygrave: No, it was purely theoretical - it felt clumsy before I added this that we would find a registration message and just _assume_ it was of the correct type. It's technically possible that a pending node registration message could be there without the org being registered, and the API would then return the wrong response.

I admit it could be overkill...

pkg/core/namespace_status.go
chrisbygrave: The term exists inside a couple of the blockchain connectors but there is no need to let it bleed out here - I'll tidy up.

internal/orchestrator/status.go
chrisbygrave: Fair point - I didn't populate it in the success case purely because my use case didn't require it, but I think option a) makes for the most simple, understandable API. I'll update the PR.

awrichar (approved):

EnriqueL8 (approved):
LGTM and merging! Thanks @chrisbygrave

PR #1518, opened 2024-05-28
Adding check URLs to GH actions

dwertent:
I've noticed some invalid URLs in the repository. This GitHub Action will help detect these invalid URLs when opening a pull request or during a scheduled scan. I recommend setting `force_pass` to `false` to prevent the approval of pull requests containing invalid URLs. You can see an example [here](https://github.com/dwertent/firefly/actions/runs/9275582818/job/25520636370#step:4:351).

image

### DIFF
--- a/doc-site/docs/reference/events.md
+++ b/doc-site/docs/reference/events.md
@@ -187,6 +187,9 @@ Once you have configured the blockchain event listener, every event detected
 from the blockchain will result in a FireFly event delivered to your application
 of type `blockchain_event_received`.
 
+As of 1.3.1 a group of event filters can be established under a single topic when supported by the connector, which has benefits for ordering. 
+See [Contract Listeners](../reference/types/contractlistener.md) for more detail
+
 Check out the [Custom Contracts Tutorial](../tutorials/custom_contracts/index.md) for
 a walk-through of how to set up listeners for the events from your smart contracts.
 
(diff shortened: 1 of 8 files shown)


dwertent:
Working on this PR made me realize that the package should support SARIF format output. As a result, I opened [this PR](https://github.com/urlstechie/urlchecker-python/pull/91) and [this PR](https://github.com/urlstechie/urlchecker-action/pull/109) to enhance the informativeness of the GitHub Action results.

EnriqueL8 (changes requested):
Thanks for the contribution here

I'm torn on the force pass always being `true` as we won't notice if thing have broken without looking at logs but at the same time I don't want this to gate PRs into the repo 

So if there is a way to be notified that it has failed on the schedule it would be good.

As part of this PR would be a great contribution if the existing URL highlighted in the description are fixed as well

.github/workflows/url_checker.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    - name: Run URLChecker
+      uses: urlstechie/urlchecker-action@0.0.34
+      with:
+        file_types: .md
+        force_pass: true
+        save: results.csv
+        exclude_patterns: 127.0.0.1,localhost,ipfs_,ca_org,ethconnect_,dataexchange_
```
EnriqueL8, line 20: Also add `tokens_0_0` here

EnriqueL8:
I can't edit my review seems like you put in your description `I recommend setting force_pass to false to prevent the approval of pull requests containing invalid URLs.` that is valid to me then

PR #1521, opened 2024-06-07
Document the confirm option on the generated invoke APIs

peterbroadhurst:
- The description was missing on the generated `/invoke` APIs
- The description incorrectly referred to a "message" on the base `/contracts/invoke` API


doc-site/docs/swagger/swagger.yaml
```diff
@@ -5069,7 +5069,8 @@ paths:
       description: Invokes a method on a smart contract. Performs a blockchain transaction.
       operationId: postContractInvoke
       parameters:
-      - description: When true the HTTP request blocks until the message is confirmed
+      - description: When true the HTTP request blocks until the blockchain transaction
+          is confirmed
```
EnriqueL8, line 5073: Not sure why there is an end line here?

EnriqueL8 (approved):
LGTM - one question on the swagger that might be fine

doc-site/docs/swagger/swagger.yaml
peterbroadhurst: Sorry @EnriqueL8 - I'm being slow. What's the problem with the string?

`"When true the HTTP request blocks until the blockchain transaction is confirmed"`

peterbroadhurst: oh, sorry - I see you're asking about the YAML formatting.

That's just a choice of the YAML serializer when the string exceeds a certain length, to use an indented newline to continue the string

PR #1525, opened 2024-06-17
fix: parallelize ensuring active pools on startup

EnriqueL8:
When a large number of token pools are active and namespace is restarted, it take a long time to activate one by one each token pool. This PR spins a new go routine for each API call.


internal/tokens/fftokens/fftokens.go
```diff
@@ -716,16 +717,31 @@ func (ft *FFTokens) handleNamespaceStarted(ctx context.Context, data fftypes.JSO
...
+			return err
+		})
+	}
 
+	// g.Wait waits for all goroutines activating token pools to complete
+	// and returns the first non-nil error returned
+	// by one of the goroutines.
+	if err := g.Wait(); err != nil {
+		// The above handleMessage will retry if errors occur
+		return err
```
peterbroadhurst, line 742: This is a change in behavior - the previous `ft.EnsureTokenPoolActive(ctx, currentPool)` code did not crash the namespace start in the case it failed to activate a token pool.

It might be a positive change in behavior, but it would need some analysis/thinking (probably from @awrichar?) as to why the previous code intentionally did _not_ fail on this condition before.

internal/tokens/fftokens/fftokens.go
```diff
@@ -716,16 +717,31 @@ func (ft *FFTokens) handleNamespaceStarted(ctx context.Context, data fftypes.JSO
 	// Make sure any pools that are marked as active in our DB are indeed active
 	namespace := data.GetString("namespace")
 	log.L(ctx).Debugf("Token connector '%s' started namespace '%s'. Ensuring all token pools active.", ft.Name(), namespace)
+
+	g, ctx := errgroup.WithContext(ctx)
```
peterbroadhurst, line 721: I feel like using `SetLimit() on the concurrency would be better than a completely unbounded parallelism against the token connector:
https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit

peterbroadhurst (changes requested):

internal/tokens/fftokens/fftokens.go
EnriqueL8: Good idea - will set that to the number of pools to activate

internal/tokens/fftokens/fftokens.go
EnriqueL8: Yeah I thought about that, it felt weird to me that it didn't fail on startup... Would love to hear @awrichar thoughts on this one

EnriqueL8:
Closing this for now as not being worked on.

PR #1526, opened 2024-06-17
Use codecov github action

EnriqueL8:
Move to using the GitHub action as the bash script is not using the token correctly


EnriqueL8:
This still doesn't work, will come back to it

.github/workflows/go.yml
```diff
@@ -25,8 +25,10 @@ jobs:
       - name: Build and Test
         run: make
 
-      - name: Upload coverage
-        run: bash <(curl -s https://codecov.io/bash)
+      - uses: codecov/codecov-action@v4
+        with:
+          codecov_yml_path: ./codecov.yml
+          token: ${{ secrets.CODECOV_TOKEN }}
```
peterbroadhurst, line 31: I'm not sure we've set up codecov tokens on other repos in HL previously.

I think this PR is pointing out that codecov is trying to deprecate the ability to use it without a token, and as such we might need one.

That might need a conversation with the HL team potentially, as while I believe the free tier would be fine still on codecov.io, there would need to be a user that could own the github account that is associated with the token.

peterbroadhurst (changes requested):

.github/workflows/go.yml
EnriqueL8: The repo has a secret `CODECOV_TOKEN` set for the Github Action to use and the previous `bash` step used the token already. I have not idea who/how that token ended up in the repository, doesn't seem to be a HL thing based on not seeing anything on discord. I do wonder how we created an account against codecov and generated a token

EnriqueL8:
This now worked https://app.codecov.io/github/hyperledger/firefly/commit/582db4c8814f471d9df0b2c8df30de20f9862dc7 

@ryjones changed the token inside the repo with a Hyperledger CodeCov token and gave me access to CodeCov as well

.github/workflows/go.yml
EnriqueL8: Conversation with HL https://discord.com/channels/905194001349627914/905205991073804308/1252995889522282599

peterbroadhurst (approved):

peterbroadhurst:
Super, thanks @EnriqueL8 and @ryjones

PR #1529, opened 2024-06-19
fix: do not start multiple blockchain plugin on retry of namespace start

### LINKED ISSUE #1527: Start up retry does not clean up previous failed attempts
As part of startup if the namespace manager fails to start a namespace it will retry until a limit to restart it, unfortunately as part of this process it will start other go routines to listen to event for example or different types of loop that listen to web sockets to other plugins. If one of the plugin fails to start the whole namespace fails and these plugins are still running go routine that is not cancelled as the context of namespace is not cancelled... That is a problem because you get multiple copies of plugins running and can cause all sort of duplicate events for example.

Fixing this would involve most probably handling the context better and when retrying to start namespace to stop previos namespace to clear already setup plugins or have a flag when a plugin can only be initialised once?

EnriqueL8:
Fixes https://github.com/hyperledger/firefly/issues/1527

Make sure that in the case where the NS fails to restart because of another plugin has failed do not start the blockchain plugin again when it was successfully started before.

Played around with the idea of using context to clear everything but it didn't make sense. Also I don't think a mutexLock is needed for this. 

If you are wondering how all the other plugins work it's because at initialisation part those plugins return a struct that we then use as part of starting the managers/plugins.  So we can check for the existence of that struct and not re-initialise that manager/plugin. In the case of the blockchain plugin, it's a shared struct and we just start a new event streams in the init part.


EnriqueL8:
Fixes https://github.com/hyperledger/firefly/issues/1527

peterbroadhurst (approved):
Thanks for the great explanation 👍

Issue #1531, opened 2024-06-20
Migration to v1.3.0 can cause missed events when contract listeners are listening from latest


EnriqueL8:
As part of the v1.3.0 release, we rearchitected the way event streams from the transaction managers are configured against the namespaces in FireFly. FireFly core will now create *one* event stream per namespace and as such we wrote code to migrate the existing contract listeners on the previous event stream to the corresponding namespace event stream.

Contract listeners have an option of where to listen from where the choices are accepted:
- `oldest` which defaults to the first block
- a specific block number 
- `latest` which will be start listen from the **creation** of the listener. 

As such when the listeners get recreated as part of the migration, the ones set to `latest` will now listen from a new latest and in the time where the chain has advance and FireFly is starting up some event could get lost! This can get worse with another bug where if FireFly cannot communicate with the Transaction Manager to create the listener it will give up and never listen to those events!

A way to solve this would be:
- Retrieve the previous contract listeners created against the default eventstream
- Find the checkpoint value - i.e the block at which this listener has caught up to!
- When creating the new contract listener against the namespace listener set the option to listen from the checkpoint! 

This was we do not lose events.

Looking at the code, this change is tricker than it sounds because the original change simply at startup creates  an event stream and apply the listeners created for the namespace. There is no notion of "migration" so this code will need to introduce some code to get the existing listeners on the "global" event stream and hopefully be able to using some identifier correlate them to the new ones that need to be created.

peterbroadhurst:
Concrete proposal - which has other benefits:

- [ ] For FFTM based connectors (EVMConnect) we do the work to ensure FireFly Core and query the latest checkpoint for a listener
- [ ] We update the migration code to _ask the Go `blockchain` plugin_ if it can get the checkpoint for any migration of listeners, and do one of these - in order:
    - [ ] Use the checkpoint if it can be obtained (EVMConnect only - avoids replaying the world!)
    - [ ] Look up the most recent event dispatched by the connector, and if one is found then require the `blockchain` plugin to decode the `protocolId` back into a `fromBlock` (all connectors - EthConnect/Fabric)
    - [ ] If no event, then fall back to the original approach

peterbroadhurst:
The listener endpoint on FFTM already has the checkpoint information embedded, so it doesn't look like changes would be needed to the FFTM or EVMConnect codebases for Core to be able to query the checkpoint information:

[...]

PR #1534, opened 2024-06-24
When re-creating listeners in EVMConnect/EthConnect/FabConnect use the last event block

peterbroadhurst:
See #1531 

Every time Core starts, it confirms all the listeners exist that it needs for all of:
- Contract listeners registered by applications
- `BatchPin` listeners for any multi-party namespaces
- Token listeners - indirectly via `/activatepool`
    - Not changed in this PR - see #1531 comments for justification

This is needed for two cases:
1. In case the listeners were deleted operationally for any reason (intentionally or otherwise) - such as a DR scenario
2. In the v1.2->v1.3 migration, where we intentionally orphan all the core-wide listeners in favor of namespace-specific ones

In both of these cases, the logic was naive for the `newest`/`latest` case - where it treats "now" as the latest event.

That is a problem, as it might be a long time since the original listener was created, and it might have delivered many events already. In which case there might be "gaps" we skip over between the last event detected by the old listener, and the first event we detect on the new one.

This PR closes that window, by looking at the `protocolId` of the last event - which is architecturally required to be a lexicographically sorted string identifying the position of the event on the blockchain.

If there's been a previous event, that's used by the blockchain connector as the `fromBlock` for the new listener.

Also, if `"oldest"` or an explicit block number is used, and the block number is before the block determined by the above process, then the newer block is used.


internal/multiparty/manager.go
```diff
@@ -155,7 +155,23 @@ func (mm *multipartyManager) configureContractCommon(ctx context.Context, migrat
...
 	}
 
-	subID, err := mm.blockchain.AddFireflySubscription(ctx, mm.namespace, current)
+	// For the case that we're establishing a listener from "latest" we obtain the protocol ID
+	// of the latest event confirmed from the blockchain for a given subscription.
+	// This protocolID should be parsed and used by the blockchain plugin if SubOptsFirstEventNewest
+	// is passed through, and the listener does not exist.
+	fb := database.BlockchainEventQueryFactory.NewFilter(ctx).Sort("-protocolid").Limit(1)
+	latestEvents, _, err := mm.database.GetBlockchainEvents(ctx, mm.namespace.Name, fb.Eq(
+		"listener", nil,
```
EnriqueL8, line 164: Why are we not using the multiparty listener ID here? We don't have access to it?

peterbroadhurst: Specifically the thing that distinguishes the `blockchainevents` for the multi-party `BatchPin` events, is the lack of a `listener` reference (there isn't one in the FF model for the `BatchPin` case).

peterbroadhurst:
PR has been updated based on the conversation here, so that in the case a specific block number is supplied the recreated listener will still restart on the block before the last event (if that's after the specified block): https://github.com/hyperledger/firefly/issues/1531#issuecomment-2187054888

EnriqueL8:
Will be testing this one today e2e

internal/multiparty/manager.go
EnriqueL8: Ah makes sense

EnriqueL8 (commented):
Re-reviewing I thought about a crazy edge case... The protocol ID is block number/transaction number /event number so if we recreate a listener with block number of the last protocol ID but we haven't consumed all the events from that block could we lose those events? Is it safer to go one down of the procotol id or am I over complicating this? Would have to see how FireFly acknowledges events from FFTM and maybe the assumption you have made is that you get all the events per block in the webscoket?

EnriqueL8:
Have tested this one E2E by doing the following experiment:
- Setup a 1.3.0 FF stack
- Deploy a contract and a listener with the option to listen from newest
- Delete the listener from EVMConnect
- Restart the FF with the code from this PR
- Find the new listener created in EVMConnect and make sure the `fromBlock` matches the last protocol ID for the last event - 1

EnriqueL8 (approved):
This looks good to me!

peterbroadhurst:
> we recreate a listener with block number of the last protocol ID but we haven't consumed all the events from that block could we lose those events? Is it safer to go one down of the procotol id or am I over complicating this?

This is the reason we go one block back:
https://github.com/hyperledger/firefly/blob/8dd5a0d2899a0b002d080a60f15ab64770ef1be0/internal/blockchain/ethereum/eventstream.go#L216-L218

EnriqueL8:
Yeah I realised with https://github.com/hyperledger/firefly/pull/1534#issuecomment-2205567325

PR #1536, opened 2024-07-08
fix: set default namespace for apis from config

### LINKED ISSUE #1535: Cannot access API without namespace query parameter
Whilst running FF locally, I created a new API using the root level `/apis` call and got a URL for the Swagger UI at `http://127.0.0.1:5001/api/v1/apis/multi-simple-storage/api` but it cannot be found.

Instead if I go to `http://127.0.0.1:5001/api/v1/namespaces/default/apis/multi-simple-storage/api/` then I can access it. So it seems that there has been a regression for the default namespace..

EnriqueL8:
Fixes #1535

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -61,6 +61,7 @@ type apiServer struct {
 	ffiSwaggerGen          FFISwaggerGen
 	apiPublicURL           string
 	dynamicPublicURLHeader string
+	defaultNamespace       string
 }
 
 func InitConfig() {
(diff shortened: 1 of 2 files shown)


peterbroadhurst (changes requested):
We've made a decision for a long time now, that we never make calls to the config on the critical path runtime.
I think `getBaseURL` is called beyond construction of the route at runtime.

Thus the default namespace needs to be locked in during the initialization of the component, and stored into a variable. Like `dynamicPublicURLHeader` above

EnriqueL8:
aah okay makes sense - let me do that at initialisation time

EnriqueL8:
@peterbroadhurst ready for re-review

peterbroadhurst (approved):
Thanks for the update 👍

PR #1544, opened 2024-07-22
docs: ahead of v1.3.1

EnriqueL8:
Update release notes for the v1.3.1 release.

v1.3.1 included mostly fixes and the main feature to the FireFly Core is the multiple filter per contract listeners


peterbroadhurst (changes requested):
I do think it's important we fully document this feature, and the reasons you might want to make changes in the application (even though you are not forced to)

awrichar:
I've proposed some of my own edits here: https://github.com/hyperledger/firefly/pull/1550

awrichar (approved):

doc-site/docs/reference/types/_includes/contractlistener_description.md
```diff
@@ -4,3 +4,100 @@ of the interface for that event.
...
+From v1.3.1 onwards, a contract listener can be created with multiple filters under a single topic when supported by the connector.
+
+Before this change, each contract listener would only support listening to one specific event from an interface previously defined. Each listener would be comprised of:
+
+- a reference to a specific blockchain event to listen for
+- (optional) a specific location/address to listen from
+- a connector-specific signature (generated from the event), which allows you to easily identify and search for the contact listener for an event
+- a `topic` which determines the ordered stream that these events are part of
+
+This format is still supported by the API. However, it may not fully guarantee accurate ordering of events coming from multiple listeners, even if they share the same `topic` (such as during cases of blockchain catch-up, or when one listener is created much later than the others).
```
peterbroadhurst, line 21:
```suggestion
For backwards compatibility the `event` and `signature` fields are still supported by the API. They contain the `event` from the first `filters` entry.
```

I _think_ this is the point being made here? Not sure. Overall this paragraph seems a little confusing.

peterbroadhurst: No, you've made that point below.

I think this paragraph just needs removing

peterbroadhurst: > it may not fully guarantee accurate ordering of events coming from multiple listeners

This statement has nothing to do with the API input/output format. I think you've elsewhere made the point how the new flexibility of allowing `len(filter) > 1` is a good thing.

doc-site/docs/reference/types/_includes/contractlistener_description.md
```diff
@@ -4,3 +4,100 @@ of the interface for that event.
...
+
+From v1.3.1, you can supply multiple filters on a single listener. Each filter contains:
+
+- a reference to a specific blockchain event to listen for
+- (optional) a specific location/address to listen from
+- a connector-specific signature (generated from the event and the location)
+
+In addition to this list of multiple filters, the listener specifies a single `topic` to identify the stream of events. This new feature will allow better management of contract listeners and strong ordering of all of the events your application cares about.
+
+Note: For backwards compatibility, the response from the API will populate top-level `event` and `location` fields with the contents of the first event filter in the array.
```
peterbroadhurst, line 31:
```suggestion
### Backwards compatibility

- The response from all query APIs of `listeners` will populate top-level `event` and `location` fields
    - The first entry from the `filters` array is duplicated to these fields
- On input to create a new `listener` the `event` and `location` fields are supported
    - They function identically to supplying a `filters` array with a single entry
- The `signature` field is preserved at the top level
   - The format has been changed as described above
   - When multiple filters exist the signature contains multiple signature strings, concatenated with `;`
```

Validated from https://github.com/hyperledger/firefly/blob/44ab1c72ffa5e47aa06746492b6bd4e72db7288b/internal/contracts/manager.go#L960

As an aside, I came across a confusingly named function on my travels (which has nothing to do with functions, which don't have an `indexed` `bool`):
https://github.com/hyperledger/firefly/blob/44ab1c72ffa5e47aa06746492b6bd4e72db7288b/internal/blockchain/ethereum/ethereum.go#L1021C6-L1021C33

awrichar: Yes, seems like `ABIMethodToIndexedSignature` does double duty by helping to parse both methods and events...

peterbroadhurst (approved):

PR #1556, opened 2024-08-13
fix: doc site release action

EnriqueL8:
This should fix the release GitHub action that did not set the version as latest for the v1.3.1 release. Unfortunately this only gets triggered on release, so a manual step will be needed to fix the version this time around


awrichar (approved):

EnriqueL8:
DCO check stuck - closing and reopening to see if that works

PR #1558, opened 2024-08-14
chore: Remove getIdentityByDID route

### LINKED ISSUE #1528: Swagger issue when working with identities
Attempting to get the details of an identity through a GET call, or update an identity through a PATCH call via the embedded Swagger UI results in an error. It appears something in the swagger isn't generated quite right. Specifically, the path argument that corresponds to the identity id is not passed and instead a constant string is included.

dwertent:
# Overview
* Fixed issues with rendering Swagger documentation.
* Addressed some linter warnings.

## Swagger Documentation Issue
The problem occurred when rendering the Swagger documentation. The code is designed to list API paths and generate [documentation based on those paths](https://github.com/hyperledger/firefly-common/blob/c214085f1d3e3aeedf23a435ce218ec10eb8b6ef/pkg/ffapi/openapi3.go#L105). However, an issue arose in the way the `Find` [function was used](https://github.com/getkin/kin-openapi/blob/master/openapi3/paths.go#L180). This function is [agnostic to the specific value of the API keys](https://github.com/getkin/kin-openapi/blob/master/openapi3/paths.go#L170), which led to the swagger generator pulling in the wrong documentation.

## Solution
I considered the following potential solutions:
1. Implement a custom `Find` function that is aware of the API key value.
2. Combine the API functions so that `identities/{ID}` can accept either a UUID or a DID.

I chose the second solution for the following reasons:
1. A different organization maintains the `Find` function, and it may take time to implement a custom version.
2. Since the API paths for handling DID and UUID are the same, it's logical to use a single function to handle both types of requests.

## Side Note
I recommend deprecating DID support in favor of using UUIDs exclusively. This would simplify our API by providing a single, consistent definition for ID, reducing potential confusion and complexity in future development.

This PR fixes #1528


internal/apiserver/route_get_identity_by_id.go
```diff
@@ -41,10 +43,30 @@ var getIdentityByID = &ffapi.Route{
...
+// isUUID checks if the given string is a UUID
+func isUUID(s string) bool {
+	_, err := uuid.Parse(s)
+	return err == nil
+}
+
+// isDID checks if the given string is a potential DID
+func isDID(s string) bool {
+	return strings.HasPrefix(s, "did:")
+}
```
EnriqueL8, line 72: Keep the same level of check and change this to a regex expression - I think we should have one for DIDs already in the codebase

internal/apiserver/route_get_identity_by_id.go
```diff
@@ -20,17 +20,19 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/google/uuid"
```
EnriqueL8, line 23: We don't use the google UUID library directly, use the one in firefly-common

internal/coremsgs/en_api_translations.go
```diff
@@ -107,7 +108,7 @@ var (
 	APIEndpointsGetGroupByHash                  = ffm("api.endpoints.getGroupByHash", "Gets a group by its ID (hash)")
 	APIEndpointsGetGroups                       = ffm("api.endpoints.getGroups", "Gets a list of groups")
 	APIEndpointsGetIdentities                   = ffm("api.endpoints.getIdentities", "Gets a list of all identities that have been registered in the namespace")
-	APIEndpointsGetIdentityByID                 = ffm("api.endpoints.getIdentityByID", "Gets an identity by its ID")
+	APIEndpointsGetIdentityByID                 = ffm("api.endpoints.getIdentityByID", "Gets an identity by its ID (UUID/DID)")
```
EnriqueL8, line 111:
```suggestion
	APIEndpointsGetIdentityByID                 = ffm("api.endpoints.getIdentityByID", "Gets an identity by its ID or DID")
```

internal/syncasync/sync_async_bridge.go
```diff
@@ -155,7 +155,7 @@ func (sa *syncAsyncBridge) getInFlight(ns string, reqType requestType, id *fftyp
 		return nil
 	}
 	inflightNS := sa.inflight[ns]
-	if inflightNS != nil && id != nil {
```
EnriqueL8, line 158: checked that this changed is fine as above we return nil if id is empty

internal/apiserver/route_get_identity_by_id.go
```diff
@@ -20,17 +20,19 @@ import (
...
 	"github.com/hyperledger/firefly-common/pkg/ffapi"
+	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly/internal/coremsgs"
 	"github.com/hyperledger/firefly/pkg/core"
 )
 
 var getIdentityByID = &ffapi.Route{
 	Name:   "getIdentityByID",
-	Path:   "identities/{iid}",
+	Path:   "identities/{id:.+}",
```
EnriqueL8, line 32: I'm not sure you need the `.+` if you are verifying the path parameter underneath

internal/apiserver/route_get_identity_by_id.go
```diff
@@ -20,17 +20,19 @@ import (
...
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/hyperledger/firefly-common/pkg/ffapi"
+	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly/internal/coremsgs"
 	"github.com/hyperledger/firefly/pkg/core"
 )
 
 var getIdentityByID = &ffapi.Route{
```
EnriqueL8, line 30:
```suggestion
var getIdentityByIDOrDID = &ffapi.Route{
```

EnriqueL8 (commented):
Based on your description the kin-openapi tool is not smart enough to differentiate between the same API paths but with different path parameters so the second one in the order will override the first one. In this case `/identities/{did}` overrides `/identities/{iid}` which results in the issue. You have decided to combine both those API calls into one generic `/identities/{id}`.

This introduces a slightly **breaking** change as the name of the path parameter is now different so the swagger will be different but the requests should still work. The description now is generic for both ID and DID. 

The side note is not valid as DIDs are different to the IDs used internal by FireFly. A Decentralized identity (DID) is used to enable decentralised verification and is pinned on-chain, associated with a key. The IDs used here are purely a way for FireFly to uniquely identify in the DB the identities that this FireFly cares about and it's only present in the context of one single FireFly instance.

internal/apiserver/server.go
```diff
@@ -372,16 +372,16 @@ func (as *apiServer) createMuxRouter(ctx context.Context, mgr namespace.Manager)
...
+	as.namespacedSwaggerHandler(hf, r, as.apiPublicURL, `/api/swagger.json`)
+	as.namespacedSwaggerHandler(hf, r, as.apiPublicURL, `/api/openapi.json`)
+	as.namespacedSwaggerHandler(hf, r, as.apiPublicURL, `/api/swagger.yaml`)
+	as.namespacedSwaggerHandler(hf, r, as.apiPublicURL, `/api/openapi.yaml`)
 	as.namespacedSwaggerUI(hf, r, as.apiPublicURL, `/api`)
 	// Dynamic swagger for namespaced contract APIs
-	as.namespacedContractSwaggerGenerator(hf, r, mgr, as.apiPublicURL, `/api/swagger.json`, ffapi.OpenAPIFormatJSON)
-	as.namespacedContractSwaggerGenerator(hf, r, mgr, as.apiPublicURL, `/api/openapi.json`, ffapi.OpenAPIFormatJSON)
-	as.namespacedContractSwaggerGenerator(hf, r, mgr, as.apiPublicURL, `/api/swagger.yaml`, ffapi.OpenAPIFormatYAML)
-	as.namespacedContractSwaggerGenerator(hf, r, mgr, as.apiPublicURL, `/api/openapi.yaml`, ffapi.OpenAPIFormatYAML)
```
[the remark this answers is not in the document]
EnriqueL8: I think it's just the linter is saying that the arguments are not being used in the function - ideally we fix the bug there and make it work properly, maybe not as part of this PR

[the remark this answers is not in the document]
dwertent: I will revert my changes and open an issue so we can address this later

internal/apiserver/route_get_identity_by_id.go
dwertent: Since we have the DID validation happening later in the code, I preferred to determine if this a DID string or not. Because if this is a DID but a not valid one, I want the user to receive the proper error code (vs in this case the error code will be "unknown format"). wdyt?

internal/apiserver/route_get_identity_by_id.go
dwertent: Same here, we are not validating the input, rather we are just filtering the input, so I didn't think we need to call the UUID validator with the context.

internal/apiserver/route_get_identity_by_id.go
dwertent: This helps the API interpret the `/` character in the DID string properly, without this the API will return a 404 error code. I will add this comment to the code.

internal/apiserver/route_get_identity_by_id.go
EnriqueL8: just more of a consistency and dependency thing to use the firefly-common library for this

internal/apiserver/route_get_identity_by_id.go
EnriqueL8: aaah yeah makes sense

EnriqueL8:
@dwertent build failing and e2e tests as well

EnriqueL8:
something wrong with the UUID verification `restclient.go:118: <!! {"error":"FF00126: Unknown identity type: %!!(MISSING)s(MISSING)"}`

internal/apiserver/route_get_identity_by_id.go
EnriqueL8: ah because you are not catching the error in the switch statement, you could catch the error there if you wish to? It just feels since you have moved the validation here make sense to construct the correct error instead of relying on the function underneath to throw the right thing - but will leave up to you

dwertent:
This PR is no longer needed as we've resolved the issue with a [different approach](https://github.com/hyperledger/firefly-common/pull/145). I will open a new PR to update the FireFly-common tag once it's published.

PR #1559, opened 2024-08-14
Refactor `maybePersistBlockchainEvent` to Return a Boolean Indicating Event Creation Status

### LINKED ISSUE #1294: Blockchain event metrics can be double-counted
Calls to `emitBlockchainEventMetric()` typically follow a call to `maybePersistBlockchainEvent()`, but since `maybePersistBlockchainEvent()` doesn't return anything to indicate if the event already existed, FF will emit for both new and existing events.

dwertent:
# Overview

This PR refactors the `maybePersistBlockchainEvent()` function to return a boolean indicating whether a blockchain event was newly created. This change prevents the double-counting of metrics by ensuring that `emitBlockchainEventMetric()` is only called for new events. 

Fixes #1294.

### DIFF
--- a/internal/events/blockchain_event.go
+++ b/internal/events/blockchain_event.go
@@ -71,20 +71,21 @@ func (em *eventManager) getChainListenerByProtocolIDCached(ctx context.Context, 
 	return l, nil
 }
 
-func (em *eventManager) maybePersistBlockchainEvent(ctx context.Context, chainEvent *core.BlockchainEvent, listener *core.ContractListener) error {
+// handleBlockchainBatchPinEvent handles a blockchain event, returning true if the event was created, false if it was a duplicate
+func (em *eventManager) maybePersistBlockchainEvent(ctx context.Context, chainEvent *core.BlockchainEvent, listener *core.ContractListener) (bool, error) {
 	existing, err := em.txHelper.InsertOrGetBlockchainEvent(ctx, chainEvent)
 	if err != nil {
-		return err
+		return false, err
 	}
 	if existing != nil {
 		log.L(ctx).Debugf("Ignoring duplicate blockchain event %s", chainEvent.ProtocolID)
 		// Return the ID of the existing event
 		chainEvent.ID = existing.ID
-		return nil
+		return false, nil
 	}
 	topic := em.getTopicForChainListener(listener)
 	ffEvent := core.NewEvent(core.EventTypeBlockchainEventReceived, chainEvent.Namespace, chainEvent.ID, chainEvent.TX.ID, topic)
-	return em.database.InsertEvent(ctx, ffEvent)
+	return true, em.database.InsertEvent(ctx, ffEvent)
 }
 
 func (em *eventManager) getChainListenerCached(cacheKey string, getter func() (*core.ContractListener, error)) (*core.ContractListener, error) {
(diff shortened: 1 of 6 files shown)


internal/events/blockchain_event_test.go
```diff
@@ -152,31 +152,53 @@ func TestContractEventWrongNS(t *testing.T) {
...
+				},
+				Listener: fftypes.NewUUID(),
+			},
+			existingID:      fftypes.NewUUID(),
+			expectedID:      fftypes.NewUUID(),
+			expectedError:   nil,
+			expectedCreated: false,
 		},
-		Listener: fftypes.NewUUID(),
+		// TODO: Add test case for event not existing
```
EnriqueL8, line 185: This TODO probably still valid

internal/events/blockchain_event.go
```diff
@@ -71,20 +71,21 @@ func (em *eventManager) getChainListenerByProtocolIDCached(ctx context.Context,
 	return l, nil
 }
 
-func (em *eventManager) maybePersistBlockchainEvent(ctx context.Context, chainEvent *core.BlockchainEvent, listener *core.ContractListener) error {
+// handleBlockchainBatchPinEvent handles a blockchain event, returning true if the event was created, false if it was a duplicate
```
EnriqueL8, line 74:
```suggestion
// handleBlockchainBatchPinEvent handles a blockchain event, returning true if the event was created, false if it was a duplicate along with an error if any failures occur
```

internal/events/blockchain_event_test.go
```diff
@@ -152,31 +152,53 @@ func TestContractEventWrongNS(t *testing.T) {
...
-		},
-		Info: fftypes.JSONObject{
-			"blockNumber": "10",
+	testCases := []testCase{
+		{
+			name: "Event already exists",
+			event: &core.BlockchainEvent{
+				ID:         fftypes.NewUUID(),
+				Name:       "Changed",
+				Namespace:  "ns1",
```
EnriqueL8, line 170: I see you have gone for a table test with multiple permutations of the same tests, we haven't really explored these in this codebase. At some point it would be nice to reduce code to have table tests but for now if you could keep to the same format and just have one func per test. 

I think because of the amount of mocking the table tests can be useful but sometimes difficult to manage as well

EnriqueL8 (commented):
Looks good - a few comments

EnriqueL8:
@dwertent just build failing because of the copyright date, we can probably write a small command to update the ones modified

internal/events/blockchain_event_test.go
```diff
@@ -151,6 +151,7 @@ func TestContractEventWrongNS(t *testing.T) {
 
 }
 
+// TODO: Add test case for event not existing
```
EnriqueL8, line 154: Are you going to address this todo?

dwertent: I believe this requires some more work with setting up mocks, etc. I can work on it is a separate PR.

EnriqueL8:
Re-ran the build to upload the code coverage as the repo was deactivated in CodeCov for some reason...

internal/events/blockchain_event_test.go
EnriqueL8: Code Coverage will force you to implement this test as you have added new code to return not created

EnriqueL8:
@dwertent more copyright failures

dwertent:
@EnriqueL8 I think we said we would merge this one.

EnriqueL8 (approved):
Yeah, good to go!

PR #1561, opened 2024-08-19
chore: Fix Cache Mismatch by Deep Copying Operation

### LINKED ISSUE #1310: Retried operations pollute the cache with incorrect values
Given operation A, which failed.
I retried operation A, which created operation B.
Now, when performing a "get one" to get operation A, it actually returns the details of operation B.
When performing a "get many", the details of operation A can be retrieved correctly.

Somehow it seems like the cache became polluted (since the "get one" API goes to the cache). I haven't pinpointed the exact path, but it may be somewhere around this area:
https://github.com/hyperledger/firefly/blob/4df24f30508b046fe57def1d386d2fb73a5d1979/internal/operations/manager.go#L214

dwertent:
Added a deep copy of the operation object to prevent ID overwriting issues. Read more [here](https://github.com/hyperledger/firefly/issues/1310#issuecomment-2297198669).

Fixes #1310

### DIFF
--- a/pkg/core/operation.go
+++ b/pkg/core/operation.go
@@ -1,4 +1,4 @@
-// Copyright © 2023 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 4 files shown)


EnriqueL8:
@awrichar Could you take a look at this one?

pkg/core/operation.go
```diff
@@ -74,6 +74,23 @@ func (op *Operation) IsTokenOperation() bool {
...
+		Status:      op.Status,
+		Plugin:      op.Plugin,
+		Input:       op.Input,
+		Output:      op.Output,
+		Error:       op.Error,
+		Created:     op.Created,
+		Updated:     op.Updated,
+		Retry:       op.Retry,
+	}
+}
```
EnriqueL8, line 92: This DeepCopy needs to handle the fact that some of the fields are pointers so that they are deep copied correctly https://github.com/hyperledger/firefly/blob/74e2cbd38e688d0943618f149068880150a67020/pkg/core/operation.go#L114-L125 so needs a more sophisticated check for `nil` fields as well

EnriqueL8: It works in your test because UUID is actually special in the case where it has no function to modify the value itself and in the test you are doing a complete replace

pkg/core/operation_test.go
```diff
@@ -97,6 +98,45 @@ func TestOperationTypes(t *testing.T) {
...
+	assert.Equal(t, op.Updated, copyOp.Updated)
+	assert.Equal(t, op.Retry, copyOp.Retry)
+
+	// Ensure the copy is a deep copy
+	op.ID = fftypes.NewUUID()
+	assert.NotEqual(t, copyOp.ID, op.ID)
+
+	// Ensure no new fields are added to the Operation struct
+	// If a new field is added, this test will fail and the DeepCopy function should be updated
+	assert.Equal(t, 12, reflect.TypeOf(Operation{}).NumField())
```
EnriqueL8, line 138: this is neat!

pkg/core/operation_test.go
```diff
@@ -97,6 +98,45 @@ func TestOperationTypes(t *testing.T) {
...
+	assert.Equal(t, op.Input, copyOp.Input)
+	assert.Equal(t, op.Output, copyOp.Output)
+	assert.Equal(t, op.Error, copyOp.Error)
+	assert.Equal(t, op.Created, copyOp.Created)
+	assert.Equal(t, op.Updated, copyOp.Updated)
+	assert.Equal(t, op.Retry, copyOp.Retry)
+
+	// Ensure the copy is a deep copy
+	op.ID = fftypes.NewUUID()
+	assert.NotEqual(t, copyOp.ID, op.ID)
```
EnriqueL8, line 134: can you add a test case with `fftypes.FFTime`

EnriqueL8, line 134: This is not really proving what the function should do. You should really not reassign the field but instead compare the address of the returning fields `assert.NotEqual(t, &op.ID, &copyOp.ID)`

Your test is working because you copying the whole struct but not modifying the data of a field in place, because the `fftypes.UUID` and `fftypes.FFTime` do not have a mutating function we are "covered"

EnriqueL8 (approved):
LGTM - just needs that coverage bumped

EnriqueL8 (changes requested):
just waiting on code coverage to be 100%

EnriqueL8 (approved):
LGTM

PR #1564, opened 2024-08-27
Fixing protocol ID docs

### LINKED ISSUE #1455: Protocol ID behaviour inconsistent with docs
[docs](https://hyperledger.github.io/firefly/reference/types/blockchainevent.html#protocol-id) say
>000003 - this is the event (/log) index within that transaction

However, what I observe is that the value in this segment increases across transactions and only resets to 0 on a new block.  So itsems to be the event /log index within a block.

e.g. when I have 1 event per transaction, 2nd segment and 3rd segment are always equal.  
```
"000000112840/000006/000006"
"000000112840/000005/000005"
"000000112840/000004/000004"
"000000112840/000003/000003"
"000000112840/000002/000002"
"000000112840/000001/000001"
"000000112840/000000/000000"
```

If I have 2 events per transaction, then I see segment 3 increase sequentially on every event and segment 2 increases by one on every other event
```
"000000147029/000004/000009"
"000000147029/000004/000008"
"000000147029/000003/000007"
"000000147029/000003/000006"
"000000147029/000002/000005"
"000000147029/000002/000004"
"000000147029/000001/000003"
"000000147029/000001/000002"
"000000147029/000000/000001"
"000000147029/000000/000000"
```

Curious if the docs are wrong or whether this is a code bug.

dwertent:
This PR corrects the documentation to reflect that the third segment of the `protocolId` represents the event (/log) index within the entire block, which aligns with the observed behavior and Ethereum's standards.
I referenced this [comment](https://github.com/hyperledger/firefly/issues/1455#issuecomment-1910275127) for the fix.

Resolves: #1455

### DIFF
--- a/doc-site/docs/reference/types/_includes/blockchainevent_description.md
+++ b/doc-site/docs/reference/types/_includes/blockchainevent_description.md
@@ -25,8 +25,8 @@ the blockchain plugins to try and create some consistency.
 An example `protocolId` string is: `000000000041/000020/000003`
 
 - `000000000041` - this is the block number
-- `000020` - this is the transaction index within that block
-- `000003` - this is the event (/log) index within that transaction
+- `000020` - this is the **transaction** index within that **block**
+- `000003` - this is the **event (/log)** index within that **block**
 
 The string is alphanumerically sortable as a plain string;
 


EnriqueL8 (approved):
Thanks - looks good to me!

PR #1565, opened 2024-08-28
chore: Ignore markdown files and GitHub workflows in Docker and Go wo…

dwertent:
This will prevent unnecessary tests from running when changes are made to the markdown files.
Here is an example: https://github.com/hyperledger/firefly/actions/runs/10594695480

### DIFF
--- a/.github/workflows/docker_main.yml
+++ b/.github/workflows/docker_main.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - '.github/**' # exclude .github directory
+      - '**.md' # exclude all markdown files
 
 jobs:
   docker:
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -3,8 +3,19 @@ name: Go
 on:
   push:
     branches: [main]
+    paths:
+      - '**'  # include all files
+      - '!.github/**' # exclude .github directory
+      - '!**.md' # exclude all markdown files
+      - 'doc-site/docs/reference/**.md' # include markdown files that are auto generated and need to be tested
+
   pull_request:
-    branches: [main]
+    paths:
+      - '**'  # include all files
+      - '!.github/**' # exclude .github directory
+      - '!**.md' # exclude all markdown files
+      - 'doc-site/docs/reference/**.md' # include markdown files that are auto generated and need to be tested
+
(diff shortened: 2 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -3,8 +3,11 @@ name: Go
 on:
   push:
     branches: [main]
+    paths-ignore:
+      - '**.md'
   pull_request:
-    branches: [main]
```
dwertent, line 7: I don't see a reason to limit the tests to only run when a PR is opened against the main branch. Removing this restriction will enable the development of feature branches and opening PRs against them.

.github/workflows/docker_main.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - '**.md'
+      - '.github/**'
```
EnriqueL8, line 9: We need the tests to run because we generate the references for the structs through the test

EnriqueL8 (changes requested):
We need the test to run for these https://github.com/hyperledger/firefly/blob/a98387745f6e1e940bd6f9ccf8ac56b9a9b7914c/internal/reference/reference_test.go#L36 to validate that the Markdown generated is correct based on the code annotations

.github/workflows/docker_main.yml
dwertent: The tests currently run whenever a `.go` file changes, as the tests generate the `.md` files and are triggered by `.go` file modifications. In this PR, I propose not running Go tests when only `.md` files are changed since these changes don't impact the Go code.
To summarize, the test `TestCheckGeneratedMarkdownPages` will run when `.go` code changes as expected.

EnriqueL8: Yeah that makes sense but the markdown generated could be incorrect compared to the code that is checked in. So if I go and manually change for some reason the markdown files generated through the tests then the test that checks that the MD generated matches the values in code won't run and the PR checks will be green. Giving the reviewer a false indication that the MD is correct

dwertent: So I will ignore specific paths

EnriqueL8 (commented):
Didn't finish my previous review that we discussed

EnriqueL8 (approved):
looks good

PR #1566, opened 2024-08-28
Tag image as latest only when releasing the latest tag

### LINKED ISSUE #1506: Fix `latest` tag in Docker image workflows
Right now our docker image GitHub workflows have a problem where if we release a 1.2.x patch release after a 1.3.0, the 1.2.x release will get tagged as `latest` which will potentially "downgrade" people if they try to pull latest. I think most people would expect 1.3.0 to remain the latest in this case.

I wrote a new way of determining whether a workflow is running as the result of being the latest release, or just some other release and it's part of the docs site build here: https://github.com/hyperledger/firefly/blob/main/.github/workflows/docs.yml#L26-L48

We should port this code into the docker release job as well: https://github.com/hyperledger/firefly/blob/main/.github/workflows/docker_release.yml

This definitely needs to happen before we do a 1.2.3 patch release or we will need to manually fix Docker tags as we have in the past.

dwertent:
Resolves: #1506

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -5,8 +5,28 @@ on:
     types: [released, prereleased]
 
 jobs:
+  fetch-latest-tag:
+    runs-on: ubuntu-latest
+    outputs:
+      latest_tag: ${{ steps.get_latest_tag.outputs.latest_tag }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


EnriqueL8 (commented):
looks good - but e2e is failing

dwertent:
@EnriqueL8 tests passed :)

EnriqueL8 (approved):
LGTM!

PR #1570, opened 2024-09-06
fix: get latest tag

dwertent:
* Fixing [this error](https://github.com/hyperledger/firefly/actions/runs/10740543298#:~:text=by%20humans.txt-,Annotations,-1%20error)
* Using ENV instead of the deprecated `set-output`

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -5,36 +5,25 @@ on:
     types: [released, prereleased]
 
 jobs:
-  fetch-latest-tag:
-    runs-on: ubuntu-latest
-    outputs:
-      latest_tag: ${{ steps.get_latest_tag.outputs.latest_tag }}
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-
-      - name: Fetch all tags
-        run: git fetch --tags
-
-      - name: Get the latest tag
-        id: get_latest_tag
-        run: |
-          latest_tag=$(git tag -l | sort -V | tail -n 1)
-          echo "::set-output name=latest_tag::${latest_tag}"
-
+   
   docker:
     runs-on: ubuntu-latest
-    needs: fetch-latest-tag
     permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -23,32 +23,24 @@ jobs:
...
 
       - name: Update doc site for release
-        if: ${{ github.event_name == 'release' && env.IS_LATEST_RELEASE != 'true' }}
+        if: github.event.action == 'released' && github.ref_name == env.LATEST_TAG
         working-directory: doc-site
         run: mike deploy ${{ github.event.release.tag_name }} --push
 
       - name: Update doc site for latest release
-        if: ${{ github.event_name == 'release' && env.IS_LATEST_RELEASE == 'true' }}
+        if: github.event.action == 'released' && github.ref_name == env.LATEST_TAG
```
EnriqueL8, line 43: They shouldn't be the same if check.. the second one should run if it's the latest tag and the first one if not

EnriqueL8 (changes requested):

.github/workflows/docs.yml
dwertent: Oh my bad

.github/workflows/docker_release.yml
```diff
@@ -5,36 +5,25 @@ on:
     types: [released, prereleased]
 
 jobs:
-  fetch-latest-tag:
```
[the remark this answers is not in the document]
dwertent: I made this change because I was using `::set-output`, which is deprecated. Since I was already updating the code, I decided to align the implementation with `docs.yaml` for consistency. In the future, I could move both implementations to a separate file so they can call the same job, but for now, I wanted to ensure they are consistent

dwertent:
I added some minor changes to the Dockerfile based on [these warnings](https://github.com/hyperledger/firefly/actions/runs/10795366695/job/29941873807#step:4:171)

EnriqueL8 (approved):
Looks good!

PR #1579, opened 2024-09-13
Update ffcommon, ffsigner, ffevmconnect

matthew1001:
This pulls in a fix in `firefly-common` related to large number handling, and updates the other related dependencies and manifest list

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6,8 +6,8 @@
   },
   "evmconnect": {
     "image": "ghcr.io/hyperledger/firefly-evmconnect",
-    "tag": "v1.3.17",
-    "sha": "0fb12537b8be4fb463df713ed541aa03e2da42c582cc4c9e822f7c59a6962fef"
+    "tag": "v1.3.18",
+    "sha": "11a782c778227d198c0bff2fc94ad7e52ceea5a9994728ba05e781541680c32d"
   },
   "fabconnect": {
     "image": "ghcr.io/hyperledger/firefly-fabconnect",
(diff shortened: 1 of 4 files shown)


manifest.json
```diff
@@ -6,8 +6,8 @@
   },
   "evmconnect": {
     "image": "ghcr.io/hyperledger/firefly-evmconnect",
-    "tag": "v1.3.17",
-    "sha": "0fb12537b8be4fb463df713ed541aa03e2da42c582cc4c9e822f7c59a6962fef"
+    "tag": "v1.3.18",
+    "sha": "11a782c778227d198c0bff2fc94ad7e52ceea5a9994728ba05e781541680c32d"
   },
   "fabconnect": {
     "image": "ghcr.io/hyperledger/firefly-fabconnect",
```
EnriqueL8, line 13: Probably need to update common here? I bet it has the same problem with JSON Numbers

EnriqueL8 (commented):
Don't we need to update FabConnect and Tezos Connect to latest transaction manager as well?

manifest.json
matthew1001: I don't think common is in the manifest list?

matthew1001:
> Don't we need to update FabConnect and Tezos Connect to latest transaction manager as well?

Yes to tezos - I'll do that now. Fabconnect doesn't pull in `fftm` I don't believe?

EnriqueL8:
Going to update the manifest with Tezos and Fabric at another point when we release a new rc

EnriqueL8 (approved):

Issue #1585, opened 2024-09-19
Observed NPEs during contract invoke


SamMayWork:
When invoking a method on a contract, we've occasionally seen FireFly get into a state where we get an NPE from the `syncasync` bridge and then no response is returned. It's not clear at the moment what causes FireFly to get into this state, but hard restarting the FireFly process seems to get it out of this state and allow invoke operations to succeed.

We've seen this issue in 1.2.2 but suspect the same kind of problem might happen in 1.3.X as it's not currently fixed yet.

<details><summary>Stack Trace</summary>

> ```
> http2: panic serving: runtime error: invalid memory address or nil pointer dereference
> goroutine 142759 [running]:
> net/http.(*http2serverConn).runHandler.func1()
> 	/usr/local/go/src/net/http/h2_bundle.go:5903 +0x125
> panic({0xf54100, 0x18a3e20})
> 	/usr/local/go/src/runtime/panic.go:884 +0x212
> github.com/hyperledger/firefly/internal/syncasync.(*syncAsyncBridge).addInFlight(0xc00037ce00, {0xc0005161a0, 0xd}, 0xc000e996b0, 0x6)
> 	/firefly/internal/syncasync/sync_async_bridge.go:142 +0x1f1
> github.com/hyperledger/firefly/internal/syncasync.(*syncAsyncBridge).sendAndWait(0xc00037ce00, {0x12c37b8, 0xc000acce40}, {0xc0005161a0, 0xd}, 0x20?, 0xfcf9c0?, 0xc000c70680)
> 	/firefly/internal/syncasync/sync_async_bridge.go:622 +0x8a
> github.com/hyperledger/firefly/internal/syncasync.(*syncAsyncBridge).WaitForInvokeOperation(0xc00020d880?, {0x12c37b8?, 0xc000acce40?}, 0xc001889440?, 0x1?)
> 	/firefly/internal/syncasync/sync_async_bridge.go:700 +0x45
> github.com/hyperledger/firefly/internal/contracts.(*contractManager).InvokeContract(0xc00020d880, {0x12c37b8, 0xc000acce40}, 0xc001889440, 0x1)
> 	/firefly/internal/contracts/manager.go:444 +0x525
[...]

PR #1586, opened 2024-09-19
fix: Clean up manager creation to avoid NPEs during background start

SamMayWork:
ref: https://github.com/hyperledger/firefly/issues/1585#issuecomment-2361706076

Without this fix when calling to create a contract manager, it's possible for both the return value and the error value to be non-nil. This ensures that strictly one or the other is set.
https://github.com/hyperledger/firefly/blob/dc97c732e4b07c32b7c60978b79bb0dab98e7460/internal/orchestrator/orchestrator.go#L540

Need to test using the steps provided above to ensure the fix is good (also pending UT coverage).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,7 +75,7 @@ ARG UI_TAG
 ARG UI_RELEASE
 RUN apk add --update --no-cache \
     sqlite=3.44.2-r0 \
-    postgresql16-client=16.3-r0 \
+    postgresql16-client=16.4-r0 \
     curl=8.9.1-r0 \
     jq=1.7.1-r0
 WORKDIR /firefly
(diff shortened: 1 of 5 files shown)


EnriqueL8:
Gonna have to fix the Docker image 
```
1.121 fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64/APKINDEX.tar.gz
1.444 ERROR: unable to select packages:
1.446   postgresql16-client-16.4-r0:
1.446     breaks: world[postgresql16-client=16.3-r0]
```

SamMayWork:
Fix tested manually using the recreate from https://github.com/hyperledger/firefly/issues/1585#issuecomment-2363830256 and all looks good.

Dockerfile
```diff
@@ -75,7 +75,7 @@ ARG UI_TAG
 ARG UI_RELEASE
 RUN apk add --update --no-cache \
     sqlite=3.44.2-r0 \
-    postgresql16-client=16.3-r0 \
+    postgresql16-client=16.4-r0 \
```
SamMayWork, line 78: Required to get the build working, looks like the old version of the client has been removed. A quick look through the PostgreSQL release notes for this version suggests it just contains bug fixes: https://www.postgresql.org/docs/release/16.4/ 

The extent of testing done on the new client version is the UTs in this repo, if there's any further testing required let me know.

peterbroadhurst (approved):
Thanks for the great investigation and fix 👍

PR #1587, opened 2024-10-03
Update manifest and release date

matthew1001:
Getting ready for the `1.3.2` release, this PR updates the manifest to use the `1.3.2` CLI and clarifies in the docs that the large number support is currently only for FF core and the EVM stack


EnriqueL8 (approved):
Awesome, looks good. Will wait for build and we can merge

PR #1588, opened 2024-10-03
Ignore release candidate tags when determining latest tag for docs

EnriqueL8:
The latests docs release picked up the wrong tag because it took into consideration release candidate versions... https://github.com/hyperledger/firefly/actions/runs/11163270495/job/31029806304 

`grep -v` removes lines with `rc` in them


peterbroadhurst (approved):
Thanks for the clear write ups on these 👍

Issue #1594, opened 2024-10-30
Batch dispatcher stuck in infinite retry loop when encountering conflict error


shorsher:
I've noticed that if there's an error (timeout/disconnect/connection loss) during a `BatchPin` submission to the blockchain connector (EVMConnect in my case), it's possible for EVMConnect to successfully process the transaction but the FireFly operation will be `Failed`. Then, the batch processor resubmits the batch but EVMConnect returns a 409, which causes the processor to indefinitely retry and prevent new batches from occurring.

Example error message:
```
FF10458: Conflict from blockchain connector: FF21065: ID 'default:f5296ba7-1b23-4c7f-8612-620dafc0e40a' is not unique d=pinned_broadcast ns=default opcache=1UGYM3mn p=did:firefly:org/org_5452a6| pid=61421 role=batchmgr
```

Currently, the only way to fix this is restarting FireFly.

From my understanding, this should be handled by FireFly idempotent retry logic https://github.com/hyperledger/firefly/blob/1939b67fa0135f59eee7452dd7b73054915eea4d/internal/operations/manager.go#L178

 I _think_ the bug is that an error is still returned by `RunOperation`, causing the batch processor to retry forever https://github.com/hyperledger/firefly/blob/1939b67fa0135f59eee7452dd7b73054915eea4d/internal/batch/batch_processor.go#L622

I'll link my naive fix that unblocks the batch processor, but I'm not sure it's fully correct because the operation still gets marked as `Failed`.

EnriqueL8:
This has been fixed

PR #1595, opened 2024-10-30
break out of infinite retry for submitted ops

shorsher:
Fix for https://github.com/hyperledger/firefly/issues/1594

I've noticed that if there's an error (timeout/disconnect/connection loss) during a BatchPin submission to the blockchain connector (EVMConnect in my case), it's possible for EVMConnect to successfully process the transaction but the FireFly operation will be Failed. Then, the batch processor resubmits the batch but EVMConnect returns a 409, which causes the processor to indefinitely retry and prevent new batches from occurring.

Example error message:

`FF10458: Conflict from blockchain connector: FF21065: ID 'default:f5296ba7-1b23-4c7f-8612-620dafc0e40a' is not unique d=pinned_broadcast ns=default opcache=1UGYM3mn p=did:firefly:org/org_5452a6| pid=61421 role=batchmgr
Currently, the only way to fix this is restarting FireFly.`

Instead, I propose we break the retry loop by having the batch processor not retry on conflict errors.

### DIFF
--- a/internal/batch/batch_processor.go
+++ b/internal/batch/batch_processor.go
@@ -636,6 +636,7 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
 				conflictErr, conflictTestOk := err.(operations.ConflictError)
 				if conflictTestOk && conflictErr.IsConflictError() {
 					// We know that the connector has received our batch, so we shouldn't need to retry
+					payload.addMessageUpdate(payload.Messages, core.MessageStateReady, core.MessageStateSent)
 					return true, nil
 				}
 			} else {
--- a/internal/batch/batch_processor.go
+++ b/internal/batch/batch_processor.go
@@ -633,6 +633,11 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
 						}
 					}
 				}
+				conflictErr, conflictTestOk := err.(operations.ConflictError)
+				if conflictTestOk && conflictErr.IsConflictError() {
+					// We know that the connector has received our batch, so we shouldn't need to retry
+					return true, nil
+				}
 			} else {
 				if core.IsPinned(payload.Batch.TX.Type) {
 					payload.addMessageUpdate(payload.Messages, core.MessageStateReady, core.MessageStateSent)
--- a/internal/batch/batch_processor_test.go
+++ b/internal/batch/batch_processor_test.go
@@ -75,6 +75,18 @@ func mockRunAsGroupPassthrough(mdi *databasemocks.Plugin) {
 	}
 }
 
+type testConflictError struct {
+	err error
+}
+
+func (tce *testConflictError) Error() string {
+	return tce.err.Error()
+}
+
+func (tce *testConflictError) IsConflictError() bool {
+	return true
+}
+
 func TestUnfilledBatch(t *testing.T) {
 	log.SetLevel("debug")
 	coreconfig.Reset()
@@ -127,6 +139,17 @@ func TestUnfilledBatch(t *testing.T) {
 	mdi.AssertExpectations(t)
 	mth.AssertExpectations(t)
 	mim.AssertExpectations(t)
+}
+
+func TestHandleDispatchConflictError(t *testing.T) {
+	cancel, _, bp := newTestBatchProcessor(t, func(c context.Context, state *DispatchPayload) error {
+		conflictErr := testConflictError{err: fmt.Errorf("pop")}
+		return &conflictErr
+	})
+	defer cancel()
+	bp.dispatchBatch(&DispatchPayload{})
+	bp.cancelCtx()
+	<-bp.done
 }
 
 func TestBatchSizeOverflow(t *testing.T) {


EnriqueL8:
Thanks @shorsher - I see the problem now

I was thinking we could approach this one differently by using the backend ID that is included in the transaction to EVMConnect so that when the operation does go ahead and retries using the EVMConnect connector the connector should be able to handle the case where the transaction successfully worked. Instead of marking it as Failed or not returning an error.

shorsher:
I think the bug is that FireFly is not correctly handling the conflict error @EnriqueL8. It gets properly classified in the operations manager as a 409 (which it is), but just missing the step that unblocks the processor.  EVMConnect receives and successfully handles the transaction, I don't see how using a different ID would make any difference. We would still be resubmitting the transaction.

internal/blockchain/ethereum/ethereum.go
```diff
@@ -623,8 +625,12 @@ func (e *Ethereum) invokeContractMethod(ctx context.Context, address, signingKey
...
 		SetError(&resErr).
 		Post("/")
-	if err != nil || !res.IsSuccess() {
-		return resErr.SubmissionRejected, common.WrapRESTError(ctx, &resErr, res, err, coremsgs.MsgEthConnectorRESTErr)
+	log.L(ctx).Infof("count %d", e.count)
+	if err != nil || !res.IsSuccess() || e.count == 0 {
+		e.count = 1
+		wrapErr := common.WrapRESTError(ctx, &resErr, res, err, coremsgs.MsgEthConnectorRESTErr)
+		log.L(ctx).Errorf("forced ethereum invokeContractMethod: %s", wrapErr)
+		return resErr.SubmissionRejected, wrapErr
```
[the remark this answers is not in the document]
shorsher: from the PR description:

> The changes in ethereum.go allow you to reproduce the issue, fyi @EnriqueL8. Start your start up, apply this patch, the first broadcast you submit will always fail. Subsequent ones will succeed, but be marked as Ready

awrichar (approved):
I think this is the correct fix.

The only alternative that I was considering, is whether Operations Manager should swallow the error. We have very specific conflict handling added there in https://github.com/hyperledger/firefly/pull/1406 for HTTP 409, but even though it does some processing of the error, that code elects to still _return_ the error upward. You could make the case that Operations Manager has handled the error and could clear it from bubbling up any further - but that would potentially have wider implications.

So all of that to say, I think re-parsing the conflict error up at this level feels like the safest solution to the bug.

internal/batch/batch_processor.go
```diff
@@ -633,6 +633,11 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
 						}
 					}
 				}
+				conflictErr, conflictTestOk := err.(operations.ConflictError)
+				if conflictTestOk && conflictErr.IsConflictError() {
+					// We know that the connector has received our batch, so we shouldn't need to retry
+					return true, nil
+				}
```
EnriqueL8, line 640: Agreed this is the right approach but we are missing the `else` section where we update the message state so in the issue the first time EVMConnect call you get EOF and the second time you get a conflict so it never updated that state for the case where it's as pinned batch tx

EnriqueL8 (commented):
Great stuff, and agree with @awrichar on the solution. The other approach we had discussed around checking the op being successful was more complicated. 

One comment on the successful flow that might have been missed  and missing codecov

internal/batch/batch_processor.go
shorsher: Updated to add the message state, thanks for catching.

internal/batch/batch_processor.go
```diff
@@ -633,6 +633,12 @@ func (bp *batchProcessor) dispatchBatch(payload *DispatchPayload) error {
 						}
 					}
 				}
+				conflictErr, conflictTestOk := err.(operations.ConflictError)
+				if conflictTestOk && conflictErr.IsConflictError() {
+					// We know that the connector has received our batch, so we shouldn't need to retry
+					payload.addMessageUpdate(payload.Messages, core.MessageStateReady, core.MessageStateSent)
```
EnriqueL8, line 639: I think we need the whole block and we can extract this to a function
```Golang
if core.IsPinned(payload.Batch.TX.Type) {
   payload.addMessageUpdate(payload.Messages, core.MessageStateReady, core.MessageStateSent)
} else {
   payload.addMessageUpdate(payload.Messages, core.MessageStateReady, core.MessageStateConfirmed)
}
```
For pinned batches we have to wait for the blockchain TX that is why we set them as sent and you might have just assumed this was always the case in conflict mode. But this batch processor is completed agnostic to how the batch is handled, it could be a conflict from data exchange or any other plugin needed to dispatch that batch. In the case where it's not pinned, we don't need to wait for TX so we can set it as confirmed

EnriqueL8: Furthermore there is a snippet of code above that handles a cancelled batch pin and this is an edge case but if the last error before the cancel is a conflict we will update the state of the initial batch incorrectly from cancelled to sent

PR #1596, opened 2024-11-01
use trivy ECR DB

shorsher:
Aqua security (trivy maintainers) now publish their vulnerability databases on ECR, which does not have rate limits. Using ECR should fix the 429 we were seeing when pulling from the default ghcr database that the trivy tool has built-in.


EnriqueL8 (approved):
Thank you, this is great

PR #1598, opened 2024-11-12
fix: DB write of upsert pins in run as group

EnriqueL8:
In the case where FireFly is restarted and the listener for BatchPin events is recreated in the blockchain connector, FireFly will re-index all the events into the system. The Pins table has an index for uniqueness of `batch_id+hash+namespace` so it will not allow inserting a duplicate record. In this case, the code falls back to upsetting the batch which is correct but this code is run in the context of DB TX which will never succeed as the previous insert fails.

You get the error `current transaction is aborted, commands ignored until end of transaction block`, explanation at 
https://stackoverflow.com/questions/10399727/psqlexception-current-transaction-is-aborted-commands-ignored-until-end-of-tra why you get this error.

So there are two options:
- When we insert we ignore on conflict, so don't error and we always upsert 
- We upsert using a different DB TX. In this case I achieve this by using a different context

### DIFF
--- a/internal/events/batch_pin_complete.go
+++ b/internal/events/batch_pin_complete.go
@@ -1,4 +1,4 @@
-// Copyright © 2023 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -125,10 +125,15 @@ func (em *eventManager) persistContexts(ctx context.Context, batchPin *blockchai
 	log.L(ctx).Warnf("Batch insert of pins failed - assuming replay and performing upserts: %s", err)
 
 	// Fall back to an upsert
+	// Use a different context to not re-use DB TX from a group run
+	// We only get here if it fails so this would never work in the same DB TX
+	upsertCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	for _, pin := range pins {
-		if err := em.database.UpsertPin(ctx, pin); err != nil {
+		if err := em.database.UpsertPin(upsertCtx, pin); err != nil {
 			return err
 		}
 	}
+
 	return nil
 }
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ containerlogs
 .vscode/*.log
 .idea
 doc-site/site
+doc-site/venv


EnriqueL8:
Tested locally and you can see that the insert fails but then the upset is fine 
```
[2024-11-12T12:04:08.983Z] DEBUG SQL<- query pins (0.30ms) dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.983Z] DEBUG Existing pin returned at sequence 4 dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.983Z] DEBUG Batch cache hit 16169315-2dcb-4505-8bb4-b01f8abbba25/b9843b5d70d2d40618a37d27574d5bc0e3a83cf91300a416e103eb4b202dd062 ns=default pid=1 role=aggregator
[2024-11-12T12:04:08.984Z] ERROR SQL insert failed: UNIQUE constraint failed: pins.namespace, pins.hash, pins.batch_id, pins.idx sql=[ INSERT INTO pins (namespace,masked,hash,batch_id,batch_hash,idx,signer,dispatched,created) VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9) ]: UNIQUE constraint failed: pins.namespace, pins.hash, pins.batch_id, pins.idx dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.984Z]  WARN Batch insert of pins failed - assuming replay and performing upserts: FF00177: Database insert failed: UNIQUE constraint failed: pins.namespace, pins.hash, pins.batch_id, pins.idx dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.985Z] DEBUG SQL<- query pins (0.28ms) dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.985Z] DEBUG Existing pin returned at sequence 5 dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.987Z] DEBUG SQL<- query pins (0.56ms) dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.987Z] DEBUG Existing pin returned at sequence 6 dbtx=tBaB41Bs ns=default pid=1 role=event-manager
[2024-11-12T12:04:08.987Z] DEBUG Batch cache hit 73cef8e4-2b65-45d7-9e0d-288cecb740a9/3c69dee7f91d4a7e5271be2421d6bdfac3c0a0e04c60f141766ca8b09852a0fd ns=default pid=1 role=aggregator
```

internal/events/batch_pin_complete.go
```diff
@@ -125,10 +125,15 @@ func (em *eventManager) persistContexts(ctx context.Context, batchPin *blockchai
 	log.L(ctx).Warnf("Batch insert of pins failed - assuming replay and performing upserts: %s", err)
 
 	// Fall back to an upsert
+	// Use a different context to not re-use DB TX from a group run
+	// We only get here if it fails so this would never work in the same DB TX
+	upsertCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	for _, pin := range pins {
-		if err := em.database.UpsertPin(ctx, pin); err != nil {
+		if err := em.database.UpsertPin(upsertCtx, pin); err != nil {
```
awrichar, line 133: Seems like a very bad idea to intentionally do this outside the database transaction.

awrichar (changes requested):

internal/events/batch_pin_complete.go
awrichar: The batch insert above should be non-fatal (should be configured to return an empty result on conflict)

EnriqueL8:
No longer need as was already fixed by https://github.com/hyperledger/firefly/commit/b26fa2717cd7df4c0d90c4809ae08a0104086706

PR #1599, opened 2024-11-20
Fix reindexing token pool created events when creation and active are different steps

EnriqueL8:
I have a set of token pools created and the first one I created then activate it to be published. 

The indexer restarted indexing events from the chain and saw the TokenPoolCreation event for the created pool and then it extract the pool information from the operation and compares that to the existing pool in DB but you get the error 
```
 fftokens event attempt 1: FF10407: Rejected token pool 'f582f2c9-06bb-4f65-8842-39cafba75959' - conflicts with existing: f582f2c9-06bb-4f65-8842-39cafba75959 pid=1 proto=fftokens role=event-loop
```

This is due to this line of code https://github.com/hyperledger/firefly/blob/62ee71f675ac5717b7c16e034b75ec1c69920d88/internal/definitions/handler_tokenpool.go#L96 but in the case of this pool being published which is the case here then we can set publish to true.

This throws an error which causes the websocket between FF and Token Connector to disconnect and then it reconnect and gets the same message of TokenPoolCreation and the loop just repeated indefinitely... 

I think if the IDs are the same we should just ignore the error and log that we received a duplicate event

### DIFF
--- a/internal/definitions/handler_tokenpool.go
+++ b/internal/definitions/handler_tokenpool.go
@@ -1,4 +1,4 @@
-// Copyright © 2023 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
--- a/internal/definitions/handler_tokenpool.go
+++ b/internal/definitions/handler_tokenpool.go
@@ -92,6 +92,11 @@ func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, stat
 			}
 		}
 
+		if existing.ID.Equals(pool.ID) {
+			log.L(ctx).Warnf("Received duplicate event for token pool creation with ID=%s, ignoring and carrying on for token connector: %s", pool.ID, pool.Connector)
+			break
+		}
+
 		// Any other conflict - reject
 		return HandlerResult{Action: core.ActionReject, CustomCorrelator: correlator}, i18n.NewError(ctx, coremsgs.MsgDefRejectedConflict, "token pool", pool.ID, existing.ID)
 	}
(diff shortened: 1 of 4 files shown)


internal/events/token_pool_created.go
```diff
@@ -131,12 +131,24 @@ func (em *eventManager) loadFromOperation(ctx context.Context, pool *tokens.Toke
...
 		return nil, nil
 	}
 
+	// If we have successfully activated the pool
+	// Then set the pool published to true
+	// As it won't be in the initial payload
+	if activateOp != nil && activateOp.Status == core.OpStatusSucceeded {
+		stagedPool.Published = true
+	}
+
```
EnriqueL8, line 151: actually don't need this https://github.com/hyperledger/firefly/blob/62ee71f675ac5717b7c16e034b75ec1c69920d88/internal/assets/token_pool.go#L150 as we are handling it here

internal/definitions/handler_tokenpool.go
```diff
@@ -92,6 +92,11 @@ func (dh *definitionHandler) handleTokenPoolDefinition(ctx context.Context, stat
 			}
 		}
 
+		if existing.ID.Equals(pool.ID) {
```
awrichar, line 95: I'm trying to understand how we got to this point with a duplicate ID. Your description indicates we are processing a `TokenPoolCreated` event, so we're coming in via `eventManager.TokenPoolCreated()` and then `definitionSender.DefineTokenPool()` and then here.

But in `eventManager.TokenPoolCreated()`, there is a check for an [existing pool](https://github.com/hyperledger/firefly/blob/62ee71f675ac5717b7c16e034b75ec1c69920d88/internal/events/token_pool_created.go#L162). Is that check failing to find the existing pool and allowing it to get down to here?

awrichar: Also worth confirming... are you running the latest release (v1.3.2)?

EnriqueL8: Yes, running the latest v1.3.2 release

EnriqueL8:
This was an issue of pointing at a different contract address and the catching the conflict was actually useful, so we do not need this change

PR #1624, opened 2025-01-29
updated docsite workflow

### LINKED ISSUE #1593: Use the template doc workflow
Reuse this https://github.com/hyperledger-labs/documentation-template/blob/main/.github/workflows/ci.yml

Veer0x1:
## Proposed changes

Added automated documentation deployment pipeline using GitHub Actions. This workflow automatically builds and deploys documentation when changes are pushed to main branch or when new version tags are created.

Fixes #1593 

<hr>

## Types of changes

- [ ] Bug fix 
- [x] New feature added
- [ ] Documentation Update 

<hr>

- [x] My Pull Request title is in format <code>< issue name ></code> eg <code>Added links in the documentation</code>.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] My changes have sufficient code coverage (unit, integration, e2e tests).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,55 +1,47 @@
-name: Docs Site
+name: update documentation site
 
 on:
   push:
     branches:
       - main
-  pull_request:
-  release:
-    types: [released]
+    tags:
+      - v*
+
+permissions:
+  contents: write
+
 jobs:
-  build:
+  deplou:
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
     steps:
-      - name: Checkout source
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
-          fetch-depth: 0
-
-      - name: Setup Docs Deploy
+          fetch-depth: 0 # fetch all commits/branches
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - uses: actions/cache@v4
+        with:
+          key: ${{ github.ref }}
+          path: .cache
+      - name: Install Python dependencies
         run: |
-          git config --global user.name "GitHub Actions"
-          git config --global user.email "noreply@github.com"
-
-      - name: Get the latest tag
+          cd doc-site
+          pip install -r requirements.txt
+      - name: Configure git user
         run: |
-          git fetch --tags
-          latest_tag=$(git tag -l | sort -V | grep -v "rc" | tail -n 1)
-          echo "latest tag: $latest_tag"
-          echo "LATEST_TAG=$latest_tag" >> $GITHUB_ENV
-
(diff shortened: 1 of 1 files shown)


EnriqueL8 (changes requested):
Sorry @Veer0x1 we had to force push a few commits into main to clean up some large files so this PR needs to be rebased on that to remove the extra commits

Veer0x1:
> Sorry @Veer0x1 we had to force push a few commits into main to clean up some large files so this PR needs to be rebased on that to remove the extra commits

done

.github/workflows/docs.yml
```diff
@@ -1,55 +1,47 @@
...
-    types: [released]
+    tags:
+      - v*
+
+permissions:
+  contents: write
+
 jobs:
-  build:
+  deplou:
```
EnriqueL8, line 14:
```suggestion
  deploy:
```

.github/workflows/docs.yml
```diff
@@ -1,55 +1,47 @@
...
+          # Strip git ref prefix from version
+          echo "${{ github.ref }}"
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && ALIAS=$(echo $VERSION | sed -e 's/^v//')
+          # If building from main, use latest as ALIAS
+          [ "$VERSION" == "main" ] && ALIAS=latest
+          echo $VERSION $ALIAS
+          mike deploy --push --update-aliases $VERSION $ALIAS
+          mike set-default latest
```
EnriqueL8, line 47: Sorry so this issue was about reusing that workflow but I incorrectly made an assumption and now I've realized that this is not the behavior we want. In the previous version, we set `latest` to be the latest released tag. So as of today `v1.3.2` but this new change would make `main` be latest.  Latest is also the default when users land on the page so I think it should stay the newest version as that is version that users should consume

EnriqueL8: Based on reading the `mike` docs https://github.com/jimporter/mike?tab=readme-ov-file I think keeping latest to point a released version instead of main is the correct behavior in my opinion

EnriqueL8: FYI @tkuhrt as I know you built this workflow and the plan was to re-use some code instead of FireFly keeping its own. We could do a GitHub dispatch action in the future but based on the above behavior might make sense to keep our own workflow

EnriqueL8 (changes requested):
Sorry @Veer0x1 the behavior might not be the one we intend to consume

EnriqueL8:
Closing this one due to inactivity

Issue #1622, opened 2025-01-29
Enhancement transaction receipt confirmations for EVM chains


EnriqueL8:
### What would you like to be added?

This is the next step in FIR-18 https://github.com/hyperledger/firefly-fir/pull/18

# Reworking Receipt handler to use new block listener

## Existing Behaviour 

Today in FireFly, the EVMConnect plugin will submit transactions to the blockchain connector and then wait for a receipt notification to be returned by the connector. The blockchain connector keeps track of all the transactions it cares about, so the ones that were submitted to it. Internally this connector has a block indexer that is fetching new blocks from the chain. It then filters out the transactions from those blocks and cross references with the transactions that were submitted. If it finds any, it will submit a notification back to FireFly Core with the status of that transaction (Successful, Failed).  Unfortunately these notifications are not guaranteed to be in the order of the blockchain so some operations might get confirmed before others. 

This diagram attempts to show the current behavior:

<img width="895" alt="Image" src="https://github.com/user-attachments/assets/82209849-b375-4abb-a470-1cd95d3be175" />

## Suggested Changes

As explained in the FIR-18 , the benefits and necessity to process transactions in the **order** of the block on the blockchain drives this piece of work to use the new block listener in FireFly core to confirm operations/transactions that have been submitted!

The ides is to create new block listener in the blockchain connector and listen to those events in FireFly core through a websocket. Then the Ethereum plugin can process those events and update the relevant operations in FireFly. This would guarantee that FireFly processes the transactions in the order of the blocks in the chain and only acknowledges to the blockchain connector the next block once it has finished processing all the transactions in the current block it received.

<img width="1398" alt="Image" src="https://github.com/user-attachments/assets/85042275-3dfc-4d13-b045-bffe3bb3b6b0" />

There is a sticky point in this approach. Today the block event form the block listener **only includes the TX hashes**, but we need the transaction receipts to confirm that the transaction went through correctly. 

Example of an event:
```
{
    "batchNumber": 1,
    "events": [
        {
            "blockHash": "0x06b0a191be7106dfe4b5dce6fd24956803d2904f6be4091c7c2bb12702cab7c2",
            "blockNumber": "33",
            "listenerId": "0194ad9b-d1ec-0420-d1f4-b641148375dd",
            "listenerName": "block",
            "listenerType": "blocks",
            "parentHash": "0x5fac4b9a3537336e4d65852095c19bc10446f56256cfdd580de23ecaeb7d95d1",
[...]

PR #1637, opened 2025-02-10
Operation handling for reliable receipts

EnriqueL8:
## Proposed changes

Add a new function in the Operation Handler called `BulkOperationUpdates` that takes in a series of OperationUpdate and will notify the submitter when the TX transaction has been committed to the database. 

This is important for blockchain plugins that want to preserve ordering of receipts and updating the operation before acknowledging to the blockchain connector that FireFly has accepted those receipts.

### Reasons why I went this way:
- The `idempotency key` is the **Operation ID** so correlation is easy in the blockchain plugin between the receipt and the operation
- Blockchain plugin framework already has the concept of **callbacks for updating operations** with transaction information 
- These callback go through the operation manager and finally end up in the DB. **Today it's a fire and forget** as the data gets offloaded to some arbitrary amount **of workers to write to the DB**.
- Need a way to communicate to the blockchain plugin when the **operation updates has been committed** so it can acknowledge a batch in the durable event stream and carry on processing. This function is now **synchronous** and it's the responsibility of the connector to wait.

### Key part of the code to looks at

- `doBatchUpdateWithRetry` runs in a `retry.Do` and in a DB RunAsGroup, it retries forever see https://github.com/hyperledger/firefly/blob/c7e6058d9139ec72302c077e29f34ba8404df802/internal/operations/operation_updater.go#L184-L199 

- Update worker get picked based on operation ID here https://github.com/hyperledger/firefly/blob/c7e6058d9139ec72302c077e29f34ba8404df802/internal/operations/operation_updater.go#L90

- Workers call `doBatchUpdateWithRetry` after timeout or reached a limit to write https://github.com/hyperledger/firefly/blob/c7e6058d9139ec72302c077e29f34ba8404df802/internal/operations/operation_updater.go#L172-L180 - **this is in memory and causes ordering problems!**

Contributes to https://github.com/hyperledger/firefly/issues/1622

Example Blockchain plugin code that has a **durable websocket** communication:

```golang

func (bc *blockchainConnector) eventLoop() {
   // Get some events from the blockchain connector through a weboscket for transactions that have completed: Receipts

   
   // Synchronous
   err := bc.callbacks.BulkOperationUpdates(ctx, namespace, updates, onCommit)
   if err != nil {
       // Retry based on what the error is
   }

  
  // Acknowledging receipts once they have been committed to the DB
  bc.ws.ack()
}
```

## Types of changes

- [ ] Bug fix 
- [x] New feature added
- [ ] Documentation Update 

<hr>

- [x] My Pull Request title is in format <code>< issue name ></code> eg <code>Added links in the documentation</code>.

<hr>


internal/operations/manager.go
```diff
@@ -77,6 +78,18 @@ type operationsManager struct {
 	cache     cache.CInterface
 }
 
+// SubmitBulkOperationUpdate implements Manager.
+func (om *operationsManager) SubmitBulkOperationUpdates(ctx context.Context, updates []*core.OperationUpdate, onCommit chan<- bool) {
```
EnriqueL8, line 82: Reworking this to be able to handle errors

pkg/core/operation.go
```diff
@@ -245,6 +245,7 @@ func ParseNamespacedOpID(ctx context.Context, nsIDStr string) (string, *fftypes.
 
 type OperationCallbacks interface {
 	OperationUpdate(update *OperationUpdate)
+	BulkOperationUpdates(ctx context.Context, updates []*OperationUpdate, onCommit chan<- error)
```
peterbroadhurst, line 248: I think passing a channel might be a little sticky. We've gone back and forth on this a few times.

My feeling is a callback function is much easier to pass in, and you then leave the caller the responsibility to decide if they want that callback to use a channel or not.

EnriqueL8: Yes! https://github.com/hyperledger/firefly/pull/1637#issuecomment-2678407214 I got to the same conclusion after reading more code recently and that is what I'm reworking 💪🏼

internal/operations/manager.go
```diff
@@ -77,6 +78,18 @@ type operationsManager struct {
...
+// SubmitBulkOperationUpdate implements Manager.
+func (om *operationsManager) SubmitBulkOperationUpdates(ctx context.Context, updates []*core.OperationUpdate, onCommit chan<- error) {
+	for _, update := range updates {
+		errString := ""
+		if update.ErrorMessage != "" {
+			errString = fmt.Sprintf(" error=%s", update.ErrorMessage)
+		}
+		log.L(om.ctx).Debugf("%s updating operation %s status=%s%s", update.Plugin, update.NamespacedOpID, update.Status, errString)
+	}
+	om.updater.SubmitBulkOperationUpdates(ctx, updates, onCommit)
```
peterbroadhurst, line 90: What is it that absolutely guarantees that this list of receipts in the order they were submitted, will be committed before any other receipts submitted to this function?

internal/operations/operation_updater.go
```diff
@@ -93,6 +93,68 @@ func (ou *operationUpdater) pickWorker(ctx context.Context, id *fftypes.UUID, up
...
+			continue
+		}
+
+		validUpdates = append(validUpdates, update)
+	}
+
+	// Notice how this is not using the workers
+	// The reason is because we want for all updates to be stored at once in this order
+	// If offloaded into workers the updates would be processed in parallel, in different DB TX and in a different order
+	go func() {
```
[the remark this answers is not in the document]
EnriqueL8: I thought about making it synchronous and wanted the connector to be able to wait for a notification back but you are right if we are treating this API as a Persist and wait until done in the connector this should absolutely be synchronous

internal/operations/operation_updater.go
```diff
@@ -93,6 +93,68 @@ func (ou *operationUpdater) pickWorker(ctx context.Context, id *fftypes.UUID, up
...
+		// Copy of the array
+		// TODO make this a deep copy
+		updates := validUpdates
+
+		// This retries forever until there is no error
+		// but returns on cancelled context
+		if len(updates) > ou.conf.maxInserts {
```
peterbroadhurst, line 124: You can choose to push this responsibility back to the connector. I don't think you need to try and do sub-pagination inside of this interace.

internal/operations/operation_updater.go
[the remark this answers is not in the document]
peterbroadhurst: Think that could be a big simplification overall to this. Push the responsibility to the connector to:
1. Manage the goroutine that is the receipt writer
2. Perform suitable performance batching into this code
3. Handle all errors/retries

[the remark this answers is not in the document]
EnriqueL8: Yeah, that simplifies massively this code to a synchronous bulk write of receipts

pkg/core/operation.go
peterbroadhurst: Cool - I think maybe with the discussion below around making the interface synchronous, this might become a moot point

internal/operations/manager.go
peterbroadhurst: 👍 - with the other discussions to make it a synchronous interface, I think you could just handle this as a comment on the API doc:

> An insertion ordering guarantee is only provided when this code is called on a single goroutine inside of the connector. It is the responsibility of the connector code to allocate that routine, and ensure that there is only one.

EnriqueL8:
@peterbroadhurst This is good to re-review. Test coverage should be good as well.

There was a slight doubt on if this function should restrict and validate the Plugin name inside each operation update, I left it up to the connector plugin to put the right one but I might add validation there or make it pass a `plugin.Named()` and use that for all the operation updates

internal/operations/operation_updater.go
```diff
@@ -93,6 +93,37 @@ func (ou *operationUpdater) pickWorker(ctx context.Context, id *fftypes.UUID, up
...
+func (ou *operationUpdater) SubmitBulkOperationUpdates(ctx context.Context, updates []*core.OperationUpdate) error {
+	validUpdates := []*core.OperationUpdate{}
+	for _, update := range updates {
+		ns, _, err := core.ParseNamespacedOpID(ctx, update.NamespacedOpID)
+		if err != nil {
+			log.L(ctx).Warnf("Unable to update operation '%s' due to invalid ID: %s", update.NamespacedOpID, err)
+			continue
+		}
+
+		if ns != ou.manager.namespace {
```
peterbroadhurst, line 106: Would be good to think carefully if this is safe/needed. This API will only be implemented by updated connectors, so will be new environments (not migration ones).

If you can do the investigation to check into why this code was in the place it was copied from, I _think_ you'll find it's related to a migration scenario back from single-listener to multi-listener days.

If you can confirm that, then it is much safer to return an error to the connector code if it supplies bad data.

internal/operations/operation_updater.go
```diff
@@ -93,6 +93,37 @@ func (ou *operationUpdater) pickWorker(ctx context.Context, id *fftypes.UUID, up
...
 }
 
+// SubmitBulkOperationUpdates will wait for the commit to DB before calling the onCommit
+func (ou *operationUpdater) SubmitBulkOperationUpdates(ctx context.Context, updates []*core.OperationUpdate) error {
+	validUpdates := []*core.OperationUpdate{}
+	for _, update := range updates {
+		ns, _, err := core.ParseNamespacedOpID(ctx, update.NamespacedOpID)
+		if err != nil {
+			log.L(ctx).Warnf("Unable to update operation '%s' due to invalid ID: %s", update.NamespacedOpID, err)
+			continue
```
peterbroadhurst, line 103: ... same as below, to returning an error now we're on a goroutine owned by the connector

EnriqueL8: Agreed, updating to return an error

internal/operations/operation_updater.go
```diff
@@ -181,25 +212,32 @@ func (ou *operationUpdater) updaterLoop(index int) {
...
 
+func (ou *operationUpdater) doBatchUpdateAsGroup(ctx context.Context, updates []*core.OperationUpdate) error {
+	err := ou.database.RunAsGroup(ctx, func(ctx context.Context) error {
+		return ou.doBatchUpdate(ctx, updates)
+	})
+	if err != nil {
+		return err
+	}
+	for _, update := range updates {
+		if update.OnComplete != nil {
```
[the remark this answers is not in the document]
EnriqueL8: I struggled with this one, it's still available in the struct `core.OperationUpdate` but it doesn't serve any purpose as we are returnin asynchronously so the top connector go routine can do what it needs when the updates have been committed. 

So we can remove but we need to leave it for the existing flow and just comment on this flow that it will not be used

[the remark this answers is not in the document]
peterbroadhurst: Why @EnriqueL8 ?
Because you think there's other stuff on `core.OperationUpdate` that needs to be on this batch API?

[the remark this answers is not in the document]
EnriqueL8: The rest of the fields in `core.OperationUpdate` are needed in this batch API so yes we could create a new type that has that subset of fields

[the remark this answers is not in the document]
EnriqueL8: Thanks @peterbroadhurst for the nudge and suggestion - move to `OperationUpdateAsync` and `OperationUpdate`

peterbroadhurst (approved):
Thanks @EnriqueL8 - looks great, and appreciate you going round the loop with me on this

PR #1652, opened 2025-02-28
[metrics] [dataexchange] [networkmap] DXConnect Callbacks for Node Identity Check Metrics

onelapahead:
## Proposed changes

Private messaging can be easily disrupted if the identity broadcasted DX certificates are out-of-sync with the certificates actually in-use by a DX plugin. This can be true for both the OSS plugins and other proprietary DX plugin implementations. We've documented this procedure, and are considering further automating this procedure within FireFly or auxiliary tooling in future releases.

For now, FireFly has all the means to monitor this and make it known to the administrators running it via metrics and logs - so that they can use their judgement on whether to update a profile on-chain, or rollback their DX certs, etc.

This PR adds a new `CheckNodeIdentityStatus` to the `NetworkMap`, a `DXConnect` callback to the DX plugin, and a `CheckNodeIdentityStatus` on the DX plugin. These methods and callbacks allow for either the networkmap or the definition handlers to check the status whenever either the DX plugin changes its cert, or the admins change the node's identity state. The DX plugin itself is namespace-less and has no concept of what local network nodes might be using it and what their config is, so it has to be provided a node identity from a namespace resource (networkmap, definition handlers) to reconcile it with its own expectations of what "status" is, relative to the DX's configuration.

The following cases are covered to ensure the metrics below are as accurate as possible:
1. Whenever a DX reconnects, either due to FireFly starting up or DX starting up or network issues, FireFly sees if its needs to re-initialize the DX. On this new connection is a great moment to ask DX for its cert and compare it to what was broadcasted, additionally the DX cert can be examined to see when a soonest expiry of its cert chain is for further monitoring. Using the `DXConnect` callback, the `Orchestrator` can be notified of the new connection, and ask the `NetworkMap` to do the status checks. Often, the first connection is during namespace initialization, and that fails bc the namespace isn't started, see case 2.
2. When starting, the `Orcehstrator`, it calls the networkmap's `CheckNodeIdentityStatus` to ensure its up-to-date since the first call failed while the namespace was started.
3. Whenever an identity is claimed, we check to see if its the local node identity i.e registration, and if so call the status check.
4. Whenever an identity is updated, we check to see if its the local node identity i.e. PATCH profile of a new cert, and if so call the status check.

New metrics `ff_multiparty_node_identity_dx_mismatch` and `ff_multiparty_node_identity_dx_expiry_epoch` are added so that the `NetworkMap` can then publish these status findings as gauges that can be easily monitoring by a timeseries and alerting system.

### Additional changes

- returns a 400 if `profile` is not set on a `PATCH /identitites/{iid}` as we found the code currently assumes a `profile` object is always provided and if its not that causes a nil error / connection abruptly closed
- all existing metrics are given a `ns` label so our users can monitoring the different metrics from the perspective of the various namespaces they may be running. **NOTE**: for Prometheus and other timeseries monitoring systems this could greatly increase the cardinality of the metrics being produced if there are a lot of namespaces running. But this is necessary to be able to identify if message, etc. are failing or pending for a particular namespace vs. another w/o inspecting logs.

<hr>

## Types of changes

- [x] Bug fix 
- [x] New feature added
- [ ] Documentation Update
- [x] New metrics added / updated

<hr>

- [ ] My Pull Request title is in format <code>< issue name ></code> eg <code>Added links in the documentation</code>.

<hr>

TODO

## Other Information

TODO


internal/definitions/handler_identity_claim.go
```diff
@@ -246,5 +246,23 @@ func (dh *definitionHandler) handleIdentityClaim(ctx context.Context, state *cor
 		event := core.NewEvent(core.EventTypeIdentityConfirmed, identity.Namespace, identity.ID, nil, core.SystemTopicDefinitions)
 		return dh.database.InsertEvent(ctx, event)
 	})
+
+	if identity.Type == core.IdentityTypeNode {
+		nodeDID, err := dh.identity.GetLocalNodeDID(ctx)
+		if err != nil {
+			return HandlerResult{Action: core.ActionRetry}, err
```
peterbroadhurst, line 253: This seems to require that we've successfully published our node identity at the point this runs, in all circumstances.

Have you done a though exercise on new network setup to confirm that is true in all cases (syncing a long lived chain and publishing your own identity after many others for example)?

peterbroadhurst (commented):
Great bit of work @onelapahead 👍 

I have popped a bunch of questions in there for you to go through before we close out on this if that's ok. Some might not require changes.

The most important is regarding safety of introducing a local node DID lookup on various paths.

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -295,6 +304,14 @@ func (h *FFDX) beforeConnect(ctx context.Context, w wsclient.WSClient) error {
 			return fmt.Errorf("DX returned non-ready status: %s", status.Status)
 		}
 	}
+
+	for _, cb := range h.callbacks.handlers {
+		err := cb.DXConnect(h)
```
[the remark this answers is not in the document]
onelapahead: correct never want the callbacks to impact finishing establishing the connection, will move the logging to the callback themselves then and do less error handling

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -448,6 +465,90 @@ func (h *FFDX) TransferBlob(ctx context.Context, nsOpID string, peer, sender fft
...
+	}()
+
+	dxPeer, err := h.GetEndpointInfo(ctx, node.Name) // should be the same as the local node
+	if err != nil {
+		return err
+	}
+
+	// if this occurs, it is either a misconfigured / broken DX or likely a DX that is compatible from an API perspective
+	// but does not have the same peer info as the HTTPS mTLS DX
+	if dxPeer.GetString("cert") == "" {
```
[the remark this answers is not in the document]
onelapahead: Our APIs as JSON return/expect these certs to have the literal `\n` characters in them, otherwise the strings are not valid for JSON blobs. So its definitely smelly and awkward, but thats how they have always worked.

We've never had FF aware code of the `cert` field in the profile up until now, but it reflects whats returned by the DX plugin's `/api/v1/id` endpoint and as to remain compatible with that. Whenever you're PATCHing the node's identity during a cert rotation you have to take a PEM bundle and do a replace of the newlines as a result.

In the future, we should be handling embedded PEM/file blobs in JSON objects as base64 encoded strings to avoid this awkwardness.

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -448,6 +465,90 @@ func (h *FFDX) TransferBlob(ctx context.Context, nsOpID string, peer, sender fft
...
+		block, rest = pem.Decode(rest)
+		if block == nil {
+			break
+		}
+
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("failed to parse non-certificate within bundle: %v", err)
+		}
+		if leafCert == nil || cert.NotAfter.Before(leafCert.NotAfter) {
```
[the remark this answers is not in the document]
onelapahead: > Would we simply not want the earliest expiry in the package?

we definitely want that, and thats what this code is doing. Calling it `leafCert` is misleading. But the `NotAfter` / `Before` are time operations, so we're essentially looking for the cert in the stack who's `NotAfter` is the soonest.

internal/definitions/handler_identity_claim.go
onelapahead: > This seems to require that we've successfully published our node identity at the point this runs, in all circumstances.

So that was not my intent, but lemme re-read the code. The _new_  `GetLocalNodeDID` func is meant to be just building the DID string from the multiparty manager's local node field, as opposed to the `GetLocalNode` func that did already exist but to your point actually cared about the local node being in the DB/caches bc it had been registered or was registering.

I had not thought that the `multiparty.LocalNode` field comes purely from config, not if the local node has been registered. From what I've seen testing locally (and the E2E passing) these assumptions seem correct.

internal/dataexchange/ffdx/ffdx.go
```diff
@@ -448,6 +465,90 @@ func (h *FFDX) TransferBlob(ctx context.Context, nsOpID string, peer, sender fft
...
+		return err
+	}
+
+	if node == nil {
+		return i18n.NewError(ctx, coremsgs.MsgNodeNotProvidedForCheck)
+	}
+
+	var mismatchState = metrics.NodeIdentityDXCertMismatchStatusUnknown
+	defer func() {
+		if h.metrics != nil && h.metrics.IsMetricsEnabled() {
```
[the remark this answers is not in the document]
onelapahead: Added more debug/warn logs in appropriate places to make this more meaningful even when metrics are disabled.

internal/definitions/handler_identity_claim.go
onelapahead: Yes, confirmed that multiparty manager's `LocalNode` is purely a config field:
```
func (mm *multipartyManager) LocalNode() LocalNode {
	return mm.config.Node
}
```

so this code does not require a node to be registered in order to work. Its just seeing if the local node DID we expect based on config is being claimed/updated.

internal/dataexchange/ffdx/ffdx.go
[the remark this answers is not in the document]
onelapahead: I've renamed the vars, left some comments, and updated one of the tests to reverse the order of certs in a bundle to ensure the soonest expiry is always chosen and that's clearer thats what the code is intended to do

internal/definitions/handler_identity_claim.go
peterbroadhurst: Great stuff, thanks for checking through that

peterbroadhurst (approved):

internal/networkmap/update_identity.go
```diff
@@ -44,6 +44,11 @@ func (nm *networkMap) updateIdentityID(ctx context.Context, id *fftypes.UUID, dt
 		return nil, i18n.NewError(ctx, coremsgs.Msg404NoResult)
 	}
 
+	// TODO is this right ? code below assumes this is true and errors otherwise
```
EnriqueL8, line 47: This is only true for Node based identity I believe - but would need to double check

EnriqueL8, line 47: I believe this is right as it's one of the things that can be updated.

internal/metrics/identity.go
```diff
@@ -0,0 +1,77 @@
...
+		Help: "Timestamp, in Unix epoch format, of node identity DX cert's expiry",
+	}, namespaceLabels)
+}
+
+func RegisterIdentityMetrics() {
+	registry.MustRegister(NodeIdentityDXCertMismatchGauge)
+	registry.MustRegister(NodeIdentityDXCertExpiryGauge)
+}
+
+// TODO should this type live elsewhere ??
```
EnriqueL8, line 50: I wonder if this should be exposed in the `/status/multiparty` API we recently added https://hyperledger.github.io/firefly/latest/swagger/#/Default%20Namespace/getStatusMultiparty

EnriqueL8 (commented):
Thanks for this contribution @onelapahead! A few TODOs in there if you can resolve in a follow up

PR #1665, opened 2025-03-18
Update contract listener documentation to be clearer

### LINKED ISSUE #1656: docs: Contract listener topic isn't clear about which fields are required for multi-event listeners
### Current State

A few areas could do with some improvement:

 - [x] The complete example shows a single-filter example, and if you just drop a multi-filter example into it (which some users are likely to do) you end up with an invalid contract listener request. It would be better if this example was converted to use the multi-filter format (i.e. no top-level `event` field in the payload) even if it only contains one event
 - [x] The example has both event ABI inline, and a reference to an interface. I don't think you need the latter if you have the former?
 - [x] Also the FireFly Interface field is described as follows: `Deprecated: Please use 'interface' in the array of 'filters' instead` but it isn't described as optional, which I believe it is since it isn't needed if you pass the ABI of the event in the filter itself. So it might be useful to remove `interface` from the examples that include event ABI inline, and possibly have an example that does reference an `interface` later in the document
 - [x] Similarly, the `eventPath` is only needed when a FireFly Interface is being referenced. So it could probably go in a later example, not the main one.

### Desired State

See notes in `current state` for recommended changes

EnriqueL8:
## Proposed changes

Fixes #1656 
<hr>

## Types of changes

- [ ] Bug fix 
- [ ] New feature added
- [x] Documentation Update 

<hr>

<hr>

### DIFF
--- a/doc-site/docs/reference/types/_includes/contractlistener_description.md
+++ b/doc-site/docs/reference/types/_includes/contractlistener_description.md
@@ -97,7 +97,7 @@ As noted throughout this document, the behavior of listeners has changed in v1.3
 
 The two input formats supported when creating a contract listener are shown below.
 
-### With Event Definition
+### With event definition
 
 In these examples, the event schema in the FireFly Interface format is provided describing the event and its parameters. See [FireFly Interface Format](../firefly_interface_format.md)
 
--- a/doc-site/docs/reference/types/_includes/contractlistener_description.md
+++ b/doc-site/docs/reference/types/_includes/contractlistener_description.md
@@ -93,7 +93,7 @@ As noted throughout this document, the behavior of listeners has changed in v1.3
 - The `signature` field is preserved at the listener level
   - The format has been changed as described above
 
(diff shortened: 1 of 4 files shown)


doc-site/docs/reference/types/_includes/contractlistener_description.md
```diff
@@ -93,12 +93,104 @@ As noted throughout this document, the behavior of listeners is changed in v1.3.
...
+    "address": "0xa5ea5d0a6b2eaf194716f0cc73981939dca26da1"
+  },
+  "options": {
+    "firstEvent": "newest"
+  },
+  "topic": "simple-storage"
+}
+```
+
+### With Interface Reference
```
matthew1001, line 189:
```suggestion
### With interface reference
```

matthew1001 (approved):
The update looks great, thanks. Just a couple of suggestions on heading casing

peterbroadhurst (approved):

PR #1668, opened 2025-03-24
Release notes 1.3.3

EnriqueL8:
## Proposed changes

Release notes for 1.3.3 release

<hr>

## Types of changes

- [ ] Bug fix 
- [ ] New feature added
- [x] Documentation Update 

<hr>

### DIFF
--- a/doc-site/docs/releasenotes/index.md
+++ b/doc-site/docs/releasenotes/index.md
@@ -2,7 +2,28 @@
 title: Release Notes
 ---
 
+# Release Notes
+
 [Full release notes](https://github.com/hyperledger/firefly/releases)
+
+## [v1.3.3 - Mar 24, 2025](https://github.com/hyperledger/firefly/releases/tag/v1.3.3)
+
+What's New:
+- Add new method to provide reliable receipt storage for blockchain plugins
+- Changes in metrics:   
+    - Added new metrics for Data Exchange 
+    - Added a namespace label to existing metrics and change gauges to vectors
+    - Added `monitoring` configuration section and deprecated `metrics` section
+    - Added HTTP Response Time and Complete Gauge Support to `firefly-common`
+- Enhancements to private messaging with callback metric support and fix to use network namespace
+- Fix update handling of duplicate batches to avoid editing existing messages
+- Miscellaneous bug fixes and minor improvements
+- Documentation updates, new troubleshooting section for multiparty messages
+- CVE fixes and adoption of OpenSSF scorecard on key repositories
+    
+### Migration consideration
+
+As part of the changes to the metrics to add the new `namespace` label, we changed from using a Prometheus `Counter` to a `CounterVec`. As a result there is no default value of `0` on the counter, which means when users query for a specific metric such as `ff_message_rejected_total` it will not be available until the `CounterVec` associated with that metric is incremented. This has been safely added as part of a patch release and determined to be an easy upgrade for SRE monitoring these metrics. 
 
 ## [v1.3.2 - Oct 3, 2024](https://github.com/hyperledger/firefly/releases/tag/v1.3.2)
 
@@ -24,7 +45,7 @@ What's New:
     See [Contract Listeners](../reference/types/contractlistener.md) for details
 - New multiparty status API at `/status/multiparty`
 
-## [v1.3.0 - April 25, 2024](https://github.com/hyperledger/firefly/releases/tag/v1.1.0)
+## [v1.3.0 - April 25, 2024](https://github.com/hyperledger/firefly/releases/tag/v1.3.0)
 
 [Migration guide](1.3_migration_guide.md)
 
@@ -50,9 +71,6 @@ What's New:
 - Custom HTTP headers can be passed through to FireFly dependency microservices
 - Evmconnect is now the default blockchain connector for Ethereum based FireFly stacks
 
-# Release Notes
-
-[Full release notes](https://github.com/hyperledger/firefly/releases)
 
 ## [v1.1.0 - September 12, 2022](https://github.com/hyperledger/firefly/releases/tag/v1.1.0)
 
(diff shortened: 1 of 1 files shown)


doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
 ---
 
+# Release Notes
+
 [Full release notes](https://github.com/hyperledger/firefly/releases)
 
+## [v1.3.3 - Mar 24, 2025](https://github.com/hyperledger/firefly/releases/tag/v1.3.3)
+
+What's New:
+- Add new method to provide reliable receipt storage for blockchain plugins
```
peterbroadhurst, line 12:
```suggestion
- Add new interface for blockchain plugins to stream receipt notifications with `ack` to core
    - Allows strictly ordered delivery of receipts from blockchain plugins that support it
    - Allows resilience on receipt delivery to core, against a checkpoint maintained in the connector
```

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
+- Changes in metrics:   
+    - Added new metrics for Data Exchange
```
peterbroadhurst, line 14: What metrics are added and why? Should this be first in the list?

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
+
 [Full release notes](https://github.com/hyperledger/firefly/releases)
 
+## [v1.3.3 - Mar 24, 2025](https://github.com/hyperledger/firefly/releases/tag/v1.3.3)
+
+What's New:
+- Add new method to provide reliable receipt storage for blockchain plugins
+- Changes in metrics:   
+    - Added new metrics for Data Exchange 
+    - Added a namespace label to existing metrics and change gauges to vectors
```
peterbroadhurst, line 15: This is two different bullets

doc-site/docs/releasenotes/index.md
EnriqueL8: Let me add them

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
+    - Added `monitoring` configuration section and deprecated `metrics` section
```
peterbroadhurst, line 19: This is a significant one - given this is about endpoint listeners. Probably a separate section to the metrics (as the purpose is actually unrelated to metrics - it's to allow the port that hosts the Open Telemetry endpoint, to also host other things like status endpoints).

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
+    - Added HTTP Response Time and Complete Gauge Support to `firefly-common`
+- Enhancements to private messaging with callback metric support and fix to use network namespace
```
peterbroadhurst, line 21: These seem like to separate bullets

doc-site/docs/releasenotes/index.md
EnriqueL8: Yeah, good point

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,29 @@
...
+- Fix update handling of duplicate batches to avoid editing existing messages
```
peterbroadhurst, line 22:
```suggestion
- Fix to issue that could result in messages being marked `Pending` on re-delivery of a batch over the network
```

doc-site/docs/releasenotes/index.md
EnriqueL8: The latter is a result of the first, we need to change to a vector to add labels

doc-site/docs/releasenotes/index.md
EnriqueL8: So it was already a separate server and we just change the configuration name to make it more readable and makes more sense

doc-site/docs/releasenotes/index.md
EnriqueL8: but I don't think we need to mention it here but instead in the migration consideration

doc-site/docs/releasenotes/index.md
EnriqueL8: so this is instead the fix

doc-site/docs/releasenotes/index.md
```diff
@@ -2,8 +2,35 @@
...
+        - This resulted in a new configuration section of `monitoring` to be more appropriate than `metrics` which has now be deprecated. 
+- Fix to issue that resulted in retried private messages using local namespace rather than the network namespace
+- Fix to issue that could result in messages being marked `Pending` on re-delivery of a batch over the network
+- Miscellaneous bug fixes and minor improvements
+- Documentation updates, new troubleshooting section for multiparty messages
+- CVE fixes and adoption of OpenSSF scorecard on key repositories
+    
+### Migration consideration
+
+As part of the changes to the metrics to add the new `namespace` label, we changed from using a Prometheus `Counter` to a `CounterVec`. As a result there is no default value of `0` on the counter, which means when users query for a specific metric such as `ff_message_rejected_total` it will not be available until the `CounterVec` associated with that metric is incremented. This has been safely added as part of a patch release and determined to be an easy upgrade for SRE monitoring these metrics.
```
peterbroadhurst, line 32:
```suggestion
As part of the changes to the metrics to add the new `namespace` label, we changed from using a Prometheus `Counter` to a `CounterVec`. As a result there is no default value of `0` on the counter, which means when users query for a specific metric such as `ff_message_rejected_total` it will not be available until the `CounterVec` associated with that metric is incremented. This has been determined to be an easy upgrade for SRE monitoring these metrics, hence inclusion in a patch release. 
```

peterbroadhurst (approved):

PR #1680, opened 2025-04-28
Update to go 1.23

SupernaviX:
## Proposed changes

Update the project's go version to 1.23. Also update the base alpine image to 3.21, and some miscellaneous build dependencies to match.

The motivation is to update a transitive dependency on golang.org/x/crypto to the latest. The current version is breaking the build, because it [has a vulnerability](https://go.googlesource.com/crypto/+/refs/tags/v0.35.0) due to some unbounded buffer.

<hr>

## Types of changes

- [x] Bug fix 
- [ ] New feature added
- [ ] Documentation Update 

<hr>

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.

<hr>

## Other Information

Y'all should consider setting up dependabot or something.


EnriqueL8:
Thanks @SupernaviX we normally try to update all the versions up from the dependencies  starting with the common library https://github.com/hyperledger/firefly-common/blob/main/go.mod but recently the Golang version have been quite good keeping backwards compatibility

SupernaviX:
@EnriqueL8 that CVE is the same one that I ran into here, which was making it impossible to open PRs. I'll match the versions from that PR.

ssmirr:
I think this PR should also update firefly-signer to v1.1.21 for another CVE patch: https://github.com/hyperledger/firefly-signer/releases/tag/v1.1.21

EnriqueL8:
Awesome thanks - yes that is correct FF Core has it's own APIServer and doesn't use that part of FF Common but it does use other parts

Makefile
```diff
@@ -23,7 +23,7 @@ lint: ${LINT}
 ${MOCKERY}:
 		$(VGO) install github.com/vektra/mockery/v2@latest
 ${LINT}:
-		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.54.0
+		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(shell $(VGO) env GOPATH)/bin v1.64.8
```
EnriqueL8, line 26: Why this change?

EnriqueL8 (commented):
Looks great - one small question

Makefile
SupernaviX: yep switching back worked without a hitch

EnriqueL8: Thanks for the clarification!

EnriqueL8 (approved):
Looks good!

doc-site/docs/swagger/swagger.yaml
```diff
@@ -106,10 +110,13 @@ paths:
                         a smart contract, this field can include a blockchain specific
                         contract identifier. For example an Ethereum contract address,
                         or a Fabric chaincode name and channel
+                      nullable: true
```
EnriqueL8, line 113: This is breaking and it's due to an issue in FF Common , going to revert this

PR #1707, opened 2025-10-31
Add manifest.json to release artifacts

### LINKED ISSUE #1706: manifest.json as release artifact.
### What would you like to be added?

Add `manifest.json` to release artifacts.

### Why is this needed?

This can help with that issue on https://github.com/hyperledger/firefly-cli/issues/346

Windol:
Ensure the `manifest.json` file is generated if it does not exist and upload it as part of the GitHub release process.

It fixes #1706

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -107,3 +107,21 @@ jobs:
             UI_RELEASE=${{ steps.manifest.outputs.UI_RELEASE }}
             BUILD_VERSION=${{ steps.build_tag_generator.outputs.RELEASE_TAG }}
             GIT_REF=${{ github.ref }}
+
+      - name: Ensure manifest.json exists
+        run: |
+          if [ ! -f manifest.json ]; then
+            echo "manifest.json not found in repo root - generating via manifestgen.sh"
+            ./manifestgen.sh || true
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_release.yml
```diff
@@ -107,3 +107,21 @@ jobs:
...
+      - name: Upload manifest.json to GitHub Release
+        if: github.event_name == 'release'
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ./manifest.json
+          asset_name: manifest.json
+          asset_content_type: application/json
```
EnriqueL8: Why is this not a separate workflow?

I also think the generating a manifest should not be done as part of the workflow because it might not match the branch. We have a tag branch per release, and it's better to make sure the release matches that imo.

EnriqueL8: Okay this is fine as it check the manifest.json exist before generating so @dan-r  is going to pick up removing the generation in case something goes wrong

EnriqueL8 (approved):
Thanks @Windol

PR #1716, opened 2026-02-27
Protect operations database from non-string data in error messages fr…

davecrighton:
## Proposed changes

This change tests the error string before making an update to the operations table checking that it is a valid UTF-8 string and does not contain null bytes. The second check is required because UTF-8 considers a single null byte to be a valid string but Postgres does not.

Part of the fix for #1717 

<hr>

## Types of changes

- [x] Bug fix 
- [ ] New feature added
- [ ] Documentation Update 

<hr>

<hr>

<hr>

## Other Information


peterbroadhurst (dismissed):

internal/operations/operation_updater.go
```diff
@@ -422,7 +425,15 @@ func (ou *operationUpdater) resolveOperation(ctx context.Context, ns string, id
...
-		update = update.Set("error", *errorMsg)
+		// PostgreSQL text columns reject null bytes and invalid UTF-8 sequences.
+		// Null bytes (0x00) are valid UTF-8 but rejected by PostgreSQL, so check both.
+		if !utf8.ValidString(*errorMsg) || strings.ContainsRune(*errorMsg, 0) {
+			hexString := hex.EncodeToString([]byte(*errorMsg))
+			log.L(ctx).Warnf("Error message contains invalid UTF-8 or null bytes - encoding as hex: %s", hexString)
+			update = update.Set("error", hexString)
+		} else {
+			update = update.Set("error", *errorMsg)
+		}
```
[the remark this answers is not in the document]
davecrighton: Suggested changes made, made small alteration to the suggested naming to indicate that this is a UTF8 string.

pkg/utils/dbstring.go
```diff
@@ -0,0 +1,42 @@
...
+	"unicode/utf8"
+
+	"github.com/hyperledger/firefly-common/pkg/log"
+)
+
+// DBSafeUTF8StringFromPtr returns a DB-safe UTF-8 string from a pointer.
+// Nil pointers return "". Strings containing invalid UTF-8 sequences or null
+// bytes (which PostgreSQL rejects in text columns even though null bytes are
+// technically valid UTF-8) are hex-encoded instead, with a warning logged.
+func DBSafeUTF8StringFromPtr(ctx context.Context, s *string) string {
```
EnriqueL8, line 32: Eventually this should be moved to FF Common

pkg/utils/dbstring.go
```diff
@@ -0,0 +1,42 @@
...
+	if s == nil {
+		return ""
+	}
+	if !utf8.ValidString(*s) || strings.ContainsRune(*s, 0) {
+		hexString := hex.EncodeToString([]byte(*s))
+		log.L(ctx).Warnf("String contains invalid UTF-8 or null bytes - encoding as hex: %s", hexString)
```
EnriqueL8, line 38: So this is an one way and then on the API surface we don't decode this right?

EnriqueL8 (commented):
Just one question if we should handle the API side of it getting the original error..

pkg/utils/dbstring.go
davecrighton: Correct if an error is hex encoded as it is inserted into the DB it remain hex encoded on the API surface.

davecrighton: More broadly the query is whether we should decode hex bytes coming out the other side of the database in order to rehydrate the initial context.

I assert that the answer is no since the entire call surface expects this field to be a human readable string, strings containing null bytes or non-UTF8 characters likely already pose a risk to consumers that are expecting  strings. These would appears as \u0000 in serialized json which at best will present a garbled error message and at worst may be treated as string terminators. It's safer to present a consistently encoded string to the user to deal with these abnormal cases.

Additionally due to https://github.com/hyperledger/firefly-evmconnect/pull/184 and https://github.com/hyperledger/firefly-signer/pull/98 a known source of these malformed error strings will now be decoded in a more friendly way preserving the context of nested ABI.

Therefore this particular change should be viewed as a belt-and-braces protection for preventing unforseen types of malformed error from completely blocking transaction processing (and blocks submission of transaction with following nonces etc) due to:

```
[2026-02-27T12:37:44.884Z] ERROR SQL update failed: pq: invalid byte sequence for encoding "UTF8": 0x00 sql=[ UPDATE operations SET opstatus = $1, error = $2, updated = $3 WHERE (id = $4 AND namespace = $5) ] dbtx=Qirvs5Un ns=ff1 opupdater=opu_004 pid=....
```

EnriqueL8: Thanks for detailed explanation here

EnriqueL8 (approved):

PR #1734, opened 2026-04-07
enable declaring custom topics for ffi interface publish and ff publish api create 

### LINKED ISSUE #1725: bug: race condition between ff_define_ffi messages and ff_define_contract_api messages
### What happened?

When a new firefly is added to an existing network, a ff_define_ffi message may be processed after the corresponding ff_define_contract_api message, causing the ff_define_contract_api message to be rejected.

### What did you expect to happen?

the ff_define_contract_api to be always processed after the ff_define_ffi message.

### How can we reproduce it (as minimally and precisely as possible)?

1. define a firefly interface
2. create and broadcast contract api
3. add new firefly member.

you may need to repeat steps 1-2 multiple times to ensure a higher chance that a message gets processed out of order.

### Anything else we need to know?

publish interface and create api actions do not take custom topic as input.

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

ka1ii:
## Proposed changes

allow clients to declare custom topics for ffi interface publish and ff publish api create

Fixes #1725

<hr>

## Types of changes

- [ x] Bug fix 
- [ x] New feature added
- [ x] Documentation Update 

<hr>

- [ ] I have read the contributing guidelines.
- [x ] I have performed a self-review of my own code or work.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generates no new warnings.
- [ x] I have added tests that prove my fix is effective or that my feature works.
- [ x] My changes have sufficient code coverage (unit, integration, e2e tests).

<hr>

image

## Other Information


internal/coremsgs/en_api_translations.go
```diff
@@ -216,4 +216,6 @@ var (
 
 	APISubscriptionStartSequenceID = ffm("api.startsequenceid", "The sequence ID in the raw event stream to start indexing through events from. Leave blank to start indexing from the most recent events")
 	APISubscriptionEndSequenceID   = ffm("api.endsequenceid", "The sequence ID in the raw event stream to stop indexing through events at. Leave blank to start indexing from the most recent events")
+
+	APICustomTopicsQueryParam = ffm("api.customTopicsQueryParam", "Custom topics for ordering definition broadcast messages")
```
EnriqueL8: Wouldn't call it query param, as it's past in at POST

EnriqueL8 (commented):
Thanks @ka1ii - would love to understand if all the comments from @awrichar have been addressed https://github.com/kaleido-io/firefly/pull/140 

Along with that suggest:
- Updating the docs
- Handling empty topics passed in better

ka1ii:
will need to update the go.mod once the pr in firefly-common gets merged https://github.com/hyperledger-firefly/common/pull/233/changes

go.mod
```diff
@@ -97,3 +97,5 @@ require (
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hyperledger-firefly/common => ../firefly-common
```
awrichar: Probably did not mean to commit this?

internal/definitions/sender_contracts.go
```diff
@@ -106,7 +106,16 @@ func (ds *definitionSender) PublishFFI(ctx context.Context, name, version, netwo
 			return i18n.NewError(ctx, coremsgs.MsgAlreadyPublished)
 		}
 		ffi.NetworkName = networkName
-		sender = ds.getFFISender(ctx, ffi)
+		// If user has explicitly defined a new set of topics at publish time
+		// we honor those and override the existing topics stored in FFI
+		// If not (i.e. when the new topics is empty), we fallback to the topics
+		// defined in the FFI
+		if len(topics) > 0 {
+			ffi.Topics = topics
```
awrichar: Do we persist these topics back to the database along this path? It seems like there could be some slight confusion here if the stored topics are not the topics that actually get used when publishing.

go.mod
EnriqueL8: https://github.com/hyperledger-firefly/common/pull/233/changes

awrichar: So this needs to be released first and then pulled in as a dependency?

EnriqueL8: yeah, getting that done

internal/definitions/sender_contracts.go
EnriqueL8: https://github.com/hyperledger-firefly/firefly/pull/1734/changes#diff-6121d61dd5e275727aee0c3b04620f64e6ca4db797bcb0ae4686cc6d30fb5584R2

awrichar: @EnriqueL8 I understand that we do persist them in general, just trying to qualify that we _definitely_ overwrite the topics along this path?

EnriqueL8: Good point it does look that way

awrichar: I wonder if persisting the "topics" list here is gaining us anything. We really only need it at the moment we call publish, and ultimately it will be stored on the attached message (which is more authoritative - as noted here there are some weird edge cases where the topics on the FFI itself could potentially not match).

awrichar (approved):
This change looks good to me (accepting the "topics" in the DTO body, but not persisting to the database).

internal/definitions/sender_contracts.go
EnriqueL8: Agreed

EnriqueL8:
I've enabled the Codecov App so can ignore the above message

EnriqueL8 (approved):

dan-r:
Tested and working in rc2. View from second member in multiparty network:

Screenshot 2026-07-27 at 16 29 35

PR #1743, opened 2026-07-07
Update firefly docker image name in build script in isolation

dan-r:
A pre-req for getting #1742 building cleanly


EnriqueL8 (dismissed):
Thanks, this is the first step in the org migration for the FF CLI to work

EnriqueL8:
https://github.com/vulnerable-code/vulndb/commit/99dec70ef9c10e4f145393917b707ea3a4f17c3e this is the new CVE caught as part of this PR, there is no version fix and we are not using OpenPGP so we can ignore it for now

EnriqueL8 (approved):
thanks @dan-r

shorsher (approved):

PR #1764, opened 2026-07-27
Bump kin-openapi and make changes to OpenAPI validation

dan-r:
Bumps kin-openapi to fix a CVE. This new version enforces a stricter OpenAPI 3 spec check than before: two path templates with the same shape can't use different parameter names. This causes validation to fail on two pre-existing routes - PUT /apis/{id} vs GET/DELETE /apis/{apiName}, and GET /identities/{iid} vs GET /identities/{did}. Changes have been made to OpenAPI validation to allow this.


internal/apiserver/swagger_validate_test.go
```diff
@@ -0,0 +1,56 @@
...
+// validateOpenAPIDoc runs the standard kin-openapi validation, but tolerates
+// ConflictingPathsError findings. FireFly intentionally addresses some
+// resources by ID in one route and by name in another at the same path
+// shape (e.g. PUT /apis/{id} vs GET/DELETE /apis/{apiName}, and
+// GET /identities/{iid} vs GET /identities/{did}). The OpenAPI 3 spec
+// disallows this, but it does not affect real routing since each operation
+// is still uniquely dispatched by its HTTP method. Any other validation
+// finding still fails as normal.
+func validateOpenAPIDoc(doc *openapi3.T) error {
+	err := doc.Validate(context.Background(), openapi3.EnableMultiError())
```
EnriqueL8, line 51: Multi Error allows the validate to return all errors individually

internal/apiserver/swagger_validate_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	var multi openapi3.MultiError
+	if !errors.As(err, &multi) {
+		multi = openapi3.MultiError{err}
+	}
+
+	var remaining openapi3.MultiError
+	for _, e := range multi {
+		var conflictErr *openapi3.ConflictingPathsError
```
EnriqueL8, line 63: I don't think we want to ignore all of these for the future, if we introduce new APIs we want to not use Conflicting Paths. So we should ignore this error for existing APIs where we cannot break consumers

EnriqueL8 (commented):
Thanks Dan, important thing to address

EnriqueL8 (approved):
Thanks @dan-r

PR #1772, opened 2026-08-05
feat(webhooks): add confirmationMode subscription option

### LINKED ISSUE #1770: Webhooks transport has no way to hold the subscription checkpoint on delivery failure, propose a `confirmationMode` option
## Summary

The webhooks event-transport plugin (`internal/events/webhooks`) never inspects the HTTP response status, or any error from the delivery attempt, before telling the event dispatcher the event was delivered. Today, **if a webhook target is down, misconfigured, or returning errors, FireFly still advances the subscription checkpoint past those events** — they're marked delivered and never redelivered, regardless of whether the target actually received them.

This may well be intentional for some existing users — a "best-effort, fire-and-forget" webhook that never blocks or backs up a subscription on a flaky endpoint is a reasonable thing to want. But there's currently no way to get the *other* behavior: hold-and-retry, at-least-once semantics, for users who need the checkpoint to reflect actual delivery success.

**This proposal does not change the current default.** It introduces a single enum option, `confirmationMode`, that names the delivery-confirmation behavior explicitly (`"fastack"`, `"besteffort"`, `"assured"`) instead of leaving it as an implicit side effect of the existing `fastack` boolean. The existing `fastack` option is kept for backward compatibility and deprecated in favor of it.

## Current behavior

`doDelivery` performs the HTTP request via `attemptRequest`, and regardless of what comes back, proceeds to acknowledge every event in the batch:

```go
func (wh *WebHooks) doDelivery(ctx context.Context, connID string, reply bool, sub *core.Subscription,
    events []*core.CombinedEventDataDelivery, fastAck, batched bool) {   // no error return

    req, res, gwErr := wh.attemptRequest(ctx, sub, events, batched)
    if gwErr != nil {
        // A network-level failure gets turned into a synthetic response...
        res = &whResponse{Status: http.StatusBadGateway, ...}
    }
    // ...and nothing here ever inspects res.Status either.

    for _, combinedEvent := range events {
        event := combinedEvent.Event
        if reply && event.Message != nil {
            // reply-mode handling
        } else if !fastAck {
            if cb, ok := wh.callbacks.handlers[sub.Namespace]; ok {
                cb.DeliveryResponse(connID, &core.EventDeliveryResponse{
                    ID:           event.ID,
                    Rejected:     false,   // <-- unconditional ack
                    Subscription: event.Subscription,
                })
            }
        }
    }
}
```

`DeliveryRequest` and `BatchDeliveryRequest` compound this — even though `doDelivery` has no way to communicate an outcome, the callers don't check anything either:

```go
func (wh *WebHooks) DeliveryRequest(ctx context.Context, connID string, sub *core.Subscription,
    event *core.EventDelivery, data core.DataArray) error {
    ...
    // TODO we don't look at the error here?
    wh.doDelivery(ctx, connID, reply, sub, []*core.CombinedEventDataDelivery{{Event: event, Data: data}}, false, false)
    return nil
}
```

The event dispatcher (`internal/events/event_dispatcher.go`) decides whether to hold the subscription's checkpoint (nack + redeliver) or advance it (ack) purely based on whether `DeliveryRequest`/`BatchDeliveryRequest` returns an `error`. Since these functions always return `nil` here, the dispatcher has no way to know a webhook delivery failed — a `500`, a connection refused, or a timeout are all indistinguishable from success.

## Scope: this is a webhooks-only gap

The dispatcher itself isn't at fault — it correctly holds or advances the checkpoint based on whatever `DeliveryRequest`/`BatchDeliveryRequest` tells it. The gap is entirely inside the webhooks transport choosing to always report success.

The **websocket** transport already gets this right by default and needs no changes here: by default (`autoAck=false`), it holds the checkpoint until the connected client sends a real ack back over the socket, and a send failure is correctly returned as an error. Its opt-in `autoAck` option shares webhooks' `fastack`-style tradeoff (ack immediately, no confirmation) — but that's an existing, already-opt-in feature, unaffected by this proposal.

**This issue and `confirmationMode` are scoped entirely to webhooks.**

## Proposed option: `confirmationMode`

Rather than a second boolean bolted onto `fastack`, name the three delivery-confirmation behaviors explicitly as a string enum:

| Value | Behavior | Is this new? |
|---|---|---|
| `"fastack"` | Ack immediately, before delivery is attempted. Delivery runs detached; its outcome never affects the checkpoint. | No — same as today's `fastack: true`. |
| `"besteffort"` | Deliver synchronously, but ack regardless of the response status or error (today's only behavior, now given a name). | No — this **is** today's default, unchanged. |
| `"assured"` | Deliver synchronously; ack only on a 2xx response. Anything else returns an error, so the dispatcher holds the checkpoint and redelivers. | **Yes — this is the new behavior this issue is asking for.** |

**Default when unset: `"besteffort"`.** This is a hard requirement it's what makes this a non-breaking change. Every subscription that doesn't explicitly set `confirmationMode` behaves exactly as it does today.

### Backward compatibility with `fastack`

- `fastack` (`WebhookSubOptions.Fastack`) is kept, but **deprecated** in favor of `confirmationMode`.
- If `confirmationMode` is unset and `fastack: true` is set (existing subscriptions created before this change), it's treated as `confirmationMode: "fastack"` — behavior is unchanged, no migration action needed.
- If `confirmationMode` is unset and `fastack` is unset/`false`, it resolves to `"besteffort"` — today's default, unchanged.
- **If `fastack: true` is set *and* `confirmationMode` is explicitly set to anything other than `"fastack"`, that's a validation error** at subscription create/update time — the two are saying contradictory things, and resolving that silently in either direction would surprise someone. Any other combination (fastack unset, or fastack true + confirmationMode "fastack") is fine.

## Proposed fix

```go
// Resolves the effective mode, handling the deprecated fastack fallback.
func confirmationModeFor(sub *core.Subscription) string {
    opts := sub.Options.TransportOptions()
    if mode := opts.GetString("confirmationMode"); mode != "" {
        return mode
    }
    if opts.GetBool("fastack") { // deprecated field, back-compat only
        return "fastack"
    }
    return "besteffort" // today's default, unchanged
}
```

```go
func (wh *WebHooks) doDelivery(ctx context.Context, connID string, reply bool, sub *core.Subscription,
    events []*core.CombinedEventDataDelivery, fastAck, batched bool) error {
    req, res, gwErr := wh.attemptRequest(ctx, sub, events, batched)
    if gwErr != nil {
        res = &whResponse{Status: http.StatusBadGateway, ...}
    }

    // Only "assured" ever inspects the response - "besteffort" (the default) and
    // "fastack" both fall straight through to the existing per-event ack loop below,
    // exactly as today.
    if !reply && confirmationModeFor(sub) == "assured" && (res.Status < 200 || res.Status >= 300) {
        return i18n.NewError(ctx, coremsgs.MsgWebhooksDeliveryFailedStatus, res.Status)
    }

    // ... existing per-event response emission, entirely unchanged ...
    return nil
}
```

```go
func (wh *WebHooks) DeliveryRequest(ctx context.Context, connID string, sub *core.Subscription,
    event *core.EventDelivery, data core.DataArray) error {
    ...
    // A non-nil error here (only possible in "assured" mode) causes the dispatcher to hold
    // the checkpoint and redeliver, rather than advancing past a failed delivery.
    return wh.doDelivery(ctx, connID, reply, sub, []*core.CombinedEventDataDelivery{{Event: event, Data: data}}, false, false)
}
```

`reply` mode is unaffected regardless of `confirmationMode`: the webhook's HTTP response *is* the payload relayed back to the original caller, whatever its status, and the existing code already routes reply mode down its own path before `fastack`/`confirmationMode` are even consulted.

Validation (in `ValidateOptions`):

```go
mode := options.TransportOptions().GetString("confirmationMode")
switch mode {
case "", "fastack", "besteffort", "assured":
default:
    return i18n.NewError(ctx, coremsgs.MsgWebhooksInvalidConfirmationMode, mode)
}
if options.TransportOptions().GetBool("fastack") && mode != "" && mode != "fastack" {
    return i18n.NewError(ctx, coremsgs.MsgWebhooksFastackConfirmationModeConflict)
}
```

## API surface

Add `ConfirmationMode` as a new typed field, and mark `Fastack` deprecated using the existing convention already established elsewhere in this codebase for deprecated fields (`SubscriptionFilter`'s `DeprecatedTag`/`DeprecatedGroup`/`DeprecatedAuthor` - Go field renamed with a `Deprecated` prefix, JSON tag unchanged for wire compatibility):

```go
// pkg/core/webhooks.go
type WebhookSubOptions struct {
    DeprecatedFastack bool   `ffstruct:"WebhookSubOptions" json:"fastack,omitempty"` // deprecated: use ConfirmationMode
    ConfirmationMode  string `ffstruct:"WebhookSubOptions" json:"confirmationMode,omitempty"`
    Reply             bool   `ffstruct:"WebhookSubOptions" json:"reply,omitempty"`
    URL               string `ffstruct:"WebhookSubOptions" json:"url,omitempty"`
    ...
}
```

Example subscription creation body (`POST /api/v1/namespaces/default/subscriptions`):

```json
{
  "name": "my-webhook-sub",
  "transport": "webhooks",
  "filter": {
    "events": "message_confirmed|message_rejected"
  },
  "options": {
    "withData": true,
    "url": "https://example.com/hook",
    "method": "POST",
    "json": true,
    "confirmationMode": "assured"
  }
}
```

## Open questions for maintainers

- Should any status ranges (e.g. `3xx` redirects) get different treatment in `"assured"` mode, or is a flat "must be 2xx" acceptable for a first pass?
- Should `"besteffort"` ever become non-default in a future major version, given it arguably violates the checkpoint's general contract? Not proposing that here - flagging as a possible future discussion.

## Testing

- **Default (nothing set):** webhook returns `500` → resolves to `"besteffort"` → behavior unchanged from today, acked, checkpoint advances. Proves no regression.
- **Existing subscription with `fastack: true`, no `confirmationMode` (migration case):** resolves to `"fastack"` → behavior unchanged from today.
- **`confirmationMode: "assured"`:** webhook returns `500` → `DeliveryRequest`/`BatchDeliveryRequest` return an error → dispatcher holds checkpoint, redelivers.
- **`confirmationMode: "assured"`, reply mode:** webhook returns `500` → still acked, response relayed as the reply payload — reply's own exemption applies regardless of mode.
- **`confirmationMode: "assured"`, batch delivery:** a `500` for the batch causes the dispatcher to redeliver the whole batch (matches existing batch nack semantics).
- **Validation - conflict:** `fastack: true` + `confirmationMode: "assured"` (or `"besteffort"`) on create/update → rejected with a clear error.
- **Validation - no conflict:** `fastack: true` + `confirmationMode: "fastack"` → accepted.
- **Validation - invalid value:** `confirmationMode: "strict"` (typo/unknown value) → rejected.

EnriqueL8:
## Proposed changes

Fixes #1770 

<hr>

## Types of changes

- [ ] Bug fix 
- [x] New feature added
- [ ] Documentation Update 

<hr>

<hr>


internal/events/webhooks/webhooks.go
```diff
@@ -457,6 +506,20 @@ func (wh *WebHooks) doDelivery(ctx context.Context, connID string, reply bool, s
 	b, _ := json.Marshal(&res)
 	log.L(ctx).Tracef("Webhook response: %s", string(b))
 
+	// Only "assured" mode inspects the response status - "besteffort" (the default) and "fastack"
+	// both fall straight through to the per-event acknowledgement loop below, exactly as before.
+	//
+	// Reply mode is exempt regardless of the mode: the webhook response *is* the payload we relay
+	// back to the original caller, whatever its status, so there is nothing to hold a checkpoint
+	// for. The 502 synthesized above for a network-level failure is treated as a failure here, as
+	// no delivery was confirmed.
```
peterbroadhurst: Think this comment could be simplified

internal/events/webhooks/webhooks.go
```diff
@@ -523,23 +587,27 @@ func (wh *WebHooks) DeliveryRequest(ctx context.Context, connID string, sub *cor
...
 				ID:           event.ID,
 				Rejected:     false,
 				Subscription: event.Subscription,
 			})
 		}
-		go wh.doDelivery(ctx, connID, reply, sub, []*core.CombinedEventDataDelivery{{Event: event, Data: data}}, true, false)
+		go func() {
+			// The event is already acknowledged, so the outcome of this detached delivery cannot
+			// affect the subscription checkpoint. doDelivery never returns an error in this mode.
+			_ = wh.doDelivery(ctx, connID, reply, sub, []*core.CombinedEventDataDelivery{{Event: event, Data: data}}, true, false)
```
peterbroadhurst, line 596: ~Suggest logging the error, rather than discarding through anon var~

^^^ thought about this, then realized as commented there's no codepath that generates an error

peterbroadhurst (dismissed):
This looks great to me @EnriqueL8 - very minor comment, but no functional concerns.

Particularly liked the Markdown docs updated on the confirmation modes. Think that's a great addition to help users choose the right mode.

internal/events/webhooks/webhooks.go
EnriqueL8: done

peterbroadhurst (approved):

PR #1782, opened 2026-09-14
Adding cache for contract API

Chengxuan:
## Proposed changes

Neither querying nor invoking a contract API in FireFly cached the contract API metadata stored in the database, so every request under load re-ran the same `GetContractAPIByName` lookup, adding unnecessary DB work and network I/O latency.

This adds a namespace-scoped, TTL-based read-through cache (`cache.contractapi.limit` / `cache.contractapi.ttl`, default 200 items / 5m, mirroring the existing `cache.methods` cache in the same manager) in front of that lookup, used by `GetContractAPI`, `InvokeContractAPI`, `AddContractAPIListener`, and `GetContractAPIListeners`. The cache is invalidated on `DeleteContractAPI`; the create/update conflict-check path (`ResolveContractAPI`) still reads the DB directly since it must see current state.

<hr>

## Types of changes

- [x] Bug fix

<hr>

- [x] My changes have sufficient code coverage (unit tests only - `internal/contracts` stays at 100% coverage; no e2e run for this change).

<hr>

N/A - backend-only change.

<hr>

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/contracts/manager.go
```diff
@@ -488,8 +502,31 @@ func (cm *contractManager) InvokeContract(ctx context.Context, req *core.Contrac
...
+// getContractAPIByName is a read-through cache in front of database.GetContractAPIByName, used by
+// every request path that resolves a contract API by name (query, invoke, listener management).
+// The cached entry is shared across callers, so it's returned as-is when httpServerURL is "".
+// When httpServerURL is set, the caller wants URLs stamped onto the result - since that mutates
+// the object, a private copy is made first so the cached entry (and other callers) are unaffected.
+func (cm *contractManager) getContractAPIByName(ctx context.Context, apiName, httpServerURL string) (*core.ContractAPI, error) {
+	cached, ok := cm.contractAPICache.Get(apiName).(*core.ContractAPI)
+	if !ok {
+		var err error
+		cached, err = cm.database.GetContractAPIByName(ctx, cm.namespace, apiName)
```
EnriqueL8:
```suggestion
		dbRecord, err = cm.database.GetContractAPIByName(ctx, cm.namespace, apiName)
```

EnriqueL8 (commented):
Thanks @Chengxuan  - this looks good , one nit

internal/contracts/manager.go
```diff
@@ -1343,6 +1378,18 @@ func (cm *contractManager) DeleteContractAPI(ctx context.Context, apiName string
...
-		return cm.database.DeleteContractAPI(ctx, cm.namespace, api.ID)
+		if err := cm.database.DeleteContractAPI(ctx, cm.namespace, api.ID); err != nil {
+			return err
+		}
+		cm.contractAPICache.Delete(apiName)
+		return nil
 	})
 }
+
+func (cm *contractManager) UpsertContractAPI(ctx context.Context, api *core.ContractAPI) error {
```
EnriqueL8, line 1389: check api is not nil?

internal/contracts/manager.go
Chengxuan: renamed to `contract` as the same variable is used to hold both the cached value and db record

internal/contracts/manager.go
Chengxuan: Not intending to introduce a new defensive check as it's a straightaway passthrough to the old function.

Keeping it as is

EnriqueL8: but you are accessing the api.Name on the cache so it's not a straight through

Chengxuan: I've added a safe guard on that new logic in case in the future `cm.database.UpsertContractAPI` no longer panics on nil `api`

EnriqueL8 (approved):
Thanks @Chengxuan !

PR #1784, opened 2026-09-16
fix(contracts): Multiple versions of an ffi that differ only by zero arg custom errors collide on cache key

davecrighton:
## Proposed changes

The symptom reported here is a failure to decode a custom error when a contract invocation is reverted.

The root cause is that `validateFFIError` builds its contribution to the method cache key from each error's parameters only. A zero-argument custom error therefore does not alter the generated cache key.

 Two interfaces in a namespace that share a method signature and their parameter-bearing errors, differing only in zero-argument errors, collide onto one methodhash_<method>_<hash> entry. 

Whichever misses the cache first supplies errorsABI for both interfaces via ParseInterface, so the other interface's extra error definitions never reach the blockchain connector. The connector cannot match those revert selectors and returns raw hex — e.g. FF23021: EVM reverted: 0x7b2ce124 instead of AmountExceedsAvailableBalance().

The solution is to include the custom error name in the cache key. 

<hr>

## Types of changes

- [x] Bug fix 
- [ ] New feature added
- [ ] Documentation Update 

<hr>

<hr>

<hr>

## Other Information


internal/contracts/manager.go
```diff
@@ -778,6 +778,7 @@ func (cm *contractManager) validateFFIError(ctx context.Context, errorDef *fftyp
 		return "", i18n.NewError(ctx, coremsgs.MsgErrorNameMustBeSet)
 	}
 	cacheKeyBuff := new(strings.Builder) // Build a big string of aggregate hashes
+	cacheKeyBuff.WriteString(errorDef.Name)
```
EnriqueL8, line 781: Thanks @davecrighton  - this looks good, spend a bit of time validating that this applies to all protocols, so all good 👍🏼

EnriqueL8 (approved):

PR #1789, opened 2026-09-21
ci: add DockSec container scan with SARIF upload

advaitpatel:
## Proposed changes

Adds a GitHub Actions workflow that scans the `Dockerfile` with
[DockSec](https://github.com/OWASP/DockSec) and uploads the results to code
scanning as SARIF, so container findings are annotated inline on pull requests.

DockSec is an OWASP Lab Project (MIT). It wraps Hadolint and Trivy and reports
Dockerfile misconfigurations and image CVEs through a single SARIF upload.

Why this repository: DockSec scores the current `Dockerfile` at **42.5/100**
across 17 findings, and four of them are Hadolint *errors* rather than style advice:

```
DL3020 error    lines 30, 32, 44, 56  Use COPY instead of ADD for files and folders
SC2261 error    line 17               Multiple redirections compete for stdout
DL3018 warning  line 17               Pin versions in apk add
DL3029 warning  line 36               Do not use --platform flag with FROM
```

`ADD` is the one worth a look: unlike `COPY` it auto-extracts archives and
accepts remote URLs, so it is a wider surface than those four lines need.
SC2261 on line 17 is a correctness bug rather than a security one - competing
stdout redirections mean one silently wins.

No issue is linked. This is a small CI-only addition, but I am happy to open an
issue first and let you confirm the approach if that is the process you prefer.

Fixes # 

<hr>

## Types of changes

- [x] Other: CI / build tooling

<hr>

- [x] I have performed a self-review of my own work.
- [x] My changes generate no new warnings (workflow passes `actionlint`).
- [ ] I have added tests - N/A, this adds no application code.
- [ ] Code coverage - N/A for a CI workflow.

<hr>

## Other Information

Scope notes:

- Runs `scan_only`, so no API key and no AI calls: fully local to the runner.
- Does not gate merges. No `fail_on`, so the job reports and never fails a build.
- Scans the Dockerfile only; it does not pull or build an image. Adding `image:`
  to the same step enables that later if you want it.
- All three actions are pinned to a commit SHA.

Disclosure: I used an AI assistant (Claude) to draft this description and generate
the workflow file from DockSec's documented Action inputs. The scan results quoted
above are from runs I did myself; the workflow passes `actionlint` and I verified its
SARIF output against the published DockSec container before opening this.

Disclosure of interest: I am the project lead for DockSec, the tool this workflow
adds. Happy to close this if it is not a fit - no hard feelings either way.


advaitpatel:
Hi @EnriqueL8 can I please get your review here? Happy to share more about the tool if needed. Thanks!

